PR #11, opened 2020-06-16
introduce common crate. Add config functionality

michaelmattig:


common/src/config/mod.rs
```diff
@@ -0,0 +1,20 @@
+use lazy_static::lazy_static;
+use std::sync::RwLock;
+use config::{Config, File};
+
+lazy_static! {
+	pub static ref CONFIG: RwLock<Config> = {
```
ChristianBeilschmidt, line 6: Why do we need an `RWLock`?
The config is loaded here and then all access is done reading. So it could just be an immutable config, I guess.

common/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+    clippy::cast_sign_loss,
+    clippy::default_trait_access,
+    clippy::missing_errors_doc,
+    clippy::module_name_repetitions,
+    clippy::must_use_candidate,
+    clippy::non_ascii_literal,
+    clippy::similar_names,
+    clippy::single_match_else,
+    clippy::type_repetition_in_bounds,
+    clippy::used_underscore_binding, // TODO: remove wrt. https://github.com/rust-lang/rust-clippy/issues/5360
```
ChristianBeilschmidt, line 18: This should not be used in `common` since it was only a workaround for `services`.

ChristianBeilschmidt: The issue is fixed by the way, so it can also be removed in `services`.

services/src/bin/main.rs
```diff
@@ -6,9 +6,14 @@ use geoengine_services::workflows::registry::HashMapRegistry;
 use geoengine_services::handlers;
 use geoengine_services::users::userdb::HashMapUserDB;
 use geoengine_services::handlers::handle_rejection;
+use geoengine_common::config::CONFIG;
+use std::error::Error;
 
 #[tokio::main]
-async fn main() {
+async fn main() -> Result<(), Box<dyn Error>> {
+    let host: std::net::IpAddr = CONFIG.read()?.get::<String>("web.host")?.parse()?;
```
ChristianBeilschmidt, line 14: Is there a way to use symbols instead of strings? This could be useful for refactoring.

cf. https://github.com/mehcode/config-rs/blob/master/examples/hierarchical-env/src/settings.rs

common/README.md
```diff
@@ -0,0 +1,2 @@
+# geo engine common
+Common functionality for the other geo engine crates
```
ChristianBeilschmidt, line 2: Is there a way to (auto) generate a config description?

services/src/bin/main.rs
```diff
@@ -6,9 +6,14 @@ use geoengine_services::workflows::registry::HashMapRegistry;
 use geoengine_services::handlers;
 use geoengine_services::users::userdb::HashMapUserDB;
 use geoengine_services::handlers::handle_rejection;
+use geoengine_common::config::CONFIG;
+use std::error::Error;
 
 #[tokio::main]
-async fn main() {
+async fn main() -> Result<(), Box<dyn Error>> {
```
ChristianBeilschmidt, line 13: Maybe we should specify our `Result` type here?

common/src/config/mod.rs
michaelmattig: This is a global variable for the whole project. If we have different binaries, e.g. webservices, CLI, administration, we may want to load additional config from other paths and sources. We also may want to dynamically update the config during runtime. For this we need an RWLock.

common/src/lib.rs
michaelmattig: 9b363240c265960a40eb6c94bcb3642e3007c538

services/src/bin/main.rs
michaelmattig: like a constant instead of "web.host"? In the link you posted they use hard-coded strings too...
Can you give an example?

common/README.md
michaelmattig: you mean generate a sample config from the config.get() calls? I dont't think so

services/src/bin/main.rs
michaelmattig: What's the benefit? This error should never occur in the first place because we can't report it to the user.

services/src/bin/main.rs
ChristianBeilschmidt: https://github.com/geo-engine/rfcs/issues/7

ChristianBeilschmidt: https://docs.rs/arc-swap/0.4.7/arc_swap/

PR #13, opened 2020-07-10
time interval display bounds fix

ChristianBeilschmidt:


datatypes/src/primitives/time_interval.rs
```diff
@@ -213,9 +213,19 @@ impl TimeInterval {
     ///     })
     /// );
     /// ```
+    #[allow(unstable_name_collisions)] // TODO: remove when `clamp` is stable
     pub fn to_geo_json_event(&self) -> serde_json::Value {
-        let start_date: DateTime<Utc> = Utc.timestamp_millis(self.start());
-        let end_date: DateTime<Utc> = Utc.timestamp_millis(self.end());
+        let min_visualizable_value = -8_334_632_851_200_001 + 1; // -262144-01-01T00:00:00+00:00
+        let max_visualizable_value = 8_210_298_412_800_000 - 1; // +262143-12-31T23:59:59.999+00:00
```
michaelmattig, line 219: use string constant instead of parsing the same date over and over again

michaelmattig:
I'm not sure that capping the date at some value defined by chrono is such ja good idea. I'd rather output either some special values for begin/end of time or don't output anything

datatypes/src/primitives/time_interval.rs
ChristianBeilschmidt: e0e9ca2

PR #14, opened 2020-07-16
Update CI and integrate rustfmt

ChristianBeilschmidt:
Please squash single commits.


.github/workflows/ci.yml
```diff
@@ -16,14 +16,22 @@ jobs:
...
-      - uses: actions-rs/clippy-check@v1
+      - name: Check with Rustfmt
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
+      - name: Check with Clippy
+        uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
```
michaelmattig, line 36: why does clippy need a github token and rustfmt doesnt?

ChristianBeilschmidt: It is a specialized GitHub action (https://github.com/actions-rs/clippy-check) that in principle posts warnings and errors to this page so that you don't have to look through the log. It is currently some kind of buggy but worked previously and they are fixing the bugs.

However, it requires the token to post here. `actions-rs/cargo` (https://github.com/actions-rs/cargo) which runs `fmt` doesn't require something like this, although they say that warnings and errors have improved UI…

PR #15, opened 2020-07-17
Equals for Feature Collections

ChristianBeilschmidt:


datatypes/src/collections/feature_collection_impl.rs
```diff
@@ -172,6 +172,45 @@ pub trait FeatureCollectionImplHelpers {
 
     /// Is the feature collection simple or does it contain multi-features?
     fn _is_simple(&self) -> bool;
+
+    fn f64_michael_equals(a: &Float64Array, b: &Float64Array) -> bool {
```
jdroenner, line 176: while i appreciate method names indicating their functionality, i suggest to change this to "f64_eq_with_nan_eq_nan". Later generations might not know what a michael is in this context.

datatypes/src/collections/feature_collection_impl.rs
```diff
@@ -670,6 +709,44 @@ macro_rules! feature_collection_impl {
...
+
+                if self.types() != other.types() {
+                    return false;
+                }
+
+                for key in self.types().keys() {
+                    let c1 = self.table().column_by_name(key).expect("column must exist");
+                    let c2 = other
+                        .table()
+                        .column_by_name(key)
```
jdroenner, line 730: is this correct? i guess two collections could have different column(names) ?
please add a test  where column names are not equal

datatypes/src/collections/feature_collection_impl.rs
ChristianBeilschmidt:
```suggestion
    /// compares two f64 typed columns
    /// treats NaN values as if they are equal
    fn f64_column_equals(a: &Float64Array, b: &Float64Array) -> bool {
```

datatypes/src/collections/feature_collection_impl.rs
```diff
@@ -670,6 +709,44 @@ macro_rules! feature_collection_impl {
...
+                        .expect("column must exist");
+
+                    match (c1.data_type(), c2.data_type()) {
+                        (DataType::Float64, DataType::Float64) => {
+                            if !Self::f64_michael_equals(downcast_array(c1), downcast_array(c2)) {
```
ChristianBeilschmidt, line 735:
```suggestion
                            if !Self::f64_column_equals(downcast_array(c1), downcast_array(c2)) {
```

PR #16, opened 2020-07-22
common access for multi features coordinates, lines and rings as traits

ChristianBeilschmidt:
@jdroenner you wanted to have a common access trait for these types.
While unfortunately, we can't have the same type (unless you tell me how to convert a Vec<Vec<X>> to &[&[X]] without creating a new vector on the fly) but I managed to find common `AsRef`s.


datatypes/src/primitives/multi_point.rs
```diff
@@ -4,6 +4,11 @@ use crate::util::arrow::ArrowTyped;
 use crate::util::Result;
 use snafu::ensure;
 
+/// A trait that allows a common access to lines of `MultiPoint`s and its references
```
michaelmattig, line 7: copy paste

ChristianBeilschmidt:
```suggestion
/// A trait that allows a common access to lines of `MultiPolygons`s and its references
```

jdroenner:
This is a very good idea to unify access to different multi feature implementations. Could you write tests for it? Also, i guess a benchmark would be helpful to see if there is a performance penalty when using the as_ref indirection.

ChristianBeilschmidt:
@jdroenner I've added tests.

I guess I don't have to benchmark the `as_ref` call for now. We should discuss our benchmarking separately, fix a current version, and then see improvements or degradations on further changes.

PR #18, opened 2020-08-12
wrap type and projection of operators in a struct

jdroenner:
this can also  hold tile size/resolution information for raster and fields for vector data

### DIFF
--- a/operators/src/engine/operator.rs
+++ b/operators/src/engine/operator.rs
@@ -13,6 +13,22 @@ pub trait Operator: std::fmt::Debug + Send + Sync {
     /// Get the sources of the `Operator`
     fn vector_sources(&self) -> &[Box<dyn VectorOperator>];
 
+    /// Get the sources of the `Operator`
+    fn raster_sources_mut(&mut self) -> &mut [Box<dyn RasterOperator>];
+
+    /// Get the sources of the `Operator`
+    fn vector_sources_mut(&mut self) -> &mut [Box<dyn VectorOperator>];
+
+    fn initialize(&mut self) -> Result<()> {
+        for ro in self.raster_sources_mut() {
+            ro.initialize()?
+        }
+        for vo in self.vector_sources_mut() {
+            vo.initialize()?
+        }
+        Ok(())
+    }
+
     fn validate_children(
         &self,
         expected_projection: ProjectionOption,
--- a/operators/src/engine/operator.rs
+++ b/operators/src/engine/operator.rs
@@ -7,6 +7,22 @@ use serde::{Deserialize, Serialize};
 /// Common methods for `Operator`s
 pub trait Operator: std::fmt::Debug + Send + Sync {}
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct OperatorImpl<P> {
+    pub params: P,
+    pub raster_sources: Vec<Box<dyn RasterOperator>>,
+    pub vector_sources: Vec<Box<dyn VectorOperator>>,
+}
+
+impl<P> Operator for OperatorImpl<P> where P: std::fmt::Debug + Send + Sync {}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SourceOperatorImpl<P> {
+    pub params: P,
+}
+
+impl<P> Operator for SourceOperatorImpl<P> where P: std::fmt::Debug + Send + Sync {}
+
 /// Common methods for `VectorOperator`s
 #[typetag::serde(tag = "type")]
 pub trait VectorOperator: Operator {
--- a/operators/src/engine/operator.rs
+++ b/operators/src/engine/operator.rs
@@ -1,6 +1,7 @@
 use super::{
     query_processor::{TypedRasterQueryProcessor, TypedVectorQueryProcessor},
-    RasterResultDescriptor, VectorResultDescriptor,
+    CloneableInitializedOperator, CloneableOperator, CloneableRasterOperator,
+    CloneableVectorOperator, RasterResultDescriptor, VectorResultDescriptor,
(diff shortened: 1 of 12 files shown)


operators/src/mock/mock_raster_point_join.rs
```diff
@@ -127,18 +119,34 @@ impl VectorOperator for MockRasterPointJoinOperator {
...
+        if self.result_descriptor.is_none() {
+            let new_result_descriptor = VectorResultDescriptor {
+                projection: self.point_sources.get(0).map_or_else(
+                    || ProjectionOption::None,
+                    |o| o.result_descriptor().projection,
+                ),
+                data_type: VectorDataType::MultiPoint,
+            };
+
+            // self.result_descriptor = Some(new_result_descriptor); //TODO: discuss if this should be interior or if method is mut...
```
jdroenner, line 132: to enable storage of the computed information, we need to change the result_descriptor method take &mut self. Is there any reason to not do this?

operators/src/engine/operator.rs
```diff
@@ -65,21 +43,236 @@ pub trait VectorOperator: Operator + CloneableVectorOperator {
...
+    /// Wrap a box around a `RasterOperator`
+    fn boxed(self) -> Box<dyn RasterOperator>
+    where
+        Self: Sized + 'static,
+    {
+        Box::new(self)
+    }
+}
+
+pub type ExecutionContext = u32;
```
ChristianBeilschmidt, line 60: Should be an empty struct I guess. Isn't the context something rather for the query method?

operators/src/engine/operator.rs
```diff
@@ -65,21 +43,236 @@ pub trait VectorOperator: Operator + CloneableVectorOperator {
...
+        self.as_ref().vector_processor()
+    }
+}
+
+pub struct InitilaizedOperatorImpl<Parameters, ResultDescriptor> {
+    pub params: Parameters,
+    pub raster_sources: Vec<Box<dyn InitializedRasterOperator>>,
+    pub vector_sources: Vec<Box<dyn InitializedVectorOperator>>,
+    pub context: ExecutionContext,
+    pub result_descriptor: ResultDescriptor,
```
ChristianBeilschmidt, line 132: Why is it not necessary to have a common trait for the result description or sufficient to just use one struct?

operators/src/engine/operator.rs
```diff
@@ -65,21 +43,236 @@ pub trait VectorOperator: Operator + CloneableVectorOperator {
...
+    }
+
+    pub fn map_projection<F>(mut self, f: F) -> Self
+    where
+        F: Fn(ProjectionOption) -> ProjectionOption,
+    {
+        self.projection = f(self.projection);
+        self
+    }
+}
```
ChristianBeilschmidt, line 274: Maybe we should put this logic into a separate file

operators/src/mock/mock_raster_point_join.rs
```diff
@@ -62,91 +63,101 @@ where
...
+        .map(InitilaizedOperatorImpl::boxed)
     }
+}
 
+impl InitializedVectorOperator
+    for InitilaizedOperatorImpl<MockRasterPointJoinParams, VectorResultDescriptor>
+{
     fn vector_processor(&self) -> Result<crate::engine::TypedVectorQueryProcessor> {
-        self.validate_children(1..2, 1..2)?;
+        // self.validate_children(self.result_descriptor().projection, 1..2, 1..2)?;
```
ChristianBeilschmidt, line 116: Why is this commented out?

operators/src/source/gdal_source.rs
```diff
@@ -391,54 +392,66 @@ where
...
+    fn initialized_operator(
+        self: Box<Self>,
+        context: crate::engine::ExecutionContext,
+    ) -> Result<Box<dyn InitializedRasterOperator>> {
+        InitilaizedOperatorImpl::create(
+            self.params.clone(),
+            context,
+            |params, _, _, _| {
+                let dataset_information =
+                    JsonDatasetInformationProvider::with_dataset_id(&params.dataset_id)?; // TODO: keep the information somewhere
```
ChristianBeilschmidt, line 408: Create state field in operator?

operators/src/engine/operator.rs
jdroenner: this context is required for information like data location and user information. One use-case is the following: Some of the sources need to check with a database and /or config file. Using this context, we can keep the config management separated from the operators.

operators/src/engine/operator.rs
jdroenner: The ResultDescriptor is (currently) specific to the kind of operator. A VectorOperator has a VectorResultDescriptor and a RasterOperator a RasterResultDescriptor. We could add a common trait for things like projection.

operators/src/engine/operator.rs
jdroenner: ok

operators/src/engine/operator.rs
jdroenner: converted it into a struct.

operators/src/mock/mock_raster_point_join.rs
jdroenner: i guess validation should be done in the initialized_operator construction step. This should be deleted.

operators/src/source/gdal_source.rs
jdroenner: done

operators/src/engine/operator.rs
ChristianBeilschmidt: I guess that trait makes sense to group common functionality, which is for now the projection.

PR #21, opened 2020-08-14
Arrow update to 1.0

ChristianBeilschmidt:


datatypes/src/error.rs
```diff
@@ -129,11 +127,86 @@ pub enum Error {
...
+            )
+            | (
+                arrow::error::ArrowError::ParquetError(e1),
+                arrow::error::ArrowError::ParquetError(e2),
+            ) => e1 == e2,
+            (arrow::error::ArrowError::DivideByZero, arrow::error::ArrowError::DivideByZero)
+            | (
+                arrow::error::ArrowError::DictionaryKeyOverflowError,
+                arrow::error::ArrowError::DictionaryKeyOverflowError,
+            ) => true,
```
jdroenner, line 189: i really don't like the chaining here. Please split it into smaller chunks.

datatypes/src/error.rs
```diff
@@ -129,11 +127,86 @@ pub enum Error {
...
+            (arrow::error::ArrowError::ExternalError(_), _)
```
jdroenner, line 190: this fails if the error is external? why not do a catch all case for other?

datatypes/src/error.rs
ChristianBeilschmidt: The linter argues against it, since its always the same result.
This conversion is somehow shit but I have no better way to do it right now.

datatypes/src/error.rs
ChristianBeilschmidt: Unfortunately, since it supplies no check for equality.
If I catch all then I have to clue when they add another case in the future that requires a check.

datatypes/src/error.rs
ChristianBeilschmidt: 4fb5cb3 
I've removed the bound and as such this overhead. All the rest compiled fine, so I guess there is no need for `PartialEq` on error types.

datatypes/src/error.rs
ChristianBeilschmidt: 4fb5cb3 
I've removed the bound and as such this overhead. All the rest compiled fine, so I guess there is no need for `PartialEq` on error types.

PR #25, opened 2020-08-21
A macro for calling generic raster processing functions

ChristianBeilschmidt:
We should introduce some macros for simplifying calling generic raster functions for all data types. Here is a first try.


michaelmattig:
please also use the macro in `operators/src/source/gdal_source.rs:439`

also, is it possible to use a similar macro for stuff like in
- `datatypes/src/collection/feature_collection.rs:236`
- `datatypes/src/operations/image/to_png.rs:49`
- `datatypes/src/primitives/feature_data.rs:59`
- `datatype/src/raster/operations/blit.rs:69`
- `services/src/handlers/wms.rs:161`
?

those look all similar to me. Is it possible to make the macro even more generic to unroll on all values of an arbitrary enum and call the same function onthere?

ChristianBeilschmidt:
I guess the technique is there and someone with actual use cases can reproduce it.
So it is up to discussion.

michaelmattig:
looks good!, but I'm not sure how operators will handle arbitrary raster data types when there are two inputs. How is the output data type determined? I would suggest to add two more macros:
- a macro for bi functions, where the data type of the first input is the bigger one and the function also return this data type
- a macro for bi functions where the data type of both inputs is the same and also is the output data type

this would make implementing operators, that know something about the relation of the data types of their inputs (e.g. they require them to be equal) much easier.

ChristianBeilschmidt:
6f11e13

ChristianBeilschmidt:
@michaelmattig can this be merged?

datatypes/src/raster/helpers.rs
```diff
@@ -0,0 +1,449 @@
...
+        match $input_raster {
+            $(
+                $crate::raster::TypedRaster2D::$variant($raster) => $function_call,
+            )+
+        }
+    };
+}
+
+/// Calls a function on two `TypedRaster2D`s by calling it on all variant combinations.
+/// Call via `call_generic_raster2d!(input, raster => function)`.
```
michaelmattig, line 44:
```suggestion
/// Call via `call_bi_generic_raster2d!(input, (raster_a, raster_b) => function)`.
```

datatypes/src/raster/helpers.rs
```diff
@@ -0,0 +1,449 @@
+/// Maps a `TypedRaster2D` to another `TypedRaster2D` by calling a function on all its variants.
```
michaelmattig, line 1: > calling a function on all its variants
technically not correct, it calls it on the actual variant combination. same for all other doc

datatypes/src/raster/helpers.rs
ChristianBeilschmidt: 009ed49

datatypes/src/raster/helpers.rs
ChristianBeilschmidt: Correct on this one but not other ones.

009ed49

michaelmattig: I meant it's not calling a function on ALL the variants, but only the actual variant of the enum. So it would be
> /// Maps a `TypedRaster2D` to another `TypedRaster2D` by calling a function on its variant.

and
> /// Calls a function on two `TypedRaster2D`s by calling it on the actual variant combination.

if it's too nitty gritty say so, we can merge regardless

ChristianBeilschmidt: ebc0176

datatypes/src/raster/helpers.rs
```diff
@@ -20,7 +20,7 @@ macro_rules! map_generic_raster2d {
     };
 }
 
-/// Calls a function on a `TypedRaster2D`  by calling it on all its variants.
+/// Calls a function on a `TypedRaster2D`  by calling it on its variant.
```
michaelmattig, line 23:
```suggestion
/// Calls a function on a `TypedRaster2D` by calling it on its variant.
```

datatypes/src/raster/helpers.rs
```diff
@@ -192,7 +192,7 @@ macro_rules! call_bi_generic_raster2d_staircase {
     };
 }
 
-/// Generates a a `TypedRaster2D` by calling a function for all variants.
+/// Generates a a `TypedRaster2D` by calling a function on its variant.
```
michaelmattig, line 195: on whose variant? In contrast to the other macros this one doesn't match a TypedRaster2D variant but a RasterDataType variant

ChristianBeilschmidt: Just make a suggestion, you are right in principle.

michaelmattig:
```suggestion
/// Generates a a `TypedRaster2D` by calling a function with the specified data type variant.
```

PR #26, opened 2020-08-25
FeatureCollection Range Filter

ChristianBeilschmidt:


datatypes/src/collections/feature_collection.rs
```diff
@@ -558,6 +570,71 @@ where
...
+            column_type.is_some(),
+            error::ColumnDoesNotExist {
+                name: column.to_string()
+            }
+        );
+
+        let column = self
+            .table
+            .column_by_name(column)
+            .expect("checked by ensure");
```
michaelmattig, line 589: ensure only checked if the column exists in types map

ChristianBeilschmidt: Yes, but that is the assumption of the whole data structure.

datatypes/src/collections/feature_collection.rs
```diff
@@ -766,11 +843,101 @@ fn struct_array_from_data(
...
+    included_lower_fn: fn(&'b A, T) -> Result<BooleanArray, ArrowError>,
+    excluded_lower_fn: fn(&'b A, T) -> Result<BooleanArray, ArrowError>,
+    included_upper_fn: fn(&'b A, T) -> Result<BooleanArray, ArrowError>,
+    excluded_upper_fn: fn(&'b A, T) -> Result<BooleanArray, ArrowError>,
+) -> Result<()>
+where
+    T: TryFrom<&'b FeatureDataValue, Error = error::FeatureCollectionError>,
+    R: RangeBounds<FeatureDataValue>,
+{
+    for range in ranges {
```
michaelmattig, line 914: wouldn't it be much more efficient to check a value against all ranges at the same time instead?

datatypes/src/collections/feature_collection.rs
michaelmattig: so it's ok that the whole process crashes if this assumption is violated?

datatypes/src/collections/feature_collection.rs
```diff
@@ -766,11 +843,101 @@ fn struct_array_from_data(
...
+    bound: Bound<&'b FeatureDataValue>,
+    array: &'b A,
+    included_fn: fn(&'b A, T) -> Result<BooleanArray, ArrowError>,
+    excluded_fn: fn(&'b A, T) -> Result<BooleanArray, ArrowError>,
+) -> Result<Option<BooleanArray>>
+where
+    T: TryFrom<&'b FeatureDataValue, Error = error::FeatureCollectionError>,
+{
+    Ok(match bound {
+        Bound::Included(v) => Some(included_fn(array, v.try_into()?)?),
```
michaelmattig, line 895: why differentiate between bound types? Don't bounds provide a contains method that abstracts this away?

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: We don't know, since the current version uses SIMD and we would have to implement it efficiently.
It is possible to achieve better, but I guess it is not the worst implementation.

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: Yes, this assumption is used throughout the collection.

datatypes/src/collections/feature_collection.rs
```diff
@@ -558,6 +570,94 @@ where
...
+        match column_type {
+            FeatureDataType::Number | FeatureDataType::NullableNumber => {
+                apply_filters(
+                    as_primitive_array::<Float64Type>(column),
+                    &mut filter_array,
+                    ranges,
+                    arrow::compute::gt_eq_scalar,
+                    arrow::compute::gt_scalar,
+                    arrow::compute::lt_eq_scalar,
+                    arrow::compute::lt_scalar,
```
michaelmattig, line 609: the last 4 arguments to the function are always the same. consider wrapping with a function that sets these

PR #27, opened 2020-08-26
moved server creation from bin to lib; added example of client commun…

michaelmattig:
…icating with server

### DIFF
--- a/services/Cargo.toml
+++ b/services/Cargo.toml
@@ -22,4 +22,6 @@ pwhash = "0.3"
 serde_urlencoded = "0.6"
 futures = "0.3"
 image = "0.23"
-reqwest = "0.10.8" # TODO: dev-dependency?
+
+[dev-dependencies]
+reqwest = "0.10.8"
--- a/services/Cargo.toml
+++ b/services/Cargo.toml
@@ -22,3 +22,4 @@ pwhash = "0.3"
 serde_urlencoded = "0.6"
 futures = "0.3"
 image = "0.23"
+reqwest = "0.10.8" # TODO: dev-dependency?
--- a/services/examples/request.rs
+++ b/services/examples/request.rs
@@ -1,4 +1,5 @@
 use geoengine_services::server;
+use std::{thread, time};
 use tokio::sync::oneshot;
 use tokio::sync::oneshot::Sender;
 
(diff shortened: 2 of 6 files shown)


services/src/server.rs
```diff
@@ -61,6 +61,7 @@ pub async fn start_server(shutdown_rx: Option<Receiver<()>>) -> Result<()> {
         .or(handlers::wms::wms_handler(workflow_registry.clone()))
         .recover(handle_rejection);
 
+    #[allow(clippy::option_if_let_else)]
```
michaelmattig, line 64: can't use option.map_or_else here because I need the handler variable in both cases and with map_or_else it would be captured by the first closure

ChristianBeilschmidt: Yeah, maybe this lint is too strict, we'll have to investigate it.

services/Cargo.toml
```diff
@@ -22,3 +22,4 @@ pwhash = "0.3"
 serde_urlencoded = "0.6"
 futures = "0.3"
 image = "0.23"
+reqwest = "0.10.8" # TODO: dev-dependency?
```
ChristianBeilschmidt, line 25: Yes, since you only use it in an example for now. We can move it back if we do Http requests from within with reqwest.

services/examples/request.rs
```diff
@@ -0,0 +1,40 @@
+use geoengine_services::server;
+use tokio::sync::oneshot;
+use tokio::sync::oneshot::Sender;
+
+// Example of a client communicating with the geo engine
```
ChristianBeilschmidt, line 5:
```suggestion
/// Example of a client communicating with the geo engine
```

services/examples/request.rs
```diff
@@ -0,0 +1,40 @@
...
+#[tokio::main]
+async fn main() {
+    // TODO: use special config for port etc. for starting the server and connecting to it
+    let (shutdown_tx, shutdown_rx) = oneshot::channel();
+
+    let (server, _) = tokio::join!(
+        server::start_server(Some(shutdown_rx)),
+        queries(shutdown_tx),
+    );
```
ChristianBeilschmidt, line 14: Isn't there a race condition that the server must start before executing `queries`?

ChristianBeilschmidt (changes requested):
Just minor stuff to address before merging.

services/examples/request.rs
michaelmattig: it worked for me, but your are right. Would it be an acceptable solution to sleep the queries task for some time before issuing the first query or do we need a backchannel from the server that indicates that the server is running? Not sure if warp supports this though...

services/Cargo.toml
michaelmattig: 92699a3d13dbdd2eb312e95b07ae9ec5bc45c9dd

services/examples/request.rs
michaelmattig: 3304dc28317951e2ab2941d70b7815d22fdaca76

ChristianBeilschmidt: This works, it's just… couldn't you instead of using `join!`, which does things in parallel, first wait for a response of the server startup and then issue the queries?

michaelmattig: there is no response from the server. The server is spawned as a tokio task which is then awaited on. This await returns only when the server is shutdown. I haven't found a hook in warp that would allow being notified once the server is ready to serve requests and I think that's because usually a server isn't used from within the server executable it is run in.

ChristianBeilschmidt: Hm, but even MAPPING can say: I'm up and running, awaiting requests. Is there an example for that? We could postpone it as a different task though.

michaelmattig: that's what I'm checking in the wait method. if there is some response from the server.

ChristianBeilschmidt: This solution is like pressing F5 in a browser. I'm asking if there is any chance that our server can state internally that it runs.
I guess it can block until it receives incoming requests.
If not, then this is fine.

michaelmattig: I haven't found anything for querying the server state or even getting notified of successful startup. The documentation of `run` states
> Run this Server forever on the current thread.

ChristianBeilschmidt (approved):

PR #28, opened 2020-08-26
Link Time Optimization

ChristianBeilschmidt:


ChristianBeilschmidt:
@michaelmattig any thoughts on that?

michaelmattig:
did you investigate the impact on build/link time, binary size and performance? Are there any drawbacks setting this flag? If it does no harm I guess we can add it.

ChristianBeilschmidt:
before LTO
grafik

with LTO
grafik

Seems to be faster, but our release time overall is quite slow.

jdroenner:
I think performance should be the factor here.

time cargo build  --release --package geoengine-services --example openlayers

with LTO:
> real    5m41,932s
> user    18m32,377s
> sys     0m34,515s

> 11M Sep 18 11:38 openlayers

> timings ~ 190ms

without LTO:
> real    3m54,383s
> user    26m6,347s
> sys     0m40,756s

> 17M Sep 18 11:44 openlayers

> timings: ~240ms

PR #29, opened 2020-08-28
Column Range Filter Operator

ChristianBeilschmidt:


operators/src/processing/column_range_filter.rs
```diff
@@ -0,0 +1,281 @@
...
+use geoengine_datatypes::collections::FeatureCollection;
+use geoengine_datatypes::primitives::{FeatureDataType, FeatureDataValue, Geometry};
+use geoengine_datatypes::util::arrow::ArrowTyped;
+use serde::{Deserialize, Serialize};
+use snafu::ensure;
+use std::convert::TryInto;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ColumnRangeFilterParams {
+    pub column: String,
```
michaelmattig, line 22: column_name? we need some convention in how to name such parameters.

michaelmattig: btw is it columns or attributes now for featurecollections? in geojson/wfs/ogr they are called properties. In mapping we called them attributes i think

operators/src/util/input.rs
```diff
@@ -0,0 +1,262 @@
+use crate::error;
+use failure::_core::convert::TryFrom;
+use serde::de::Visitor;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+/// A type that allows use inputs to be either strings or numbers
+#[derive(Debug, Clone, PartialEq)]
+pub enum StringOrNumber {
```
michaelmattig, line 8: use https://crates.io/crates/either crate to save code

ChristianBeilschmidt: Would have been an option indeed, but we have three variants, `either` provides two. And we don't need the dependency now, since I tailor-fitted it to our data types :smile:.

operators/src/processing/column_range_filter.rs
ChristianBeilschmidt: :shrug: if guess we can use the table and column naming… but you are right if it we should decide on `column_name` or just `column`… what is better for the user input?

michaelmattig: where do we use the term "table" in our project? We have to really think about going against the convention of the major tools and standards of calling attributes of features properties instead of columns.

About user input: maybe we should prefer more concise terms. The parameters need to have documentation anyway and we can specify the semantics more precisely there.

operators/src/util/input.rs
michaelmattig: Either<String,Either<Number, Decimal>> ;) Anyway I think the implementation could be improved see my other comment

operators/src/processing/column_range_filter.rs
```diff
@@ -0,0 +1,283 @@
...
+    pub ranges: Vec<(StringOrNumber, StringOrNumber)>,
```
michaelmattig, line 24: this allows mixed ranges like (0, "hi"). It only fails when the query processor is executed. I'd prefer it to fail much earlier. So either a Vec<StringOrNumberRange> that ensures validity, or maybe even two semantically separated operators NumericRangeFilter, TextualRangeFilter that are implemented using a generic ColumnRangeFilter with a parametrized typed range handler.

operators/src/processing/column_range_filter.rs
ChristianBeilschmidt: So you vote for attributes?

operators/src/processing/column_range_filter.rs
ChristianBeilschmidt: On the other hand, you want mixed ranges like `(0, 1.5)`.

Is there an advantage for separate operators? I mean, either we can check something prior to processing or we can't.

operators/src/processing/column_range_filter.rs
michaelmattig: _Attributes_ sound most natural for me, however _properties_ is more compliant to other software and standard **and** reflects the simple feature model more than _columns_. There are no columns in a simple feature collection as each feature can have a separate set of properties. Of course columns do reflect **our** model more.

operators/src/processing/column_range_filter.rs
michaelmattig: why do we want mixed ranges? just state `(0.0, 1.5)`.

Advantage of separate opterators:
- more visible distinction of the operators
- more easily extensible for data type specific additional parameters. e.g. ignore case for text

but I'm not hard set on separation.

At least we should check the validity of the input on parsing however, s.t. we don't end up with workflows in the registry that will always fail when trying to create a query processor from them

ChristianBeilschmidt: 47f20b4

operators/src/processing/column_range_filter.rs
ChristianBeilschmidt: -> separate Task

operators/src/util/input.rs
ChristianBeilschmidt: Voting for keeping the own implementation now.

operators/src/adapters/feature_collection_merger.rs
```diff
@@ -0,0 +1,250 @@
...
+        let merged_collection = if let Some(old_collection) = accum.take() {
+            // TODO: execute on separate thread?
+            old_collection.append(&new_collection)
+        } else {
+            Ok(new_collection)
+        };
+
+        match merged_collection {
+            Ok(collection)
+                if !collection.is_empty() && collection.byte_size() >= chunk_size_bytes =>
```
michaelmattig, line 62: computing the byte size is not for free. maybe store the current acum byte size and add the byte size of the added chunk to it instead of recomputing the total byte size over and over again

operators/src/adapters/feature_collection_merger.rs
```diff
@@ -0,0 +1,250 @@
...
+        } = self.as_mut().project();
+
+        let mut output: Option<Poll<Option<St::Item>>> = None;
+
+        while output.is_none() {
+            if stream.is_terminated() {
+                return Self::output_remaining_chunk(accum);
+            }
+
+            let next = ready!(stream.as_mut().poll_next(cx));
```
michaelmattig, line 103: does it really work when the result of `ready!`ist not `Ready(Ok(_))` ? The tests don't cover that case if I see it correctly

operators/src/mock/mock_feature_collection_source.rs
```diff
@@ -0,0 +1,267 @@
...
+        .unwrap();
+
+        let source = MockFeatureCollectionSource {
+            params: MockFeatureCollectionSourceParams { collection },
+        }
+        .boxed();
+
+        let serialized = serde_json::to_string(&source).unwrap();
+        let collection_bytes = [
+            65, 82, 82, 79, 87, 49, 0, 0, 144, 1, 0, 0, 16, 0, 0, 0, 0, 0, 10, 0, 14, 0, 12, 0, 11,
```
michaelmattig, line 144: why are there so many 0s?

operators/src/adapters/feature_collection_merger.rs
ChristianBeilschmidt: ebe2168

operators/src/mock/mock_feature_collection_source.rs
ChristianBeilschmidt: Empty Buffers?… I don't know and what shall I address here?

operators/src/adapters/feature_collection_merger.rs
michaelmattig: sorry for being imprecise here, I meant: what if the result is not Ready but pending?

operators/src/mock/mock_feature_collection_source.rs
michaelmattig: check the validity of the serialization result? Why does it take 1.4kB to save 3 features?

operators/src/adapters/feature_collection_merger.rs
ChristianBeilschmidt: I guess calculating the byte size is not more expensive for collections with more rows, since it depends on the number of buffers (attributes).
This means, calculating the size on the merged collection equals calculating it on the collection to append and summing the byte sizes.

The remaining question is, is the same byte size calculated multiple times?
I say no since there are two scenarios:
1. The collection is large enough and out goes the collection as well as the validity of the calculated byte size.
2. The collection is not large enough. Then it gets merged and a new byte size is the valid one which must be computed no matter what.
Hence, I doubt that it will save us computation time to store it in the merger.

operators/src/adapters/feature_collection_merger.rs
ChristianBeilschmidt: `ready!` returns early. As soon as the source calls the waker, this function will be called again and it returns `Ready` then.
How can I create a test that uses a stream that returns Pending for once without blocking…?
Is it even necessary? The logic is from `ready!` downwards.

operators/src/mock/mock_feature_collection_source.rs
ChristianBeilschmidt: Well, it is a workaround that uses the IPC format (https://arrow.apache.org/docs/format/Columnar.html#serialization-and-interprocess-communication-ipc) to have some kind of serialization.
I guess it is obvious that it is far from ideal, but at least it works for now.

operators/src/adapters/feature_collection_merger.rs
michaelmattig: I think it should be more efficient to calculate the byte_size of the individual `new_collection`s and add them up as they come. Because the way you do it you have to compute the size of the first merged chunk again every time a new chunk is merged. If you have a chunk_size of e.g. 100MB and the underlying stream produces chunks of 1 MB this might be already significant.

Of course that all only makes sense if byte_size(collection1 + collection2) = byte_size(collection1) + byte_size(collection2)

operators/src/adapters/feature_collection_merger.rs
michaelmattig: We need some stream that really produces results asynchronously to check the correct behavior. Something like my DelayOperator that somehow got deleted...

```
        self.points[0]
            .query(query, ctx)
            .then(async move |x| {
                let _ = tokio::spawn(async move {
                    thread::sleep(time::Duration::from_secs(seconds));
                })
                .await;
                x
            })
            .boxed()
    }
```

ChristianBeilschmidt: I doubt that we should use delays in tests.
Could you create something that has your desired behavior without blocking with time?

operators/src/adapters/feature_collection_merger.rs
ChristianBeilschmidt: Could you re-check that the calculation is row-bound? My previous post doubted that since if sums up the size of the buffers.

operators/src/adapters/feature_collection_merger.rs
michaelmattig: why should'nt we use delays in tests? Sleep does not use up cpu cycles, and as long as we don't sleep for multiple minutes it shouldn't be a problem for runtime. If we don't use sleep we need to constuct some interaction between two object and trigger the wake() externally which would be much more complicated.

ChristianBeilschmidt: You will block up test threads (especially if `t=1`) with sleep.

93b6272

michaelmattig: I don't think so, because tokio::spawn uses a different thread pool than the testing framework

operators/src/adapters/feature_collection_merger.rs
```diff
@@ -295,4 +295,54 @@ mod tests {
...
+                .map_err(Error::from),
+            )),
+        ];
+        stream_history.reverse();
+
+        let stream = futures::stream::poll_fn(move |cx| {
+            let item = stream_history.pop().unwrap_or(Poll::Ready(None));
+
+            if let Poll::Pending = item {
+                cx.waker().wake_by_ref();
```
michaelmattig, line 327: I think this might cause problems in multithreaded executors, as the caller is woken up before the Pending-poll arrives. Either make sure that it works in multithreaded executors or ensure that this test is run with single threaded executor (don't run as tokio::test and create a new scheduler

operators/src/adapters/feature_collection_merger.rs
ChristianBeilschmidt: Yes, but it would not spawn another test while the tests with the delays are blocking.

operators/src/adapters/feature_collection_merger.rs
ChristianBeilschmidt: dec39c6

PR #31, opened 2020-09-08
OpenLayers WMS

ChristianBeilschmidt:
The current state successfully registers a workflow with our server.
However, the WMS request of OL fails due to our case sensitive deserialization.
This is a major issue.

Example:
```
http://localhost:3030/wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=9776a5cb-2441-5e04-8b62-e8246df0f25c&CRS=EPSG%3A3857&STYLES=&WIDTH=658&HEIGHT=355&BBOX=-13888944.635145342%2C2867630.7173607955%2C-7451112.364854658%2C6340929.2826392045
```

Result:
`404 Not Found`


ChristianBeilschmidt:
@michaelmattig maybe `#[serde(alias = "name")]` is to the rescue for us?

services/examples/openlayers-wms-static/index.html
```diff
@@ -0,0 +1,72 @@
...
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.4.3/css/ol.css" type="text/css">
+    <style>
+        html, body, .map {
+            margin: 0;
+            padding: 0;
+            width: 100%;
+            height: 100%;
+        }
+    </style>
+    <script src="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.4.3/build/ol.js"></script>
```
michaelmattig, line 15: is an external dependency a good idea? We can't show this demo without internet. Also: why not use a release version?

services/Cargo.toml
```diff
@@ -12,7 +12,7 @@ edition = "2018"
 chrono = { version = "0.4", features = ["serde"] }
 geoengine-datatypes = { path = "../datatypes" }
 geoengine-operators = { path = "../operators" }
-tokio = { version = "0.2", features = ["macros"] }
+tokio = { version = "0.2", features = ["macros", "signal"] }
```
michaelmattig, line 15: dev-dependency only or add add it to main.rs as well. might make sense

services/src/ogc/wms/request.rs
```diff
@@ -29,29 +29,45 @@ pub enum GetCapabilitiesFormat {
     TextXml, // TODO: remaining formats
 }
 
+// TODO: remove serde aliases and use serde-aux and case insensitive keys
 #[derive(PartialEq, Debug, Deserialize, Serialize)]
 pub struct GetMap {
+    #[serde(alias = "VERSION")]
```
michaelmattig, line 35: couldn't we use serde_aux here?

ChristianBeilschmidt: https://docs.rs/serde-aux/0.6.1/serde_aux/container_attributes/fn.deserialize_struct_case_insensitive.html

I think it won't work because it would need to be within the enum variant.

services/Cargo.toml
ChristianBeilschmidt: 928702e

services/examples/openlayers-wms-static/index.html
ChristianBeilschmidt: 928702e

PR #32, opened 2020-09-08
first version of OpenCL support for operations on rasters (and soon f…

michaelmattig:
…eature collections)

Co-authored-by: Christian Beilschmidt <beilschmidt@mathematik.uni-marburg.de>


operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,578 @@
...
+
+        Ok(())
+    }
+
+    fn set_arguments(&mut self, params: &CLProgramParameters) -> Result<()> {
+        ensure!(
+            params.input_rasters.iter().all(Option::is_some),
+            error::CLProgramUnspecifiedRaster
+        );
+        self.output_buffers.clear();
```
ChristianBeilschmidt, line 284: Why are only the output buffers cleared?

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,578 @@
...
+                TypedRaster2D::I64(raster) => {
+                    self.add_output_buffer(idx, raster, OutputBuffer::I64)?
+                }
+                TypedRaster2D::F32(raster) => {
+                    self.add_output_buffer(idx, raster, OutputBuffer::F32)?
+                }
+                TypedRaster2D::F64(raster) => {
+                    self.add_output_buffer(idx, raster, OutputBuffer::F64)?
+                }
+            }
```
ChristianBeilschmidt, line 333: Use a macro for that?

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,578 @@
...
+use geoengine_datatypes::call_generic_raster2d;
+use geoengine_datatypes::raster::{
+    DynamicRasterDataType, Pixel, Raster2D, RasterDataType, TypedRaster2D,
+};
+use ocl::builders::{ContextBuilder, KernelBuilder, ProgramBuilder};
+use ocl::{Buffer, Device, Kernel, MemFlags, Platform, Queue};
+use snafu::ensure;
+use snafu::ResultExt;
+
+/// Whether the kernel iterates over pixels or features
```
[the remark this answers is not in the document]
michaelmattig: this is used to determine the global work size of the kernel. For rasters it is the number of pixels in the output raster and for features it would be the number of features. I don't think there will be a case where the work size is determined by the amount of attributes.

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,578 @@
...
+#[derive(PartialEq, Clone, Copy, Debug)]
+pub enum IterationType {
+    Raster,
+    Vector,
+}
+
+/// Specifies in and output types of CL program and compiles the source into a reusable `CompiledCLProgram`
+pub struct CLProgram {
+    input_rasters: Vec<RasterDataType>,
+    output_rasters: Vec<RasterDataType>,
```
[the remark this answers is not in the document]
michaelmattig: the time of rasters should be added to the RasterInfo I guess. Features are still to do.

operators/src/opencl/cl_program.rs
michaelmattig: because the input buffers are filled with data from the host slices

operators/src/opencl/cl_program.rs
michaelmattig: this macro would only be used a single time. Maybe we can create some more generic macro for such cases where two enums must comply.

michaelmattig:
Unfortunately there is an issue that results in tests failing when they are run in parallel. I created an issue https://github.com/cogciprocate/ocl/issues/189

michaelmattig:
> 
> 
> Unfortunately there is an issue that results in tests failing when they are run in parallel. I created an issue [cogciprocate/ocl#189](https://github.com/cogciprocate/ocl/issues/189)

workaround: b0179ca6a5ff2d88ea05a6190f7d868e12516bbf

datatypes/src/collections/multi_point_collection.rs
```diff
@@ -100,6 +100,44 @@ impl GeoFeatureCollectionRowBuilder<MultiPoint> for FeatureCollectionRowBuilder<
...
+
+    pub fn coordinates_offsets(&self) -> &[i32] {
+        let geometries_ref = self
+            .table
+            .column_by_name(Self::GEOMETRY_COLUMN_NAME)
+            .expect("There must exist a geometry column");
+        let geometries: &ListArray = downcast_array(geometries_ref);
+
+        let data = geometries.data();
+        let buffer = &data.buffers()[0];
```
ChristianBeilschmidt, line 135: You didn't check for an offset within the buffer.
You would need to test that somehow by subsetting the geometry column.

datatypes/src/collections/multi_point_collection.rs
```diff
@@ -100,6 +100,44 @@ impl GeoFeatureCollectionRowBuilder<MultiPoint> for FeatureCollectionRowBuilder<
     }
 }
 
+impl MultiPointCollection {
+    pub fn coordinates(&self) -> &[Coordinate2D] {
```
ChristianBeilschmidt, line 104: Maybe this method can be part of FeatureCollection?

datatypes/src/collections/multi_point_collection.rs
michaelmattig: when is the offset not zero? I need a scenario to test it

datatypes/src/collections/multi_point_collection.rs
michaelmattig: how? the nesting is different for point/lines/polygons...

jdroenner: maybe add trait for all geom types and use this in collections?

michaelmattig:
I implemented functionality for computing opencl kernels on Rasters and MultiPointCollections (and an example expression operator).

Still missing functionality:
 - Support for lines and polygons
 - Support for time intervals
 
There is also a lot of room for improvement for code conciseness and performance.
 
There is also a problem where sometimes arrow buffers are not correctly allocated. It happens sometimes when executing the tests multiple times in successions.
 
Please give me some feedback on the work and input on how we can limit the scope of this pull request s.t. it can be merged eventually.

datatypes/src/collections/multi_point_collection.rs
ChristianBeilschmidt: I guess when a buffer is a subset of another one.

datatypes/src/raster/typed_raster.rs
```diff
@@ -203,3 +366,113 @@ where
...
+    fn get_raster(raster: TypedRaster2D) -> Option<Raster2D<Self>>
+    where
+        Self: Pixel;
+
+    fn get_typed_raster(_raster: Raster2D<Self>) -> TypedRaster2D
+    where
+        Self: Pixel,
+    {
+        unimplemented!()
+    }
```
ChristianBeilschmidt, line 380: Do we need this block? This method exists on the trait anyway, but the compiler won't warn us to implement it.

datatypes/src/primitives/coordinate.rs
```diff
@@ -2,16 +2,19 @@ use crate::util::arrow::ArrowTyped;
...
 
-#[derive(Clone, Copy, Debug, Deserialize, PartialEq, PartialOrd, Serialize)]
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, PartialOrd, Serialize, Default)]
 #[repr(C)]
 pub struct Coordinate2D {
     pub x: f64,
     pub y: f64,
 }
 
+unsafe impl OclPrm for Coordinate2D {}
```
ChristianBeilschmidt, line 16: What is this? Maybe you could write a comment on that.

datatypes/src/primitives/coordinate.rs
```diff
@@ -2,16 +2,19 @@ use crate::util::arrow::ArrowTyped;
 use arrow::array::{BooleanArray, Float64Builder};
 use arrow::datatypes::DataType;
 use arrow::error::ArrowError;
+use ocl::OclPrm;
 use serde::{Deserialize, Serialize};
 use std::{fmt, slice};
 
-#[derive(Clone, Copy, Debug, Deserialize, PartialEq, PartialOrd, Serialize)]
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, PartialOrd, Serialize, Default)]
```
ChristianBeilschmidt, line 9: What is a default coordinate and why should we use it?

datatypes/src/collections/feature_collection_builder.rs
```diff
@@ -97,6 +98,15 @@ where
             _collection_type: PhantomData,
         }
     }
+
+    pub fn batch_builder(
+        self,
+        output_type: VectorDataType, // TODO: derive from CollectionType
```
ChristianBeilschmidt, line 104: Yeah, I guess it makes no sense as a parameter.

datatypes/src/collections/feature_collection.rs
```diff
@@ -805,6 +805,24 @@ where
 
         table_size + map_size
     }
+
+    /// Return the names of the columns of this type
+    pub fn column_names(&self, column_type: FeatureDataType) -> Vec<String> {
```
ChristianBeilschmidt, line 810: I guess the method name should include something like filtered since we cannot get all column names here.

datatypes/src/collections/data_types.rs
```diff
@@ -9,3 +12,37 @@ pub enum VectorDataType {
...
+    pub fn vector_data_type(&self) -> VectorDataType {
+        match self {
+            TypedFeatureCollection::Data(_) => VectorDataType::Data,
+            TypedFeatureCollection::MultiPoint(_) => VectorDataType::MultiPoint,
+            TypedFeatureCollection::MultiLineString(_) => VectorDataType::MultiLineString,
+            TypedFeatureCollection::MultiPolygon(_) => VectorDataType::MultiPolygon,
+        }
+    }
+
+    // TODO: other types
```
ChristianBeilschmidt, line 34: Mabye resolve this by adding the other two.

datatypes/src/collections/batch_builder.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    pub fn set_default_time_intervals(&mut self) -> Result<()> {
+        let mut time_intervals_builder = TimeInterval::arrow_builder(self.num_features);
+
+        let default = TimeInterval::default();
+        for _ in 0..self.num_features {
+            let date_builder = time_intervals_builder.values();
+            date_builder.append_value(default.start().inner())?;
+            date_builder.append_value(default.end().inner())?;
```
ChristianBeilschmidt, line 142: You could extract the two ints once before the loop.

datatypes/src/collections/batch_builder.rs
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        let array = Arc::new(time_intervals_builder.finish()) as ArrayRef;
+
+        self.time_array = Some(array);
+
+        Ok(())
+    }
+
+    pub fn set_points(&mut self, coords: Buffer, offsets: Buffer) -> Result<()> {
```
ChristianBeilschmidt, line 154: I'm not quite sure if our interface should use the arrow types. If the builder is only meant to be used for OpenCL, then it's fine, but otherwise we need some wrapper type I guess.

operators/src/processing/expression.rs
```diff
@@ -0,0 +1,361 @@
...
+use std::convert::TryFrom;
+use std::marker::PhantomData;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+pub struct ExpressionParams {
+    pub expression: String,
+    pub output_type: RasterDataType,
+}
+
+// TODO: custom type or simple string?
```
ChristianBeilschmidt, line 25: Since we have to put a string to OpenCL, we can keep it as a string as long as it is safe.

operators/src/processing/expression.rs
```diff
@@ -0,0 +1,361 @@
...
+            context,
+            |_, _, _, _| Ok(()),
+            |params, _, _, _, _| {
+                Ok(RasterResultDescriptor {
+                    data_type: params.output_type,
+                    spatial_reference: SpatialReferenceOption::None, // TODO
+                })
+            },
+            self.raster_sources,
+            vec![], // TODO: is it ok to omit them?
```
ChristianBeilschmidt, line 69: I guess we should check that the input vector sources are empty.

operators/src/processing/expression.rs
```diff
@@ -0,0 +1,361 @@
...
+    IN_TYPE0 B = in_data1[gid];
+    if (ISNODATA1(B, in_info1)) {
+        out_data[gid] = out_info->no_data;
+        return;
+    }
+
+    OUT_TYPE0 result = %%%EXPRESSION%%%;
+	out_data[gid] = result;
+}"#
+        .replace("%%%EXPRESSION%%%", &expression.expression);
```
ChristianBeilschmidt, line 164: Why don't you use a format string?

operators/src/processing/expression.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        let mut cl_program = CLProgram::new(IterationType::Raster);
+        cl_program.add_input_raster(RasterArgument::new(T1::TYPE));
+        cl_program.add_input_raster(RasterArgument::new(T2::TYPE));
+        cl_program.add_output_raster(RasterArgument::new(TO::TYPE));
+
+        cl_program.compile(&source, "expressionkernel").unwrap()
```
ChristianBeilschmidt, line 171: Wasn't there a problem in having clashing Program names?…

If so, maybe roll a dice and append some tokens to the end and safe it to later on destroy it.

operators/src/processing/expression.rs
```diff
@@ -0,0 +1,361 @@
...
+    TO: Pixel,
+{
+    type Output = RasterTile2D<TO>;
+
+    fn query(
+        &self,
+        query: QueryRectangle,
+        ctx: QueryContext,
+    ) -> BoxStream<Result<RasterTile2D<TO>>> {
+        // TODO: remove if zip-map approach suffices
```
ChristianBeilschmidt, line 188: I thought that was the idea.

operators/src/processing/expression.rs
```diff
@@ -0,0 +1,361 @@
...
+            __global OUT_TYPE0* out_data,
+            __global const RasterInfo *out_info)
+{
+    uint const gid = get_global_id(0) + get_global_id(1) * in_info0->size[0];
+    if (gid >= in_info0->size[0]*in_info0->size[1]*in_info0->size[2])
+        return;
+
+    IN_TYPE0 A = in_data0[gid];
+    if (ISNODATA0(A, in_info0)) {
+        out_data[gid] = out_info->no_data;
```
ChristianBeilschmidt, line 151: What if there is no NODATA in the output raster? Do we have a convention here?

ChristianBeilschmidt (changes requested):
Looks very good so far. You missed at your "missing out…" list that the expression is limited to two rasters for now.

Otherwise, I found no major issues to not merge a first version, but still have some comments that should be addressed first.

datatypes/src/collections/data_types.rs
jdroenner: We could also add a trait for this and implement it for the Types.

datatypes/src/collections/data_types.rs
```diff
@@ -9,3 +12,37 @@ pub enum VectorDataType {
...
+
+#[derive(Debug)]
+pub enum TypedFeatureCollection {
+    Data(DataCollection),
+    MultiPoint(MultiPointCollection),
+    MultiLineString(MultiLineStringCollection),
+    MultiPolygon(MultiPolygonCollection),
+}
+
+impl TypedFeatureCollection {
```
jdroenner, line 24: We cloud also add a trait for this and implement it for the types similar to RasterDAtaType

datatypes/src/collections/feature_collection_builder.rs
jdroenner: yes get it from the Geometry trait. And/or use a new trait for this.

datatypes/src/primitives/coordinate.rs
jdroenner: Is this an OpenCL parameter? why is it needed?

datatypes/src/raster/data_type.rs
```diff
@@ -150,3 +152,21 @@ where
         R::TYPE
     }
 }
+
+impl RasterDataType {
+    pub fn ocl_type(self) -> String {
```
jdroenner, line 157: i think this is an "ocl_type_name"?

datatypes/src/raster/helpers.rs
```diff
@@ -214,10 +214,35 @@ macro_rules! generate_generic_raster2d {
     };
 }
 
+/// Calls a function on a `TypedRaster2D` and some `RasterDataType`-like enum, effectively matching
+/// the raster with the corresponding enum value of the other enum.
+/// Call via `call_generic_raster2d_ext!(input, (raster, e) => function)`.
+#[macro_export]
+macro_rules! call_generic_raster2d_ext {
```
jdroenner, line 221: will this scale up to more then 3 inputs?

datatypes/src/raster/typed_raster.rs
```diff
@@ -91,96 +93,257 @@ where
...
     }
-}
 
-impl<D> Into<TypedRasterNDim<D>> for BaseRaster<D, i64, Vec<i64>>
-where
-    D: GridDimension,
-{
-    fn into(self) -> TypedRasterNDim<D> {
-        TypedRasterNDim::I64(self)
+    pub fn get_i64_ref(&self) -> Option<&BaseRaster<D, i64, Vec<i64>>> {
```
jdroenner, line 146: please sort them by size i8, i16, i32...

datatypes/src/raster/typed_raster.rs
```diff
@@ -91,96 +93,257 @@ where
...
+// }
+//
+impl<T> Into<TypedRasterNDim<Dim<[Ix; 2]>>> for BaseRaster<Dim<[Ix; 2]>, T, Vec<T>>
 where
-    D: GridDimension,
+    T: Pixel,
 {
-    fn into(self) -> TypedRasterNDim<D> {
-        TypedRasterNDim::F32(self)
+    fn into(self) -> TypedRasterNDim<Dim<[Ix; 2]>> {
```
jdroenner, line 334: can't be ndim if you set it to "2"

datatypes/src/raster/typed_raster.rs
```diff
@@ -91,96 +93,257 @@ where
...
+        T::get_typed_raster(self)
     }
 }
 
-impl<D> Into<TypedRasterNDim<D>> for BaseRaster<D, f64, Vec<f64>>
+impl<T> TryFrom<TypedRasterNDim<Dim<[Ix; 2]>>> for BaseRaster<Dim<[Ix; 2]>, T, Vec<T>>
```
jdroenner, line 339: cant be ndim if you set it to "2"

datatypes/src/raster/typed_raster.rs
```diff
@@ -91,96 +93,257 @@ where
...
 where
-    D: GridDimension,
+    T: Pixel,
 {
-    fn into(self) -> TypedRasterNDim<D> {
-        TypedRasterNDim::F64(self)
+    type Error = crate::error::Error;
+
+    fn try_from(raster: TypedRasterNDim<Dim<[Ix; 2]>>) -> Result<Self, Self::Error> {
```
jdroenner, line 345: good use of try_from :+1:

datatypes/src/raster/typed_raster.rs
jdroenner: yes you should remove the default implementation and add a dimension.

operators/src/engine/mod.rs
```diff
@@ -77,3 +77,46 @@ macro_rules! call_on_generic_raster_processor {
         }
     };
 }
+
+/// Calls a function on two `TypedRQueryProcessor`s by calling it on their variant combination.
+/// Call via `call_bi_generic_processor!(input_a, input_b, (processor_a, processor_b) => function)`.
+#[macro_export]
+macro_rules! call_bi_generic_processor {
```
jdroenner, line 84: i'am still not conviced that this will scale up for more then 3 inputs. We should really discuss why type conversion as adapter step for n >=4 inputs is such a problem if we do projection, scaling and other expensive operations anyway.

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
jdroenner: what about a sum per attribute?

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+}
+
+/// Specifies in and output types of CL program and compiles the source into a reusable `CompiledCLProgram`
+pub struct CLProgram {
+    input_rasters: Vec<RasterArgument>,
+    output_rasters: Vec<RasterArgument>,
+    input_features: Vec<VectorArgument>,
+    output_features: Vec<VectorArgument>,
+    iteration_type: IterationType,
+}
```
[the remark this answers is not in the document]
jdroenner: what about other arguments? A scale or offset factor for example?

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+	ushort crs_code;
+	ushort has_no_data;
+} RasterInfo;
+
+#define R(t,x,y) t ## _data[y * t ## _info->size[0] + x]
+"####,
+        );
+
+        for (idx, raster) in self.input_rasters.iter().enumerate() {
+            s += &format!("typedef {} IN_TYPE{};\n", raster.data_type.ocl_type(), idx);
```
jdroenner, line 139: can we move this to a header file and include it in the kernel?

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+            .zip(self.output_feature_types.iter())
+            .enumerate()
+        {
+            let features = builder.as_ref().expect("checked");
+
+            let geo_buffers = if argument.include_geo {
+                Some(match features.output_type {
+                    VectorDataType::MultiPoint => {
+                        let coords = Buffer::<Coordinate2D>::builder()
+                            .queue(kernel.default_queue().expect("expect").clone())
```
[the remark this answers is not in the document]
jdroenner: why can default_queue() fail?

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+
+    fn set_raster_arguments(&mut self, kernel: &Kernel) -> Result<()> {
+        ensure!(
+            self.input_rasters.iter().all(Option::is_some),
+            error::CLProgramUnspecifiedRaster
+        );
+
+        for (idx, raster) in self.input_rasters.iter().enumerate() {
+            let raster = raster.expect("checked");
+            call_generic_raster2d!(raster, raster => {
```
jdroenner, line 642: why use a macro here? i think this is something a trait method could handle very well..

operators/src/processing/expression.rs
jdroenner: if there is no nodata set then its just some number?

operators/src/processing/expression.rs
```diff
@@ -0,0 +1,361 @@
...
+        // TODO: validate that tiles actually fit together, otherwise we need a different approach
+        let mut cl_program = self.cl_program.clone();
+        self.source_a
+            .query(query, ctx)
+            .zip(self.source_b.query(query, ctx))
+            .map(move |(a, b)| match (a, b) {
+                (Ok(a), Ok(b)) => {
+                    let mut out = Raster2D::new(
+                        *a.dimension(),
+                        vec![TO::zero(); a.data.data_container.len()], // TODO: correct output size; initialization required?
```
jdroenner, line 207: yes. we want to initialize them. If this is really a bottleneck then we can change this later

jdroenner:
there are multiple things which we should discuss:
- is the opencl code only for expression?
- if not, what about operations which require overlapping on tile borders? Will this be possible with this approach?
- if not, what about numeric parameters? How will i get them to a kernel?
- Is the macro approach really scalable? Why not introduce data type changing adapters? they will never be more expensive as any other operator..

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: this would either require a 2d-kernel or custom kernel code that iterates over features but computes all the sums at once for each feature. but what would you do with the ouput? sounds like a plot output to me and goes beyond the scope of this PR I feel

datatypes/src/collections/multi_point_collection.rs
michaelmattig: can that happen here though?

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: adding output slots fors plots should be possible. I am not sure what plots should be computed on the gpu directly. One can always compute a output feature collections with the desired data and serialize them on the cpu. Or am I missing something?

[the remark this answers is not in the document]
michaelmattig: Scalar arguments sound like a good idea. For now they could however also be hardcoded in the opencl source code. This limits the reusablity of course

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+    _coords: Buffer<Coordinate2D>,
+    _ring_offsets: Buffer<i32>,
+    _polygon_offsets: Buffer<i32>,
+    _feature_offets: Buffer<i32>,
+}
+
+struct ColumnBuffer<T: OclPrm> {
+    column_name: String,
+    values: Buffer<T>,
+    nulls: Option<Buffer<i32>>, // OpenCl does not support bool for host device transfer
```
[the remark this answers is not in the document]
michaelmattig: I am not sure if this is a good idea. I think I read somewhere that bitwise operations are not advised on GPUs for performance reasons.

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: if there is no default_queue set it returns None

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+        nulls: Option<&[bool]>,
+    ) -> Result<()> {
+        let buffer = Buffer::<T>::builder()
+            .queue(kernel.default_queue().expect("expect").clone())
+            .len(len)
+            .copy_host_slice(data)
+            .build()?;
+        kernel.set_arg(format!("IN_POINT{}_COLUMN_{}", idx, column), &buffer)?;
+
+        if let Some(nulls) = nulls {
```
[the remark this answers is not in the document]
michaelmattig: the opencl kernel needs to know that there are no nulls. If I write an operator and require a column to be non-nullable then I don't need to transfer nulls to the device and the kernel knows that all values are non-null.

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+                raster.dimension().size_of_y_axis().as_(),
+                1, // TODO
+            ],
+            origin: [0., 0., 0.],
+            scale: [0., 0., 0.],
+            min: 0.,
+            max: 0.,
+            no_data: raster.no_data_value.map_or(0., AsPrimitive::as_),
+            crs_code: 0,
+            has_no_data: u16::from(raster.no_data_value.is_some()),
```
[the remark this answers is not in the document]
michaelmattig: the u16 is what we used in mapping. You are right we should'nt mix types. Which one should I use?

datatypes/src/primitives/coordinate.rs
michaelmattig: it's a marker trait that signals data of this type can be transferred to the device. Only works for repr-c types

datatypes/src/primitives/coordinate.rs
michaelmattig: this is necessary to be able to copy coordinates to opencl device. the OclPrm marker trait requires data to impl Default

datatypes/src/collections/feature_collection.rs
michaelmattig: or make the column_type optional?

datatypes/src/collections/batch_builder.rs
michaelmattig: maybe we can call it ArrowBufferBuilder or something like that. It's not necessarily only meant for OpenCl, but for cases where the arrow buffers are built externally and we want to avoid copies. Maybe we can also use this for Python integration or results computed on a different machine.

I don't think wrapping the buffers makes any sense because nothing of it works without arrow ;)

operators/src/processing/expression.rs
michaelmattig: because that would require escaping all the curly braces and make the code worse the read

operators/src/processing/expression.rs
michaelmattig: why do you think so? Mapping doesn't do this.

operators/src/processing/expression.rs
michaelmattig: I guess the expression operator should check that if at least one input has a no data value, then the output data needs to have a no data value defined. If not, the operator should throw an error. What do you think?

datatypes/src/raster/typed_raster.rs
michaelmattig: they are sorted first by type i,u,f and then by size

datatypes/src/raster/typed_raster.rs
michaelmattig: It's N dim with N=2, or do you mean I should use TypedRaster2D?

operators/src/opencl/cl_program.rs
michaelmattig: this is dynamically generated so I can't read it from a file. Also: I already include the source codes for the kernel and the typedes separately:
```
let program = ProgramBuilder::new()
            .src(typedefs)
            .src(source)
            .build(&ctx)?;
```

operators/src/engine/mod.rs
michaelmattig: we should discuss this independently of this PR

operators/src/opencl/cl_program.rs
michaelmattig: I need to create a `Buffer<T>` of the correct type and access to the rasters data container. I am open to suggestions on how to accept rasters as traits with a mixture of different T's.

operators/src/processing/expression.rs
michaelmattig: the kernel is executed for each kernel anway so the inizialization may be redundant but yes, we can investigate it later.

michaelmattig:
> is the opencl code only for expression?

no, then we wouldn't need support for feature collections. it's meant as generic mechanism for transfering data for opencl processing s.t. not every operators need to performt the mapping of data

> if not, what about operations which require overlapping on tile borders? Will this be possible with this approach?

handling overlapping tiles is responsibility of the opencl kernel code and the operator that feeds the data into the cl program. I don't see any inherent limitations in my code as It's solely responsible for encapsulating the transfer to and from the opencl device

> if not, what about numeric parameters? How will i get them to a kernel?

For now, they can be hardcoded into the kernel which of course limits reusablity. I think it makes sense to add scalar arguments to the CLProgram. But I think we can do it later when we need it.

> Is the macro approach really scalable? Why not introduce data type changing adapters? They will never be more expensive as any other operator..

The scalability is indeed limited. We should be careful with data type conversion, because otherwise we could have saved ourselves from having such a great type variety in the first and only support one type of rasters. Either way I think this is independent of the main focus of this PR which is the open cl interface.

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,578 @@
...
+    output_rasters: Vec<Option<&'a mut TypedRaster2D>>,
+}
+
+impl<'a> CLProgramParameters<'a> {
+    fn new(
+        input_raster_types: Vec<RasterDataType>,
+        output_raster_types: Vec<RasterDataType>,
+    ) -> Self {
+        let mut v = Vec::new();
+        v.resize_with(output_raster_types.len(), || None);
```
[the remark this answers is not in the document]
michaelmattig: somehow it doesn't work like
> error[E0277]: the trait bound `&mut geoengine_datatypes::raster::typed_raster::TypedRasterNDim<geoengine_datatypes::raster::Dim<[usize; 2]>>: std::clone::Clone` is not satisfied
    --> operators/src/opencl/cl_program.rs:286:29
     |
286  |             output_rasters: vec![None; output_raster_types.len()],
     |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::clone::Clone` is not implemented for `&mut geoengine_datatypes::raster::typed_raster::TypedRasterNDim<geoengine_datatypes::raster::Dim<[usize; 2]>>`

not sure why it can't clone `None`...

datatypes/src/collections/multi_point_collection.rs
michaelmattig: a trait would be possible, but is out of scope of this PR. Not sure where we would need direct access to the coordinates of a multipolygoncollection eg. without infos about the rings, other than in tests.

datatypes/src/collections/data_types.rs
michaelmattig: please elaborate how this would help here

datatypes/src/raster/helpers.rs
michaelmattig: unsure but not the focus of this PR

datatypes/src/raster/typed_raster.rs
michaelmattig: It's N dim with N=2, or do you mean I should use TypedRaster2D?

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,578 @@
...
+use snafu::ResultExt;
+
+/// Whether the kernel iterates over pixels or features
+#[derive(PartialEq, Clone, Copy, Debug)]
+pub enum IterationType {
+    Raster,
+    Vector,
+}
+
+/// Specifies in and output types of CL program and compiles the source into a reusable `CompiledCLProgram`
```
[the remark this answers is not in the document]
michaelmattig: 91ef24f5b869f75563586b32b932afa0a27fce0f

operators/src/opencl/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod cl_program;
```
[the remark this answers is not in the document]
michaelmattig: 91ef24f5b869f75563586b32b932afa0a27fce0f

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+/// Whether the kernel iterates over pixels or features
+#[derive(PartialEq, Clone, Copy, Debug)]
+pub enum IterationType {
+    Raster,            // 2D Kernel, width x height
+    VectorFeatures,    // 1d kernel width = number of features
+    VectorCoordinates, // 1d kernel width = number of coordinates
+}
+
+#[derive(PartialEq, Clone, Copy, Debug)]
+pub struct RasterArgument {
```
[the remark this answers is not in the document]
michaelmattig: 91ef24f5b869f75563586b32b932afa0a27fce0f

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+    }
+
+    pub fn add_output_features(&mut self, vector_type: VectorArgument) {
+        self.output_features.push(vector_type);
+    }
+
+    fn create_type_definitions(&self) -> String {
+        let mut s = String::new();
+
+        if self.input_rasters.len() + self.output_rasters.len() == 0 {
```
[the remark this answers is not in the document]
michaelmattig: 91ef24f5b869f75563586b32b932afa0a27fce0f

datatypes/src/primitives/coordinate.rs
michaelmattig: 91ef24f5b869f75563586b32b932afa0a27fce0f

operators/src/processing/expression.rs
michaelmattig: 91ef24f5b869f75563586b32b932afa0a27fce0f

operators/src/processing/expression.rs
michaelmattig: 91ef24f5b869f75563586b32b932afa0a27fce0f

datatypes/src/raster/data_type.rs
michaelmattig: 91ef24f5b869f75563586b32b932afa0a27fce0f

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+
+        for (idx, raster) in self.output_rasters.iter().enumerate() {
+            s += &format!("typedef {} OUT_TYPE{};\n", raster.data_type.ocl_type(), idx);
+        }
+
+        s
+    }
+
+    pub fn compile(self, source: &str, kernel_name: &str) -> Result<CompiledCLProgram> {
+        ensure!(
```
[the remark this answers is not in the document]
michaelmattig: 88017ba0b9c44beb15047cbe7ce710c819e53541

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+impl RasterArgument {
+    pub fn new(data_type: RasterDataType) -> Self {
+        Self { data_type }
+    }
+}
+
+#[derive(PartialEq, Clone, Debug)]
+pub struct VectorArgument {
+    pub vector_type: VectorDataType,
+    // TODO: merge columns and types into one type
```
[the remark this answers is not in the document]
michaelmattig: bad20b7bb06f7781905c0e1d4486bb158c4d1dd3

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+        output_raster_types: Vec<RasterArgument>,
+        input_feature_types: Vec<VectorArgument>,
+        output_feature_types: Vec<VectorArgument>,
+    ) -> Self {
+        let mut output_rasters = Vec::new();
+        output_rasters.resize_with(output_raster_types.len(), || None);
+
+        let mut output_features = Vec::new();
+        output_features.resize_with(output_feature_types.len(), || None);
+
```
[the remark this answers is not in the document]
michaelmattig: doesnt work because None is not Clone

operators/src/opencl/cl_program.rs
```diff
@@ -0,0 +1,1743 @@
...
+                        let buffer = Buffer::builder()
+                            .queue(kernel.default_queue().expect("expect").clone())
+                            .len(coordinates_offsets.len())
+                            .copy_host_slice(coordinates_offsets)
+                            .build()?;
+
+                        kernel.set_arg(format!("IN_POINT_OFFSETS{}", idx), &buffer)?;
+                    }
+                    TypedFeatureCollection::MultiLineString(_)
+                    | TypedFeatureCollection::MultiPolygon(_) => todo!(), // TODO: lines, polygons
```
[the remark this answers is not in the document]
michaelmattig: Then it won't be listed under TODOs in your IDE though.

datatypes/src/collections/data_types.rs
michaelmattig: 0c148c7dc15557c77a7dd6d7f9cd11c43545bfcd

datatypes/src/raster/typed_raster.rs
michaelmattig: 0c148c7dc15557c77a7dd6d7f9cd11c43545bfcd I did not add a dimension though as we do not support any operations on 3d rasters anyway

datatypes/src/collections/feature_collection_builder.rs
michaelmattig: 3b07edb26b881591ae01ad68e7d7132d3d436635

datatypes/src/collections/batch_builder.rs
michaelmattig: 72ac38a30e27628e999209f40b1d6d8c20a95b0f

datatypes/src/raster/typed_raster.rs
michaelmattig: 0c148c7dc15557c77a7dd6d7f9cd11c43545bfcd I stuck with 2d for now

datatypes/src/raster/typed_raster.rs
michaelmattig: I stuck with 2d for now 0c148c7dc15557c77a7dd6d7f9cd11c43545bfcd

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Your Option would have to be clonable.

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Unfortunately, still, it is either 
```
[…] output types of of a cl program
``` 
or or 
```
[…] output types of `ClProgram`
```
but not _output types of CL program_.

datatypes/src/collections/multi_point_collection.rs
ChristianBeilschmidt: I guess that depends if we ever will use the `slice` functionality of arrow to slice a collection.

datatypes/src/collections/multi_point_collection.rs
ChristianBeilschmidt: For the OpenCL implementation of lines and polygons that is not part of this PR ;-).

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: But for performance reasons it is I guess also not wise to create boolean vectors for each attribute I guess all the time.
Do we have resources for that?

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: grafik

When you created that code using IntelliJ Rust it actually did.
What about VSCode?

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Unfortunately, the size of a boolean is not defined in the OpenCL spec, this is also why it can't be transferred to the device. So we can't just use the same size here.

There are other approaches:
* What is the size of bool on our devices?
* What aligns better with the memory we use, e.g., if we use only 32 bit types…?
* Is it better to use the smallest type to keep the transfer low? Then we should use 1 byte.

datatypes/src/primitives/coordinate.rs
ChristianBeilschmidt: Hmm, I guess it is not harmful to have this `Default` then. @jdroenner what is your opinion on that?

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: This is an option. However, it seems quite strange to get all column names by calling `collection.column_names(None)`.

datatypes/src/collections/batch_builder.rs
ChristianBeilschmidt: This is valid. It is like an builder from internals, raw parts, or something like that.

operators/src/processing/expression.rs
ChristianBeilschmidt: You did not change this part in this commit.

operators/src/processing/expression.rs
ChristianBeilschmidt: Mapping had the problem to not being able to use rasters without NODATA since it does, what Johannes suggests, just set it to some number. If the range is exhausted, e.g. for a byte raster, this becomes a problem.
So either we disallow rasters without NODATA at all or we implement Michaels approach.

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
jdroenner: i thing it is also a tuple of some sort. How does postgres handle this?

operators/src/opencl/cl_program.rs
jdroenner: yes. no macro needed here.

datatypes/src/collections/multi_point_collection.rs
jdroenner: then this should be addressed later.

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
jdroenner: i also think a plot should be generated by some sort of sink as a last operation from "real" data.

operators/src/processing/expression.rs
jdroenner: then just force a NoData value and increase the datatype if needed. Error if no nodata is possible is ok i guess...

datatypes/src/collections/data_types.rs
jdroenner: just an idea.

datatypes/src/raster/helpers.rs
jdroenner: then we will discuss this as a different issue.

datatypes/src/raster/typed_raster.rs
jdroenner: why can't you use a generic Dimension?

operators/src/engine/mod.rs
jdroenner: ok

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: what does postgres have to do with opencl?

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: Option of &mut can't be clone I guess

datatypes/src/collections/multi_point_collection.rs
michaelmattig: Currently there are two ways to build a feature collection: RowBuilder and BatchBuilder. In both cases the buffers are not built by slicing. Maybe you can design a test case that fails for the current implementation. Or: we leave it for now.

operators/src/processing/expression.rs
michaelmattig: right now this value is 0. I think it would be better to force the output to have nodata if at least one input has no data. but right now we cant check if the input has no data before actually getting the first tile.

datatypes/src/raster/typed_raster.rs
michaelmattig: I could but the rest of the code only uses 2d rasters

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: 7ed0dcf6d9b9b5f6006b1f81b6d0fb44bb7a81c4

operators/src/processing/expression.rs
michaelmattig: 7ed0dcf6d9b9b5f6006b1f81b6d0fb44bb7a81c4

operators/src/processing/expression.rs
ChristianBeilschmidt: Trying to create NODATA is okay I guess but isn't it strange to magically change the data type?

michaelmattig: Currently our rasters only have a data type and not min/max values right? Mapping uses the unit info about min/max to insert a no data value. we are not there yet. I am also against automatically increasing the data type. The user should say the desired output data type and the system should tell him to set a no data value if needed. then she can decide whether to choose a bigger data type to get a value. But this cannot be solved right now because of the lack of meta info in raster data and also has nothing to do with opencl

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
jdroenner: was ich meinte war: eine Summe über eine Spalte muss man ja irgendwie abbilden. Ich denke nicht, das z.b. Postgres eienn Typ "Plot" hat. Deshalb war meine Frage wie man so etwas abbilden würde. Es hat nichts mit OpenCL zutun...

operators/src/processing/expression.rs
jdroenner: yes magic type changes should be avoided

datatypes/src/raster/typed_raster.rs
jdroenner: why?

datatypes/src/raster/typed_raster.rs
jdroenner: well then it is ready for more dimensions in the future

operators/src/opencl/cl_program.rs
jdroenner: well you could use a Trait with a method into_buffer and then implement it for Raster<T> and the TypedRaster. Then you can just call it and you will get the correct type.

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: you can computate aggregates over the whole table of via `GROUP BY` and get a new table with the sum as a new column. we could return a new DataCollection in this case

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: > Do we have resources for that?

computational resources or information on the cost of creating a i32 vector out of a boolean vector? I could create a benchmark but my guess is the overhead is neglible

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: VSCode doesn't show it

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: yeah I really have no idea why you would use u16. I agree either use smallest (u8) or best suited for access, but I am not sure which that is. Should I run some benchmarks?

operators/src/opencl/cl_program.rs
michaelmattig: but then I would need a large match block in the impl for `TypedRaster`. We explicitly added the macro to avoid the match blocks.

datatypes/src/collections/feature_collection.rs
michaelmattig: ec9ee747f1e86433c9466fe506e5ddf216651827

datatypes/src/collections/batch_builder.rs
michaelmattig: ec9ee747f1e86433c9466fe506e5ddf216651827

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: No, creating boolean vectors out of bitmasks.

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: It would be nice, otherwise, just use a type from an educated guess.

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: I pass it directly now f67bb462d19bffb7e5471846484e1c1c577bb33e

operators/src/opencl/cl_program.rs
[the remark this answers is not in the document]
michaelmattig: I use i8 consistently now d11f857f76f5c874662f622748dea347aaa954a4

operators/src/processing/expression.rs
michaelmattig: no data is now mandatory 3fe3ae0f1429bfac027c6091a1b6c8ac2638272c

datatypes/src/raster/typed_raster.rs
michaelmattig: i generified it 1f4f5dd019bdf919624fbc059f71b396c701db6d

datatypes/src/raster/typed_raster.rs
michaelmattig: I generified it 1f4f5dd019bdf919624fbc059f71b396c701db6d

operators/src/opencl/cl_program.rs
michaelmattig: I tried the trait method but it led to much more code and still couldn't avoid the match-macro. We should leave it as it is for now

datatypes/src/raster/typed_raster.rs
jdroenner: ok.  i think we can work on the raster traits/types as a separate task.

ChristianBeilschmidt (approved):

PR #35, opened 2020-09-22
rename Projection* to SpatialReference*

jdroenner:


datatypes/src/spatial_reference.rs
```diff
@@ -127,36 +127,36 @@ impl FromStr for Projection {
 }
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
-pub enum ProjectionOption {
-    Projection(Projection),
+pub enum SpatialReferenceOption {
+    Projection(SpatialReference),
```
ChristianBeilschmidt, line 131:
```suggestion
    SpatialReference(SpatialReference),
```

datatypes/src/lib.rs
```diff
@@ -24,6 +24,6 @@ pub mod error;
 pub mod operations;
 pub mod plots;
 pub mod primitives;
-pub mod projection;
 pub mod raster;
+pub mod spatial_reference;
```
ChristianBeilschmidt, line 28: Will there be more functionality in `spatial_reference`? Otherwise we could say
```
mod spatial_reference;

pub use self::spatial_reference::{SpatialReference, SpatialReferenceAuthority, SpatialReferenceOption};
```

ChristianBeilschmidt (changes requested):
You missed on name ;-).

datatypes/src/lib.rs
jdroenner: maybe.... i think we will see when projection operations are added.

ChristianBeilschmidt (approved):

PR #38, opened 2020-09-28
refactor service parameters as context

michaelmattig:
makes service parameters more uniform and more easily extensible

### DIFF
--- a/services/src/handlers/mod.rs
+++ b/services/src/handlers/mod.rs
@@ -7,9 +7,10 @@ use crate::users::hashmap_userdb::HashMapUserDB;
 use crate::users::session::{Session, SessionToken};
 use crate::users::userdb::UserDB;
 use crate::workflows::registry::{HashMapRegistry, WorkflowRegistry};
+use async_trait::async_trait;
 use std::str::FromStr;
 use std::sync::Arc;
-use tokio::sync::RwLock;
+use tokio::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 use warp::Filter;
 use warp::{Rejection, Reply};
 
@@ -43,9 +44,12 @@ fn authenticate<C: Context>(
 ) -> impl warp::Filter<Extract = (C,), Error = warp::Rejection> + Clone {
     async fn do_authenticate<C: Context>(mut ctx: C, token: String) -> Result<C, warp::Rejection> {
         let token = SessionToken::from_str(&token).map_err(|_| warp::reject())?;
-        let user_db = ctx.user_db();
-        let db = user_db.read().await;
-        ctx.set_session(db.session(token).map_err(|_| warp::reject())?);
+        let session = ctx
+            .user_db_ref()
+            .await
+            .session(token)
+            .map_err(|_| warp::reject())?;
+        ctx.set_session(session);
         Ok(ctx)
     }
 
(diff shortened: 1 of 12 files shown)


services/src/handlers/mod.rs
```diff
@@ -32,20 +38,68 @@ pub async fn handle_rejection(error: Rejection) -> Result<impl Reply, Rejection>
...
+        self.user_db.clone()
+    }
+
+    fn project_db(&self) -> DB<Self::ProjectDB> {
+        self.project_db.clone()
+    }
+
+    fn workflow_registry(&self) -> DB<Self::WorkflowRegistry> {
+        self.workflow_registry.clone()
+    }
```
ChristianBeilschmidt, line 94: Is it necessary to always clone it or to hand out a reference for some cases?

services/src/handlers/mod.rs
```diff
@@ -32,20 +38,68 @@ pub async fn handle_rejection(error: Rejection) -> Result<impl Reply, Rejection>
...
     }
 
     warp::any()
-        .and(warp::any().map(move || Arc::clone(&user_db)))
+        .and(warp::any().map(move || ctx.clone()))
         .and(warp::header::<String>("authorization"))
         .and_then(do_authenticate)
 }
+
+pub trait Context: 'static + Send + Sync + Clone {
```
ChristianBeilschmidt, line 58: I recommend writing some doc-comments.

services/src/handlers/mod.rs
```diff
@@ -32,20 +38,68 @@ pub async fn handle_rejection(error: Rejection) -> Result<impl Reply, Rejection>
...
+    fn user_db(&self) -> DB<Self::UserDB>;
+    fn project_db(&self) -> DB<Self::ProjectDB>;
+    fn workflow_registry(&self) -> DB<Self::WorkflowRegistry>;
+    fn session(&self) -> Result<&Session>;
+
+    fn set_session(&mut self, session: Session);
+}
+
+#[derive(Clone, Default)]
+pub struct DefaultContext {
```
ChristianBeilschmidt, line 72:
```suggestion
pub struct InMemoryContext {
```

ChristianBeilschmidt (changes requested):

services/src/handlers/mod.rs
michaelmattig: Arc is already kind of a reference. Does it really make sense to hand out a reference to a reference? Maybe it makes more sense to (additionally) provide a method that directly returns the contained Mutex or even the Read/Write guard. Latter one would introduce two methods for each contained value for read/write...

services/src/handlers/mod.rs
michaelmattig: 3a6bd99fc4447b0eaec36c53dfa09c76ec330398

services/src/handlers/mod.rs
michaelmattig: 3a6bd99fc4447b0eaec36c53dfa09c76ec330398

services/src/handlers/mod.rs
ChristianBeilschmidt: I see lots of calls where it is not necessary to clone it.
So you could hand out some read guard with `project_db_ref()` and write guard with `project_db_mut()`.

michaelmattig: c6abcb51f4bd8e4afa329878a50068a4b73ec59f

PR #39, opened 2020-09-29
towards postgres backend for the dbs

michaelmattig:


services/src/error.rs
```diff
@@ -60,6 +60,12 @@ pub enum Error {
     InvalidWFSTypeNames,
 
     NoWorkflowForGivenId,
+
+    BB8,
```
ChristianBeilschmidt, line 64: Does it have a source error?

services/src/users/postgres_userdb.rs
```diff
@@ -0,0 +1,175 @@
...
+use async_trait::async_trait;
+use bb8_postgres::bb8::Pool;
+use bb8_postgres::tokio_postgres::error::SqlState;
+use bb8_postgres::tokio_postgres::NoTls;
+use bb8_postgres::PostgresConnectionManager;
+use snafu::ResultExt;
+
+struct PostgresUserDB {
+    conn_pool: Pool<PostgresConnectionManager<NoTls>>, // TODO: support Tls connection as well
+}
```
ChristianBeilschmidt, line 18: Maybe `LocalPostgresUserDB`, because Tls/NoTls is a generic?

services/src/users/postgres_userdb.rs
```diff
@@ -0,0 +1,175 @@
...
+impl PostgresUserDB {
+    async fn new(conn_pool: Pool<PostgresConnectionManager<NoTls>>) -> Result<Self> {
+        let a = Self { conn_pool };
+        a.update_schema().await?;
+        Ok(a)
+    }
+
+    async fn schema_version(&self) -> Result<i32> {
+        // TODO: handle if version table not exists
+        let conn = self.conn_pool.get().await.map_err(|_| error::Error::BB8)?; // TODO: error reporting
```
ChristianBeilschmidt, line 29: You could put the error as source into the custom error type.

services/src/users/postgres_userdb.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        let stmt = match conn.prepare("SELECT version from version").await {
+            Ok(stmt) => stmt,
+            Err(e) => {
+                if let Some(code) = e.code() {
+                    if *code == SqlState::UNDEFINED_TABLE {
+                        // TODO: log
+                        println!("UserDB: Uninitialized schema");
```
ChristianBeilschmidt, line 37: Doesn't Johannes log output already in his gdal source?

services/src/users/postgres_userdb.rs
```diff
@@ -0,0 +1,175 @@
...
+                        );\
+                        INSERT INTO version VALUES (1);\
+
+                        CREATE TABLE users (
+                            id UUID PRIMARY KEY,
+                            email character varying (256) UNIQUE,
+                            password_hash character varying (256),
+                            real_name character varying (256)
+                        );
+                        ",
```
ChristianBeilschmidt, line 73: Looks like a reasonable approach, maybe can put the instructions into *.sql files and include them here with `include_str!`.

services/src/users/postgres_userdb.rs
```diff
@@ -0,0 +1,175 @@
...
+    use crate::util::user_input::UserInput;
+    use bb8_postgres::bb8::Pool;
+    use bb8_postgres::{tokio_postgres, PostgresConnectionManager};
+    use std::str::FromStr;
+
+    #[tokio::test]
+    async fn test() {
+        // TODO: load from test config
+        // TODO: add postgres to ci
+        // TODO: clear database
```
ChristianBeilschmidt, line 154: Do we start Postgres here or do we mock the returned commands?

ChristianBeilschmidt (commented):
Some cents to start with.

services/src/users/postgres_userdb.rs
michaelmattig: couldn't find it

services/src/users/postgres_userdb.rs
michaelmattig: what's the advantage?

services/src/users/postgres_userdb.rs
michaelmattig: maybe we don't test it at all. We have the In-Memory stores to test the rest of service for that reason that it's difficult to test with databases

services/src/users/postgres_userdb.rs
ChristianBeilschmidt: Syntax-highlighting and the chance to execute them directly more easily.
Otherwise, I would at least use raw strings instead of escaping all lines, etc,

services/src/users/postgres_userdb.rs
ChristianBeilschmidt: Can't we test it with a postgres configuration but only execute them with a special flag? We need some chance to test it I guess.

services/src/users/postgres_userdb.rs
ChristianBeilschmidt: Okay, can be a new task. But it should be `eprintln!`.

services/src/users/postgres_userdb.rs
michaelmattig: currently the tests are ignore, the CI could also include ignore tests. However the more severe problem is that the rust test framework doesn't allow configuring running inidividual tests non-concurrently and also there is no way for setUp and tear down of tasks. So all tests would have to be in a single test function that creates the database in the beginning and drops it at the end. Right now I am testing user and project db separately but you have to clear the database manually and don't run them concurrently.

services/src/users/postgres_userdb.rs
michaelmattig: Future schema changes might be only single lines. creating separate files doesn't scale well. The new line escapes are no longer there.

services/src/users/postgres_userdb.rs
michaelmattig: why? it's not an error.

ChristianBeilschmidt: https://en.wikipedia.org/wiki/Standard_streams

`stderr` is not only about errors but also about status and diagnostics. For instance, if you have a program that converts A to B, then you don't want to print the loading status to stdout since it is expected to be the result of the program.

michaelmattig: da6a0baf66b03f687446b12cf4e672b69836995d

services/src/users/postgres_userdb.rs
ChristianBeilschmidt: I hope that schema changes are not a scalability issue :o. 

You could take a final look at diesel https://diesel.rs/guides/getting-started/, which has migration scripts in files, and then we conclude this comment.

services/src/users/postgres_userdb.rs
ChristianBeilschmidt: Here is a quick workaround for setup and tear down:
https://medium.com/@ericdreichert/test-setup-and-teardown-in-rust-without-a-framework-ba32d97aa5ab

Maybe we can take a look at Diesel here how to do tests against PostgreSQL:
https://github.com/diesel-rs/diesel/blob/master/diesel/src/test_helpers.rs

services/src/users/postgres_userdb.rs
michaelmattig: I meant that its more easy to keep lots of small schema changes in one file than having them all in individual files. We can change this easily later on anyway. For now I vote for keeping the statments in the source

services/src/users/postgres_userdb.rs
michaelmattig: neat workaround. there is still the issue where tests may not run concurrently. Otherwise one test deletes the data that another test wants to check exists. We would need to have database transactions spanning multiple calls of our user/projectdb functions, but that's not possible right now and not necessary in practice. If we had this each test could just roll back the transaction at the end, leaving the database empty and other transactions undisturbed.

I don't think diesel solved this problem as they most likely test individual functionalities and not correct modifications of data stored in tables.

ChristianBeilschmidt: Tests have to be complete and not depend on other tests.
If you want to have something in order, you should make one test out of it (that could call different methods).

You could lazy_static your instance in the test module and then acquire a mutex in the tests.

michaelmattig: its hard to make them independend if they all use the same data base. putting everything into one test seems more feasible. could we create a new task called testing databases where we tackle setting up postgres in the ci and combinding the tests. I would like to get this PR merged to continue working on other api changes.

ChristianBeilschmidt: A task it is.

PR #40, opened 2020-10-07
config support 2nd try

michaelmattig:


services/src/util/config.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub list_limit: u32,
+}
+
+impl ConfigElement for ProjectService {
+    const KEY: &'static str = "project_service";
+}
+
+#[derive(Debug, Deserialize)]
+pub struct GdalSource {
+    pub raster_data_root_path: String, // TODO: Path?
```
jdroenner, line 108: yes paths everywhere

services/src/util/config.rs
```diff
@@ -0,0 +1,113 @@
...
+            .map(File::from)
+            .collect();
+
+        settings.merge(files).unwrap();
+
+        settings
+    });
+}
+
+pub fn get_config<'a, T>(key: &str) -> Result<T>
```
jdroenner, line 40: will this de-serialize the config on each call?

jdroenner (commented):
looks good so far. How does config handle get() calls? Will it always de-serialize?

services/src/util/config.rs
michaelmattig: good point. Yes it deserializes on each call. Fortunately, it doesn't read the file each time ;)

It also only deserializes lazily once the config key is accessed. So we might end up with the problem, where a faulty configuration is not detected until the value is actually used for the first time.

Maybe we should just bundle all the configs in a Config struct and deserialize into it from the config file once at startup. Then we at least know that all required values are present and valid. I am not sure if we even need the config crate in this case. It makes it at least easy to merge multiple config files like Settings-default.toml and Settings.toml

services/src/util/config.rs
michaelmattig: I could use `PathBuf` as `Path` is not owned. However it doesn't seems to check if it's really a valid path in the file system. So makes no differences to using string

services/src/util/config.rs
jdroenner: We could deserialize into a typed map/list. Then we can keep the get<T> thing. I also think that a validation of the config when it is loaded is important.

services/src/util/config.rs
jdroenner: i think PathBuf is fine since String is also an owned type.

services/src/util/config.rs
michaelmattig: but what ist the type of the value of the map? `Map<String, ?>`. Do we put all config structs in an enum?

jdroenner: we will have to use a crate like https://docs.rs/typemap/0.3.3/typemap/ or build something similar

services/src/util/config.rs
michaelmattig: e44b19d69a1fa7d2452c8255a59f491b6b6353c3

services/src/util/config.rs
michaelmattig: I created a task for this https://dbs-projects.mathematik.uni-marburg.de/T1227

PR #41, opened 2020-10-08
T1177 CsvSource: Use spawn_blocking instead of spawn

ChristianBeilschmidt:


operators/src/source/csv.rs
```diff
@@ -131,13 +133,10 @@ pub struct CsvSourceStream {
     parameters: CsvSourceParameters,
     bbox: BoundingBox2D,
     chunk_size: usize,
-    state: Arc<Mutex<CsvSourceStreamState>>,
-}
-
-pub struct CsvSourceStreamState {
-    csv_reader: ReaderState,
+    reader_state: Arc<Mutex<ReaderState>>,
+    thread_is_computing: Arc<AtomicBool>,
```
michaelmattig, line 137: can't you use AtomicBool directly?

operators/src/source/csv.rs
```diff
@@ -286,26 +284,37 @@ impl Stream for CsvSourceStream {
     type Item = Result<MultiPointCollection>;
 
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        let mut state = self.state.lock().unwrap(); // TODO: handle lock poisoning
-        if state.poll_result.is_some() {
-            let x = state.poll_result.take().unwrap();
+        // TODO: handle lock on multiple occasions
```
michaelmattig, line 287: what does this comment mean?

operators/src/source/csv.rs
```diff
@@ -286,26 +284,37 @@ impl Stream for CsvSourceStream {
...
+
+        if self.thread_is_computing.load(Ordering::Relaxed) {
```
michaelmattig, line 289: why is this extra indicator needed? You could just check poll result being `Some` like before.

operators/src/source/csv.rs
ChristianBeilschmidt: No, things that I put into the thread must be `'static`.

operators/src/source/csv.rs
ChristianBeilschmidt:
```suggestion
        // TODO: handle lock poisoning on multiple occasions
```

operators/src/source/csv.rs
ChristianBeilschmidt: No, because if you call `poll_next` for the first time, it is `None` and you have no clue that you should start the thread.
On the other hand, if you already started the thread, but did not wait for the notification to arrive, you want to get pending again and not start the thread a second time. I introduced this, since moving to `spawn_blocking` lead to test errors based on race conditions.

michaelmattig: wouldn't it make more sense to have a bool variable like `init` to check if it's the first run instead of updating this variable all the time?

ChristianBeilschmidt: Well, then you have the problem a second and a third time.

* first poll, set this variable to true and start thread
* second (intermediate) poll, the result is None, so return pending
* thread is finished and calls waker
* fourth poll, you return Some(thing), leaving None in-place
* fifth poll, now you would like to start a new work thread, but you see that there is None in the variable, so you think there is already a thread.
* …

operators/src/source/csv.rs
ChristianBeilschmidt: 7723119

PR #42, opened 2020-10-08
Project versions

michaelmattig:
I removed the check that versions must have a greater timestamp than the previous one because it causes tests to fail on my maschine because creation and update seem to happen at the same system time. The alternative would be to `sleep` shortly. However, as projects can be shared it's always possible two modifications happen at the same timestamp and that's ok as we shouldn't aim to solve phantom update problem here, as long as 
- the history is not lost
- there is only one true latest version of a project

this is both guaranteed for the postgres version

Another item to address: The `ProjectVersion` struct gets the local current time on `new` but the postgres item doesn't use this value but rather states `CURRENT_TIMESTAMP` and uses the database`s clock.
See:  https://github.com/geo-engine/geoengine/blob/versioning/services/src/projects/project.rs#L330 and https://github.com/geo-engine/geoengine/blob/versioning/services/src/projects/postgres_projectdb.rs#L430 . Shouldn't be a problem in practice but is ugly nonetheless. thoughts?

### DIFF
--- a/services/src/contexts/postgres.rs
+++ b/services/src/contexts/postgres.rs
@@ -130,9 +130,8 @@ where
                             latest boolean
                         );
 
-                        -- TODO: unique constraint poject_id, lates = true?
-                        -- TODO: index on latest
-
+                        CREATE INDEX project_version_latest_idx 
+                        ON project_versions (project_id, latest DESC, time DESC, author_user_id DESC);
 
                         CREATE TYPE layer_type AS ENUM ('raster', 'vector'); -- TODO: distinguish points/lines/polygons
 
@@ -160,8 +159,6 @@ where
                             id UUID PRIMARY KEY,
                             workflow json NOT NULL
                         );
-
-                        -- TODO: indexes
                         ",
                     )
                     .await?;
--- a/services/src/projects/hashmap_projectdb.rs
+++ b/services/src/projects/hashmap_projectdb.rs
@@ -130,11 +130,6 @@ impl ProjectDB for HashMapProjectDB {
 
         let project_update = project.update_project(update, user);
 
-        ensure!(
-            project_update.version > project.version,
-            error::ProjectUpdateFailed
-        );
-
         project_versions.push(project_update);
 
         Ok(())
--- a/services/src/projects/postgres_projectdb.rs
+++ b/services/src/projects/postgres_projectdb.rs
@@ -527,7 +527,10 @@ where
         let conn = self.conn_pool.get().await?;
         let stmt = conn
             .prepare(
-                "SELECT id, time, author_user_id FROM project_versions WHERE project_id = $1 ORDER BY time DESC",
+                "
+                SELECT id, time, author_user_id
+                FROM project_versions WHERE project_id = $1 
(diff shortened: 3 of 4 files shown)


ChristianBeilschmidt:
I'm confused, since the idea not using clock times makes sense but the code changes look rather strange to me.

1. The modified date should be for the user rather than for us.
2. We can use lamport timestamps aka an incrementing counter on each update

But why do you just remove the check and why does it solve the problem?

The database clock problem should only occur if the database is not on the same system. We could use the time of geo engine, though.

michaelmattig:
- we could use an incrementing counter. my thinking was that if two users work on the same version n of a project and save it, we get two new versions that both should be version n+1, but one is now n+1 and one n+2, and maybe this implies that n+2 is based on n+1?! In this case it would make more sense to store a prev-version-id to more properly model the lineage of the data. But then again the first version would have no prev-version-id and now we have to visualize a version tree... maybe I'm overthinking it and we should just use an incrementing counter.

- using the geo engine instance clock is potentially dangerous if we have multiple instances that use the same database and I think this will be the case in the feature

ChristianBeilschmidt:
>     * we could use an incrementing counter. my thinking was that if two users work on the same version n of a project and save it, we get two new versions that both should be version n+1, but one is now n+1 and one n+2, and maybe this implies that n+2 is based on n+1?! In this case it would make more sense to store a prev-version-id to more properly model the lineage of the data. But then again the first version would have no prev-version-id and now we have to visualize a version tree... maybe I'm overthinking it and we should just use an incrementing counter.

Okay, but then it is still unclear which version is the current one when queried. So… do you implement one of the options now or is this PR finished?

michaelmattig:
basically postgres decides which one is the latest when both have the same timestamp by its transaction ordering. 

still there are two things to solve:
- If we want to guarantee a stable order of the `version()` output I guess we either need to add this version number or `ORDER BC time, author_user_id DESC` ( I vote order by)
- which time do we use, geo engine or database time (I vote database time)

What do you think?

ChristianBeilschmidt:
> basically postgres decides which one is the latest when both have the same timestamp by its transaction ordering.
> 
> still there are two things to solve:
> 
>     * If we want to guarantee a stable order of the `version()` output I guess we either need to add this version number or `ORDER BC time, author_user_id DESC` ( I vote order by)
order by for now -- and don't forget to add an index.


> 
>     * which time do we use, geo engine or database time (I vote database time)
> 
> 
> What do you think?
Yeah, just use the database one.

michaelmattig:
> 
> 
> > basically postgres decides which one is the latest when both have the same timestamp by its transaction ordering.
> > still there are two things to solve:
> > ```
> > * If we want to guarantee a stable order of the `version()` output I guess we either need to add this version number or `ORDER BC time, author_user_id DESC` ( I vote order by)
> > ```
> 
> order by for now -- and don't forget to add an index.
> 
> > ```
> > * which time do we use, geo engine or database time (I vote database time)
> > ```
> > 
> > 
> > What do you think?
> 
> Yeah, just use the database one.

f911415822c97021d0567a2fd562abaac660c11d

PR #45, opened 2020-10-12
RasterTimeAdapter

michaelmattig:
Merges two raster streams into a stream of tuples of rasters with matching temporal validity


ChristianBeilschmidt:
Could you write a test for multiple spatial tiles per time step? This is crucial for processing.

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,457 @@
...
+                    next_qrect.time_interval.start = min(end.0, end.1);
+                    *time_end = None;
+
+                    stream.set(source_a(next_qrect).zip(source_b(next_qrect)));
+
+                    // TODO: handle error
+                    let next = ready!(stream.as_mut().poll_next(cx))
+                        .map(|(a, b)| (a.unwrap(), b.unwrap()));
+                    if let Some((tile_a, tile_b)) = next {
+                        return Poll::Ready(Some(Ok(Self::align_tiles(tile_a, tile_b))));
```
jdroenner, line 116: add a todo to scale data if required.

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,457 @@
...
+                        return Poll::Ready(None);
+                    }
+                }
+            } else {
+                // first tile in time slice
+                *time_end = Some((tile_a.time.end, tile_b.time.end));
+                return Poll::Ready(Some(Ok(Self::align_tiles(tile_a, tile_b))));
+            }
+
+            return Poll::Ready(Some(Ok(Self::align_tiles(tile_a, tile_b))));
```
jdroenner, line 128: is this the case if the intervals are equal?

jdroenner (commented):
i think the approach is the one to go for non-optimized cases. We should discuss how the tile information of the tile alignment branch can help here.

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,457 @@
...
+            time_end,
+            mut stream,
+        } = self.project();
+
+        // TODO: handle error
+        let next = ready!(stream.as_mut().poll_next(cx)).map(|(a, b)| (a.unwrap(), b.unwrap()));
+
+        if let Some((tile_a, tile_b)) = next {
+            if let Some(end) = time_end {
+                if tile_a.time.start >= end.0 || tile_b.time.start >= end.1 {
```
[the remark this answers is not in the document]
michaelmattig: 03ec9cbf453bed58cbb8e49b3590bd8cd4526b9c

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,457 @@
...
+        let RasterTimeAdapterProjection {
+            source_a,
+            source_b,
+            query_rect,
+            time_end,
+            mut stream,
+        } = self.project();
+
+        // TODO: handle error
+        let next = ready!(stream.as_mut().poll_next(cx)).map(|(a, b)| (a.unwrap(), b.unwrap()));
```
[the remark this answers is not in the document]
michaelmattig: f7ac443a1b83523535362901d8f86e08bba8de8b

operators/src/adapters/raster_time.rs
michaelmattig: f7ac443a1b83523535362901d8f86e08bba8de8b

operators/src/adapters/raster_time.rs
michaelmattig: yes, I think so

ChristianBeilschmidt:
[Raster Time Adapter](https://trello.com/c/XBQFYSmm/14-raster-time-adapter)

datatypes/src/primitives/time_interval.rs
```diff
@@ -14,8 +14,8 @@ use std::fmt::{Debug, Display};
 #[derive(Clone, Copy, Deserialize, Serialize, PartialEq, Eq, ToSql, FromSql)]
 #[repr(C)]
 pub struct TimeInterval {
-    start: TimeInstance,
-    end: TimeInstance,
+    pub start: TimeInstance,
+    pub end: TimeInstance,
```
ChristianBeilschmidt, line 18: You can't do this. `TimeInterval` ensures the invariant `start <= end`, so you can't just make it modifiable.

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,379 @@
...
+    // TODO: move to TimeInterval
+    fn intersect_time_intervals(t1: TimeInterval, t2: TimeInterval) -> Option<TimeInterval> {
+        if t1.intersects(&t2) {
+            let start = std::cmp::max(t1.start, t2.start);
+            let end = std::cmp::min(t1.end, t2.end);
+            Some(TimeInterval::new_unchecked(start, end))
+        } else {
+            None
+        }
+    }
```
ChristianBeilschmidt, line 71: Yeah, move it :+1: !

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    fn align_tiles(
+        mut tile_a: RasterTile2D<T1>,
+        mut tile_b: RasterTile2D<T2>,
+    ) -> (RasterTile2D<T1>, RasterTile2D<T2>) {
+        // TODO: scale data if measurement unit requires it?
+        let time = Self::intersect_time_intervals(tile_a.time, tile_b.time)
+            .expect("intervals must overlap");
```
ChristianBeilschmidt, line 79: Looks like you should use a `Result` in the first place

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,379 @@
...
+            time_end,
+            num_spatial_tiles,
+            current_spatial_tile,
+            mut stream,
+            ended,
+        } = self.project();
+
+        if *ended {
+            return Poll::Ready(None);
+        }
```
ChristianBeilschmidt, line 122: Maybe you can implement `FusedStream`, use `ended` there, and use the ` is_terminated(&self)` here.

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,379 @@
...
+
+        // TODO: handle error
```
ChristianBeilschmidt, line 124: which one? does ready panic?

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,379 @@
...
+        // let s1 = qp1.query(query_rect, query_ctx).fuse();
+
+        let qp2 = mrs2
+            .initialize(&exe_ctx)
+            .unwrap()
+            .query_processor()
+            .unwrap()
+            .get_u8()
+            .unwrap();
+        // let s2 = qp2.query(query_rect, query_ctx).fuse();
```
ChristianBeilschmidt, line 327: You could remove the commented-out code.

operators/src/adapters/raster_time.rs
```diff
@@ -0,0 +1,379 @@
...
+        RasterResultDescriptor,
+    };
+    use crate::mock::{MockRasterSource, MockRasterSourceParams};
+    use futures::StreamExt;
+    use geoengine_datatypes::primitives::{BoundingBox2D, SpatialResolution};
+    use geoengine_datatypes::raster::{Grid, RasterDataType};
+    use geoengine_datatypes::spatial_reference::SpatialReference;
+
+    #[tokio::test]
+    async fn adapter() {
```
ChristianBeilschmidt, line 178: I guess we need a second example where the timestamps don't match. Then, you have to introduce NODATA, right?
I wonder if we have to check the data results themselves in the tests as well and not just the timestamps.

ChristianBeilschmidt (changes requested):

datatypes/src/primitives/time_interval.rs
michaelmattig: c57310d69674b43e41291f85aaf0e2b5cda060e3

operators/src/adapters/raster_time.rs
michaelmattig: c57310d69674b43e41291f85aaf0e2b5cda060e3

operators/src/adapters/raster_time.rs
michaelmattig: no, at the struct definition the documentation says;
> /// Assumes that the raster tiles already align spatially.
/// Assumes that raster tiles are contiguous temporally, with no-data-tiles filling gaps.

so the two tiles *must* overlap. Otherwise the source operators validate this assumption and I have to panic here

operators/src/adapters/raster_time.rs
michaelmattig: I tried but it wouldn't let me:
> error[E0382]: borrow of moved value: `self`
   --> operators\src\adapters\raster_time.rs:111:12
    |
99  |     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
    |                  ---- move occurs because `self` has type `Pin<&mut raster_time::RasterTimeAdapter<T1, T2, St1, St2, F1, F2>>`, which does not implement the `Copy` trait
...
109 |         } = self.project();
    |                  --------- `self` moved due to this method call
110 | 
111 |         if self.is_terminated() {
    |            ^^^^ value borrowed here after move
    |
note: this function consumes the receiver `self` by taking ownership of it, which moves `self`
   --> operators\src\adapters\raster_time.rs:18:1

operators/src/adapters/raster_time.rs
michaelmattig: obsolete
ab11218997dc5eba42e7a3fd4d25034f54396ebe

operators/src/adapters/raster_time.rs
michaelmattig: ab11218997dc5eba42e7a3fd4d25034f54396ebe

operators/src/adapters/raster_time.rs
michaelmattig: The timestamps must match because:
> /// Assumes that raster tiles are contiguous temporally, with no-data-tiles filling gaps.

ChristianBeilschmidt: Yes, but you could have two contiguous streams, but the second one ends earlier.

operators/src/adapters/raster_time.rs
ChristianBeilschmidt: lets look together into it

michaelmattig: 9d6080c75e065be0264b666d22a1d8d08e4bda2f

operators/src/adapters/raster_time.rs
jdroenner: you could also have one which starts later.

PR #52, opened 2020-10-26
Arrow Update and CI Fix

ChristianBeilschmidt:


datatypes/src/collections/feature_collection.rs
```diff
@@ -1161,7 +1137,7 @@ mod tests {
         for i in 0..10 {
             assert_eq!(
                 gen_collection(i).byte_size(),
-                empty_hash_map_size + struct_stack_size + 264 + time_interval_size(i)
+                empty_hash_map_size + struct_stack_size + 192 + time_interval_size(i)
```
jdroenner, line 1140: why did this magic number change?

operators/src/mock/mock_feature_collection_source.rs
```diff
@@ -141,7 +141,7 @@ mod tests {
 
         let serialized = serde_json::to_string(&source).unwrap();
         let collection_bytes = [
-            65, 82, 82, 79, 87, 49, 0, 0, 144, 1, 0, 0, 16, 0, 0, 0, 0, 0, 10, 0, 14, 0, 12, 0, 11,
+            65, 82, 82, 79, 87, 49, 0, 0, 148, 1, 0, 0, 16, 0, 0, 0, 0, 0, 10, 0, 14, 0, 12, 0, 11,
```
jdroenner, line 144: is there a new test dataset?

jdroenner (commented):
looks fine. just some questions from my side.

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: It seems like the calculation method changed, I copy-pasted it from their master branch.

operators/src/mock/mock_feature_collection_source.rs
ChristianBeilschmidt: No, but this is the flight format. I was also curious why this changed, but the byte representation is very suboptimal for now. In the future, we should use the json writer if it is available.

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: 76b4180

PR #53, opened 2020-11-05
make api more REST conform and secure workflow registration

michaelmattig:


ChristianBeilschmidt:
Nothing to complain about so far. Just one question, do the examples of WMS and WFS still work with the added authentication and authorization step for workflows?

michaelmattig:
> 
> 
> Nothing to complain about so far. Just one question, do the examples of WMS and WFS still work with the added authentication and authorization step for workflows?

good point, it doesn't work and is missed by the CI. Should I change the example to try to register (might fail when example is opened second time) and then login or should I pass the workflow directly (only supported by WFS I think)?

ChristianBeilschmidt:
> > Nothing to complain about so far. Just one question, do the examples of WMS and WFS still work with the added authentication and authorization step for workflows?
> 
> good point, it doesn't work and is missed by the CI. Should I change the example to try to register (might fail when example is opened second time) and then login or should I pass the workflow directly (only supported by WFS I think)?

Yes, try to register and don't crash, but output _already registered_ to the console. This way, the example should be working again.

michaelmattig:
> 
> 
> > > Nothing to complain about so far. Just one question, do the examples of WMS and WFS still work with the added authentication and authorization step for workflows?
> > 
> > 
> > good point, it doesn't work and is missed by the CI. Should I change the example to try to register (might fail when example is opened second time) and then login or should I pass the workflow directly (only supported by WFS I think)?
> 
> Yes, try to register and don't crash, but output _already registered_ to the console. This way, the example should be working again.

it works now but doesn't log yet. For this to work I still need more detailed error reporting. Right now it only outputs a message but we need to output the error kind/code to the client to correctly detect that the error is in fact that the user already exists and there is no other error.

PR #54, opened 2020-11-09
Point-in-Polygon Operator

ChristianBeilschmidt:


operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+        );
+
+        let vector_sources = self
+            .vector_sources
+            .into_iter()
+            .map(|o| o.initialize(context))
+            .collect::<Result<Vec<Box<InitializedVectorOperator>>>>()?;
+
+        ensure!(
+            vector_sources[0].result_descriptor().data_type == VectorDataType::MultiPoint,
```
michaelmattig, line 51: why is the order of the inputs important?

operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+    }
+}
+
+pub type InitializedPointInPolygonFilter = InitializedOperatorImpl<(), VectorResultDescriptor, ()>;
+
+impl InitializedOperator<VectorResultDescriptor, TypedVectorQueryProcessor>
+    for InitializedPointInPolygonFilter
+{
+    fn query_processor(&self) -> Result<TypedVectorQueryProcessor> {
+        let point_processor = if let TypedVectorQueryProcessor::MultiPoint(point_processor) =
```
michaelmattig, line 82: `self.vector_sources[0].query_processor()?.multi_point().unwrap()` would be shorter

operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+        let mut filter = Vec::with_capacity(points.len());
+
+        let tester = PointInPolygonTester::new(polygons);
+
+        let coordinates = points.coordinates();
+
+        for ((&coordinates_start_index, &coordinates_end_index), time_interval) in points
+            .multipoint_offsets()
+            .iter()
+            .tuple_windows()
```
michaelmattig, line 131: elegant solution but requires a new dependency and is more expensive than looping over an index and taking value at index+1 as end index. is it worth it?

operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+            .zip(points.time_intervals())
+        {
+            let (coordinates_start_index, coordinates_end_index) = (
```
michaelmattig, line 134: why as tuple? would be 2 fewer lines if you cast them individually. you don't need to change it, just curious.

operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+                coordinates_start_index as usize,
+                coordinates_end_index as usize,
+            );
+
+            let is_multi_point_in_polygon_collection = coordinates
+                [coordinates_start_index..coordinates_end_index]
```
michaelmattig, line 140: could also just cast here as it's the only place where the values are used

operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+
+    /// Is the coordinate contained in any polygon of the collection?
+    ///
+    /// The function returns `true` if the `Coordinate2D` is inside the multi polygon, or
+    /// `false` if it is not. If the point is exactly on the edge of the polygon,
+    /// then the function may return `true` or `false`.
+    ///
+    /// TODO: check boundary conditions separately
+    ///
+    pub fn is_coordinate_in_collection(
```
michaelmattig, line 359: maybe rename `is_coordinate_in_any_polygon`. not super important though

operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+
+impl VectorQueryProcessor for PointInPolygonFilterProcessor {
+    type VectorType = MultiPointCollection;
+
+    fn vector_query(
+        &self,
+        query: QueryRectangle,
+        ctx: QueryContext,
+    ) -> BoxStream<'_, Result<Self::VectorType>> {
+        // TODO: multi-threading
```
michaelmattig, line 159: must be cautious when using simd AND multithreading. not all cores have simd units so might get slower the more threads we use.

operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+
+        let filtered_stream = self
+            .points
```
michaelmattig, line 162: wouldn't it make more sense to loop over the polygons first? Otherwise you perform the precomputations for the polygons multiple times.

operators/src/processing/point_in_polygon.rs
```diff
@@ -0,0 +1,775 @@
...
+    /// then the function may return `true` or `false`.
+    ///
+    /// TODO: check boundary conditions separately
+    ///
+    pub fn is_coordinate_in_collection(
+        &self,
+        coordinate: &Coordinate2D,
+        time_interval: &'p TimeInterval,
+    ) -> bool {
+        self.coordinate_in_multi_polygon_iter(coordinate, time_interval)
```
michaelmattig, line 364: `contains(true)`

michaelmattig (changes requested):

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: No, because the output of the operator is `MultiPointCollection`. I would need to cache the point feature flags otherwise for the whole stream.

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: Yes, but you can compute the boolean vector in parallel and then use simd for `or`.

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: Please rephrase, I can't make sense of it.

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: It's Point-in-Polygon, the first operator is points and the second polygons.

datatypes/src/collections/data_types.rs
```diff
@@ -14,6 +15,17 @@ pub enum VectorDataType {
     MultiPolygon,
 }
 
+impl std::fmt::Display for VectorDataType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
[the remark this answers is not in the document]
ChristianBeilschmidt: 4d93f84

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: 358aa07 

Yeah, these shortcuts came after the initial implementation, but are super handy

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: 0be66e7

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: ffa4f2d

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: ffa4f2d

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: 65ee56c

operators/src/processing/point_in_polygon.rs
michaelmattig: but how can the user know which order to input? we could also just require that one input is points and one input ins polygons and extend the code to be able to handle that.

operators/src/processing/point_in_polygon.rs
michaelmattig: so it's a trade-off between space and runtime? A bool array for all features in a stream does not strike me as infeasible to store. Or what do you mean with feature flags?

operators/src/processing/point_in_polygon.rs
michaelmattig: nvm

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: Well, the total runtime is most-likely cheaper, since you do the polygon computations exactly once.
On the other hand, you have to store booleans for the whole stream, which is possible for sizes that are in our mind now, but imposes some limitation nevertheless.
Furthermore, you will get the first results later. For now, you can output the first point collection early, as soon as the first polyogn iteration is finished.
If you do it the other way around, you have to wait for len(polygons) * len(points) iterations for the first output, only in the last point iteration for the last set of polygons, you will get all the outputs.

operators/src/processing/point_in_polygon.rs
ChristianBeilschmidt: I've created the comment: 
362d4e4 

But for the thing about named input, this PR is not the place to change our default implementation of `Operator`.

PR #55, opened 2020-11-10
remove Settings-test.toml from versioning

michaelmattig:
to avoid accidentally comitting local configs to the reposiroty


michaelmattig:
Needs more work, postponed for now.

PR #56, opened 2020-11-10
fix new clippy warnings

michaelmattig:
as we always use the most recent nightly version of clippy, new PRs might fail with errors that have nothing to do with the branch but rather the master that now fails the CI. Maybe we could fix the clippy version to avoid this.

### DIFF
--- a/datatypes/src/collections/multi_point_collection.rs
+++ b/datatypes/src/collections/multi_point_collection.rs
@@ -641,6 +641,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::eq_op)]
     fn nan_equals() {
         let collection = {
             let mut builder = MultiPointCollection::builder();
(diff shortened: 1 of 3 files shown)


ChristianBeilschmidt:
> as we always use the most recent nightly version of clippy, new PRs might fail with errors that have nothing to do with the branch but rather the master that now fails the CI. Maybe we could fix the clippy version to avoid this.

But who will update this version then? We would need automation for that.

michaelmattig:
> 
> 
> > as we always use the most recent nightly version of clippy, new PRs might fail with errors that have nothing to do with the branch but rather the master that now fails the CI. Maybe we could fix the clippy version to avoid this.
> 
> But who will update this version then? We would need automation for that.

you mean something like dependabot we use for wave? If there exists something like that.  Anyway I think the master should always pass CI and right now thats not the case because every run uses a different version of clippy.

michaelmattig:
or we could whitelist lints instead of blacklist s.t. no new rules pop up out of the blue

PR #57, opened 2020-11-10
OpenCL: Lines and Polygons

ChristianBeilschmidt:
Now, our OpenCL context allows us to use and create multi line and polygon collections.

### DIFF
--- a/datatypes/Cargo.toml
+++ b/datatypes/Cargo.toml
@@ -24,7 +24,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 snafu = "0.6"
 vega_lite_3 = { version = "0.2.4", features = ["ndarray", "show_vega"]} # upgrade to 4 if published
-postgres-types = { version = "0.1", features = ["derive"] }
+postgres-types = { version = "0.1", features = ["derive", "with-chrono-0_4"] }
 
 [dev-dependencies]
 criterion = "0.3"
--- a/datatypes/Cargo.toml
+++ b/datatypes/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2018"
 [dependencies]
 # simd compiles with nightly only, allow default features to build with arrow-flight once it is possible
 arrow = {version = "2.0", features = ["simd"] }
+bb8-postgres = { version = "0.4", features = ["with-uuid-0_8", "with-chrono-0_4", "with-serde_json-1"] }
 chrono = "0.4"
 float-cmp = "0.6"
 geo = "0.15"
--- a/datatypes/Cargo.toml
+++ b/datatypes/Cargo.toml
@@ -11,7 +11,6 @@ edition = "2018"
 [dependencies]
 # simd compiles with nightly only, allow default features to build with arrow-flight once it is possible
 arrow = {version = "2.0", features = ["simd"] }
-bb8-postgres = { version = "0.4", features = ["with-uuid-0_8", "with-chrono-0_4", "with-serde_json-1"] }
 chrono = "0.4"
 float-cmp = "0.6"
 geo = "0.15"
--- a/datatypes/src/collections/batch_builder.rs
+++ b/datatypes/src/collections/batch_builder.rs
@@ -95,6 +95,28 @@ impl RawFeatureCollectionBuilder {
         }
     }
 
+    pub fn polygons(
+        types: HashMap<String, FeatureDataType>,
+        num_features: usize,
+        num_polygons: usize,
+        num_rings: usize,
+        num_coords: usize,
+    ) -> Self {
+        Self {
+            types,
+            column_arrays: HashMap::new(),
+            time_array: None,
+            geo_array: None,
+            num_features,
+            num_coords,
+            num_lines: None,
+            num_polygons: Some(num_polygons),
+            num_rings: Some(num_rings),
+            output: None,
+            output_type: VectorDataType::MultiPolygon,
+        }
+    }
+
     pub fn num_features(&self) -> usize {
         self.num_features
     }
@@ -185,7 +207,97 @@ impl RawFeatureCollectionBuilder {
         Ok(())
     }
 
-    // TODO: set lines/polygons
+    pub fn set_lines(
+        &mut self,
+        coords: Buffer,
+        line_offsets: Buffer,
+        feature_offsets: Buffer,
+    ) -> Result<()> {
+        let num_features = (feature_offsets.len() / std::mem::size_of::<i32>()) - 1;
+        let num_lines = (line_offsets.len() / std::mem::size_of::<i32>()) - 1;
+        let num_coords = coords.len() / std::mem::size_of::<Coordinate2D>();
+        let num_floats = num_coords * 2;
+        let data = ArrayData::builder(MultiLineString::arrow_data_type())
+            .len(num_features)
+            .add_buffer(feature_offsets)
+            .add_child_data(
+                ArrayData::builder(arrow::datatypes::DataType::List(
+                    Coordinate2D::arrow_data_type().into(),
+                ))
+                .len(num_lines)
+                .add_buffer(line_offsets)
+                .add_child_data(
+                    ArrayData::builder(Coordinate2D::arrow_data_type())
+                        .len(num_coords)
+                        .add_child_data(
+                            ArrayData::builder(DataType::Float64)
+                                .len(num_floats)
+                                .add_buffer(coords)
+                                .build(),
+                        )
+                        .build(),
+                )
+                .build(),
+            )
+            .build();
+
+        let array = Arc::new(ListArray::from(data)) as ArrayRef;
+
+        self.geo_array = Some(array);
+
+        Ok(())
+    }
+
+    pub fn set_polygons(
+        &mut self,
+        coords: Buffer,
+        ring_offsets: Buffer,
+        polygon_offsets: Buffer,
+        feature_offsets: Buffer,
+    ) -> Result<()> {
+        let num_features = feature_offsets.len() / std::mem::size_of::<i32>() - 1;
+        let num_polygons = polygon_offsets.len() / std::mem::size_of::<i32>() - 1;
+        let num_rings = ring_offsets.len() / std::mem::size_of::<i32>() - 1;
+        let num_coords = coords.len() / std::mem::size_of::<Coordinate2D>();
+        let num_floats = num_coords * 2;
+        let data = ArrayData::builder(MultiPolygon::arrow_data_type())
+            .len(num_features)
+            .add_buffer(feature_offsets)
+            .add_child_data(
+                ArrayData::builder(arrow::datatypes::DataType::List(
+                    arrow::datatypes::DataType::List(Coordinate2D::arrow_data_type().into()).into(),
+                ))
+                .len(num_polygons)
+                .add_buffer(polygon_offsets)
+                .add_child_data(
+                    ArrayData::builder(arrow::datatypes::DataType::List(
+                        Coordinate2D::arrow_data_type().into(),
+                    ))
+                    .len(num_rings)
+                    .add_buffer(ring_offsets)
+                    .add_child_data(
+                        ArrayData::builder(Coordinate2D::arrow_data_type())
+                            .len(num_coords)
+                            .add_child_data(
+                                ArrayData::builder(DataType::Float64)
+                                    .len(num_floats)
+                                    .add_buffer(coords)
+                                    .build(),
+                            )
+                            .build(),
+                    )
+                    .build(),
+                )
+                .build(),
+            )
+            .build();
+
+        let array = Arc::new(ListArray::from(data)) as ArrayRef;
+
+        self.geo_array = Some(array);
+
+        Ok(())
+    }
 
     /// Set the column values for the given column from the given buffers.
     /// `values_buffer` buffer with data of values for construction of primitive array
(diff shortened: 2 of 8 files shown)


operators/src/opencl/cl_program.rs
```diff
@@ -404,31 +429,61 @@ impl<'a> CLProgramRunnable<'a> {
...
+                            format!("IN_FEATURE{}_COORDS", idx),
+                            &create_coordinate_buffer(queue, points.coordinates())?,
+                        )?;
+
+                        kernel.set_arg(
+                            format!("IN_FEATURE{}_OFFSETS", idx),
+                            &create_offset_buffer(queue, points.multipoint_offsets())?,
+                        )?;
+
+                        kernel.set_arg(format!("IN_FEATURE{}_LEN", idx), points.len() as i32)?;
```
michaelmattig, line 442: don't you need offsets_len and coords_len?

operators/src/opencl/cl_program.rs
```diff
@@ -404,31 +429,61 @@ impl<'a> CLProgramRunnable<'a> {
...
+                            format!("IN_FEATURE{}_COORDS", idx),
+                            &create_coordinate_buffer(queue, multi_lines.coordinates())?,
+                        )?;
+
+                        kernel.set_arg(
+                            format!("IN_FEATURE{}_OFFSETS", idx),
+                            &create_offset_buffer(queue, multi_lines.multi_line_string_offsets())?,
+                        )?;
+                        kernel
+                            .set_arg(format!("IN_FEATURE{}_LEN", idx), multi_lines.len() as i32)?;
```
michaelmattig, line 455: don't you need offsets_len. line_offsets_len and coords_len?

operators/src/opencl/cl_program.rs
```diff
@@ -404,31 +429,61 @@ impl<'a> CLProgramRunnable<'a> {
...
+                            format!("IN_FEATURE{}_COORDS", idx),
+                            &create_coordinate_buffer(queue, multi_polygons.coordinates())?,
+                        )?;
+
+                        kernel.set_arg(
+                            format!("IN_FEATURE{}_OFFSETS", idx),
+                            &create_offset_buffer(queue, multi_polygons.multi_polygon_offsets())?,
+                        )?;
+                        kernel.set_arg(
+                            format!("IN_FEATURE{}_LEN", idx),
```
michaelmattig, line 473: don't you need offsets_len. ring_offsets_len and polygon_offsets_len and coords_len?

operators/src/opencl/cl_program.rs
```diff
@@ -491,21 +553,63 @@ impl<'a> CLProgramRunnable<'a> {
...
-                            .build()?;
-                        kernel.set_arg(format!("OUT_POINT_OFFSETS{}", idx), &offsets)?;
-
-                        FeatureGeoOutputBuffer::Points(PointBuffers { coords, offsets })
+                        let coords = create_buffer::<Double2>(queue, features.num_coords())?;
+                        kernel.set_arg(format!("OUT_FEATURE{}_COORDS", idx), &coords)?;
+
+                        let feature_offsets =
+                            create_buffer::<i32>(queue, features.num_features() + 1)?;
+                        kernel.set_arg(format!("OUT_FEATURE{}_OFFSETS", idx), &feature_offsets)?;
```
michaelmattig, line 561: len parameters?

operators/src/opencl/cl_program.rs
```diff
@@ -404,31 +429,61 @@ impl<'a> CLProgramRunnable<'a> {
...
-                        let coordinates_offsets = points.multipoint_offsets();
-                        let buffer = Buffer::builder()
-                            .queue(queue.clone())
-                            .len(coordinates_offsets.len())
-                            .copy_host_slice(coordinates_offsets)
-                            .build()?;
-
-                        kernel.set_arg(format!("IN_POINT_OFFSETS{}", idx), &buffer)?;
+                        kernel.set_arg(
+                            format!("IN_FEATURE{}_COORDS", idx),
```
michaelmattig, line 433: maybe call it `IN_COLLECTION{}` because its not one feature but a collection?

michaelmattig: or plural: features

operators/src/opencl/cl_program.rs
```diff
@@ -738,7 +843,45 @@ impl<'a> CLProgramRunnable<'a> {
                     builder.set_points(coords_buffer, offsets_buffer)?;
                 }
                 None => {}
```
michaelmattig, line 845: move to last place in match block

operators/src/opencl/cl_program.rs
michaelmattig: also maybe calll it num_features? because in_feature{} is not an array so why does it have len?

operators/src/opencl/cl_program.rs
```diff
@@ -1006,17 +1150,56 @@ impl CompiledCLProgram {
                         // no geo
                     }
                     VectorDataType::MultiPoint => {
+                        kernel.arg_named(
+                            format!("IN_FEATURE{}_COORDS", idx),
+                            None::<&Buffer<Double2>>,
+                        );
+
+                        kernel.arg_named(format!("IN_FEATURE{}_LEN", idx), i32::zero());
```
michaelmattig, line 1158: why not None::<i32>? then we know it it wasn't set properly

operators/src/opencl/cl_program.rs
```diff
@@ -1532,6 +1749,271 @@ __kernel void points(
...
+                    __constant const int *IN_FEATURE0_LINE_OFFSETS,
+                    __global double2 *OUT_FEATURE0_COORDS,
+                    __global int *OUT_FEATURE0_OFFSETS
+            ) {
+                int idx = get_global_id(0);
+                
+                OUT_FEATURE0_COORDS[idx].x = IN_FEATURE0_COORDS[idx].x;
+                OUT_FEATURE0_COORDS[idx].y = IN_FEATURE0_COORDS[idx].y + 0.1;
+                
+                if (idx <= IN_FEATURE0_LEN) {
```
michaelmattig, line 1789: need to check output len too

operators/src/opencl/cl_program.rs
```diff
@@ -1532,6 +1749,271 @@ __kernel void points(
...
+                __global int *OUT_FEATURE0_POLYGON_OFFSETS,
+                __global int *OUT_FEATURE0_RING_OFFSETS
+            ) {
+                const int idx = get_global_id(0);
+                const int IN_FEATURE0_POLYGON_LEN = IN_FEATURE0_OFFSETS[IN_FEATURE0_LEN];
+                const int IN_FEATURE0_RING_LEN = IN_FEATURE0_POLYGON_OFFSETS[IN_FEATURE0_POLYGON_LEN];
+                
+                OUT_FEATURE0_COORDS[idx] = IN_FEATURE0_COORDS[idx];
+                
+                if (idx <= IN_FEATURE0_LEN) {
```
michaelmattig, line 1976: check w.r.t. OUT_... array len

datatypes/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2018"
 [dependencies]
 # simd compiles with nightly only, allow default features to build with arrow-flight once it is possible
 arrow = {version = "2.0", features = ["simd"] }
+bb8-postgres = { version = "0.4", features = ["with-uuid-0_8", "with-chrono-0_4", "with-serde_json-1"] }
```
michaelmattig, line 14: ugly dependency. can we feature gate this or limit the dependency to the ToSql/FromSql implementations?

operators/src/opencl/cl_program.rs
michaelmattig: offsets_len is basically the same as num_features but for lines and polygons there are multiple offset arrays and each array needs a len doesn't it?

michaelmattig (changes requested):

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: One is enough to get the other length from within the kernel, cf. my test kernels. I will reconsider the naming.

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: One is enough to get the other length from within the kernel, cf. my test kernels.

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: One is enough to get the other length from within the kernel, cf. my test kernels.

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: Could you re-phrase this comment?

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: Hm, but a None-pointer is of equivalent size as a int32 pointer, but i32 is 4 byte and None is... idk-bytes long?

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: in the case that the output length is wrong, then this check won't let the program output something valid, so is it still necessary?

datatypes/Cargo.toml
ChristianBeilschmidt: You mean the `postgres_types` crate?

datatypes/src/collections/batch_builder.rs
```diff
@@ -185,7 +207,97 @@ impl RawFeatureCollectionBuilder {
         Ok(())
     }
 
-    // TODO: set lines/polygons
+    pub fn set_lines(
```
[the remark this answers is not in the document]
ChristianBeilschmidt: eac3e63

datatypes/src/collections/batch_builder.rs
```diff
@@ -185,7 +207,97 @@ impl RawFeatureCollectionBuilder {
...
+            .build();
+
+        let array = Arc::new(ListArray::from(data)) as ArrayRef;
+
+        self.geo_array = Some(array);
+
+        Ok(())
+    }
+
+    pub fn set_polygons(
```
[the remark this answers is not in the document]
ChristianBeilschmidt: 871a8de

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: fc8f2b9

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: From the documentation

> Scalar and vector arguments may not be null; use zero (e.g. `&0`) instead.

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: cf. other comment

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: b9b625f

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: b9b625f

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: b9b625f

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: b9b625f

datatypes/Cargo.toml
ChristianBeilschmidt: it seems to work without it now :shrug: 

e40a840

ChristianBeilschmidt: 516c46

operators/src/opencl/cl_program.rs
michaelmattig: I meant we need a len-parameter for each output array as well. If we check that we only read valid data from input arrays we should only write data to valid positions in the output array. Of course you can argue that its the caller responsibility to allocate enough memory but I'm not so sure. Additionaly maybe there are algorithms which are flexible in the output length, e.g. a histogram opencl kernel might need the information about the number of buckets which is equal to the length of the output array. Also it would be more uniform if all arrays come as a duo of pointer and length.

operators/src/opencl/cl_program.rs
michaelmattig: If we check that we only read valid data from input arrays we should only write data to valid positions in the output array. Of course you can argue that its the caller responsibility to allocate enough memory but I'm not so sure. Additionaly maybe there are algorithms which are flexible in the output length, e.g. a histogram opencl kernel might need the information about the number of buckets which is equal to the length of the output array. Also it would be more uniform if all arrays come as a duo of pointer and length.

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: e172d81

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: e172d81

Okay, I see the point where the out length determines the algorithm. However, OpenCL is an interplay between runtime parameter specification and kernel work, so I don't see checking the output lengths in this kernel is helpful.

PR #59, opened 2020-11-16
OpenCL: Time of feature collection as in- and output

ChristianBeilschmidt:


operators/src/opencl/cl_program.rs
```diff
@@ -2063,6 +2098,79 @@ __kernel void gid(
...
+        ));
+
+        let mut compiled = cl_program.compile(kernel, "points").unwrap();
+
+        let mut runnable = compiled.runnable();
+        runnable.set_input_features(0, &typed_input).unwrap();
+        runnable.set_output_features(0, &mut out).unwrap();
+        compiled.run(runnable).unwrap();
+
+        let result = out.output.unwrap().get_points().unwrap();
```
michaelmattig, line 2169: please check if time is actually set correctly

datatypes/src/collections/batch_builder.rs
```diff
@@ -147,7 +147,12 @@ impl RawFeatureCollectionBuilder {
     pub fn set_time_intervals(&mut self, values_buffer: Buffer) -> Result<()> {
         let data = ArrayData::builder(TimeInterval::arrow_data_type())
             .len(self.num_features)
-            .add_buffer(values_buffer)
+            .add_child_data(
```
michaelmattig, line 150: please add test that uses this method

michaelmattig:
looks good. please add test and we can merge it

michaelmattig (changes requested):

operators/src/opencl/cl_program.rs
ChristianBeilschmidt: Here, I check if the result equals the input, i.e., it needs to have equal time intervals.

datatypes/src/collections/batch_builder.rs
ChristianBeilschmidt: 5b87247

PR #60, opened 2020-11-19
Common operations for FeatureCollection and TypedFeatureCollection

ChristianBeilschmidt:
This PR draft explores the option to extract functionality of `FeatureCollection`s and make it available for `TypedFeatureCollection`s via a trait and function call propagation.

### DIFF
--- a/datatypes/src/collections/data_types.rs
+++ b/datatypes/src/collections/data_types.rs
@@ -4,8 +4,8 @@ use serde::{Deserialize, Serialize};
 
 use crate::collections::{
     DataCollection, FeatureCollectionError, FeatureCollectionInfos, FeatureCollectionModifications,
-    FilterArray, MultiLineStringCollection, MultiPointCollection, MultiPolygonCollection,
-    ToGeoJson,
+    FilterArray, GeometryCollection, MultiLineStringCollection, MultiPointCollection,
+    MultiPolygonCollection, ToGeoJson,
 };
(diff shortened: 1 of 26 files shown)


datatypes/src/collections/data_types.rs
```diff
@@ -101,3 +101,47 @@ impl TypedFeatureCollection {
         }
     }
 }
+
+/// Implements a function by forwarding its output
+macro_rules! impl_function_by_forwarding_ref {
```
michaelmattig, line 106: can't we use something like this, instead of delegation in each method?
`fn common(&self> -> Box<dyn &FeatureCollectionOperations>)`
or some Deref magic

michaelmattig: also: do we even need the FeatureOperations trait if you call it on a concrete type here anyway?

ChristianBeilschmidt: We could say `deref()` gives you &dyn FeatureCollectionOperations but it is unclear if this is faster than delegating the functions or even handier when working with either collections directly or typed collections (and refs).

ChristianBeilschmidt: The result of the discussion was that it is not possible to do it for operations that return the type itself, so all is done via forwarding to provide a common interface.

michaelmattig:
Please resolve conflicts then I merge

PR #61, opened 2020-11-24
thread-pool

ChristianBeilschmidt:
This PR contains an initial worker thread pool implementation.
1. We have a simple work stealing queue, where each worker thread has a queue and there is a global queue for incoming tasks.
2. There is no scheduling behavior, but the infrastructure should allow extensions on how to schedule tasks. First of all, all tasks have a `group_id` for the context in which they are invoked. But this is not used currently for scheduling.
3. Feel free to provide suggestions for the `TODO`s if it is reasonable to tackle them already in this PR.

### DIFF
--- a/operators/src/concurrency/thread_pool.rs
+++ b/operators/src/concurrency/thread_pool.rs
@@ -83,6 +83,8 @@ impl ThreadPool {
     {
         self.global_queue.push(Box::new(task));
     }
+
+    // TODO: scope for threads
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -109,6 +111,10 @@ impl<'pool> ThreadPoolContext<'pool> {
             task_group_id: TaskGroupId::new(),
             thread_pool,
         }
+    }
+
+    fn degree_of_parallelism(&self) -> usize {
+        self.thread_pool.degree_of_parallelism
     }
 
     pub fn compute<F>(&self, task: F)
@@ -184,6 +190,26 @@ mod tests {
 
         let backoff = Backoff::new();
         while foo.load(Ordering::SeqCst) != 42 {
+            backoff.snooze();
+        }
+    }
+
+    #[test]
+    fn lots_of_tasks() {
+        let thread_pool = ThreadPool::new(2);
+
+        let number_of_tasks = 1_000_000;
+        let tasks_completed = Arc::new(AtomicI32::new(0));
+
+        for _ in 0..number_of_tasks {
+            let tasks_completed = tasks_completed.clone();
+            thread_pool.compute(TaskGroupId::new(), move || {
+                tasks_completed.fetch_add(1, Ordering::SeqCst);
+            });
+        }
+
+        let backoff = Backoff::new();
+        while tasks_completed.load(Ordering::SeqCst) != number_of_tasks {
             backoff.snooze();
         }
     }
--- a/operators/src/concurrency/thread_pool.rs
+++ b/operators/src/concurrency/thread_pool.rs
@@ -50,7 +50,7 @@ impl Task {
 ///
 #[derive(Debug)]
 pub struct ThreadPool {
-    degree_of_parallelism: usize,
+    target_thread_count: usize,
     global_queue: Arc<Injector<Task>>,
     stealers: Vec<Stealer<Task>>,
     threads: Vec<JoinHandle<()>>,
@@ -69,7 +69,7 @@ impl ThreadPool {
             (0..number_of_threads).map(|_| Worker::new_fifo()).collect();
 
         let mut thread_pool = Self {
-            degree_of_parallelism: number_of_threads,
+            target_thread_count: number_of_threads,
             global_queue: Arc::new(Injector::new()),
             stealers: worker_deques.iter().map(Worker::stealer).collect(),
             threads: Vec::with_capacity(number_of_threads),
@@ -156,10 +156,10 @@ impl<'pool> ThreadPoolContext<'pool> {
         }
     }
 
-    /// What is the degree of parallelism that the `ThreadPool` provides?
+    /// What is the degree of parallelism that the `ThreadPool` aims for?
     /// This is helpful to determine how to split the work into tasks.
     pub fn degree_of_parallelism(&self) -> usize {
-        self.thread_pool.degree_of_parallelism
+        self.thread_pool.target_thread_count
     }
 
     /// Compute a task in the `ThreadPool`
--- a/operators/src/concurrency/thread_pool.rs
+++ b/operators/src/concurrency/thread_pool.rs
@@ -448,8 +448,8 @@ mod tests {
         assert_eq!((0..NUMBER_OF_TASKS).collect::<Vec<_>>(), result);
     }
 
-    #[tokio::test]
-    async fn compute_results() {
+    #[test]
+    fn compute_results() {
         const NUMBER_OF_TASKS: usize = 42;
 
         let thread_pool = ThreadPool::new(2);
@@ -463,7 +463,7 @@ mod tests {
             }
         });
 
-        let result = future::join_all(futures).await;
+        let result = futures::executor::block_on(future::join_all(futures));
 
         assert_eq!(result, (0..NUMBER_OF_TASKS).collect::<Vec<_>>());
     }
--- a/operators/src/concurrency/thread_pool.rs
+++ b/operators/src/concurrency/thread_pool.rs
@@ -268,7 +268,10 @@ impl<R> Default for TaskResultOption<R> {
 impl<R> TaskResult<R> {
     fn set(&self, result: R) {
         match self.option.swap(TaskResultOption::Result(result)) {
-            TaskResultOption::None | TaskResultOption::Result(_) => {} // do nothing
+            TaskResultOption::None => {} // do nothing
+            TaskResultOption::Result(_) => {
+                unreachable!("There must not be a second computation of the result")
+            }
             TaskResultOption::Waiting(waker) => waker.wake(),
         };
     }
--- a/operators/src/concurrency/thread_pool.rs
+++ b/operators/src/concurrency/thread_pool.rs
@@ -7,6 +7,9 @@ use crossbeam::atomic::AtomicCell;
 use crossbeam::deque::{Injector, Steal, Stealer, Worker};
 use crossbeam::sync::WaitGroup;
 use crossbeam::utils::Backoff;
+use futures::task::{Context, Poll, Waker};
+use futures::Future;
+use std::pin::Pin;
 
 /// A chunk of work with some metadata
 struct Task {
@@ -123,8 +126,6 @@ impl ThreadPool {
     fn compute(&self, task: Task) {
         self.global_queue.push(task);
     }
-
-    // TODO: scope for threads
 }
 
 /// A computation context for a group that spawns tasks in a `ThreadPool`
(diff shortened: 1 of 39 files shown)


services/src/handlers/wms.rs
```diff
@@ -140,8 +142,10 @@ async fn get_map<C: Context>(
 
     let operator = workflow.operator.get_raster().context(error::Operator)?;
 
+    let thread_pool = ThreadPool::new(1); // TODO: use global thread pool
```
michaelmattig, line 145: I think it should be managed in the services-context together with the different dbs. the services-context could then have a method to create a new executioncontext instance

operators/src/engine/execution_context.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub thread_pool: ThreadPoolContext<'pool>,
+}
+
+/// Create a provider for execution contexts in test environments
+pub struct MockExecutionContextCreator {
+    raster_data_root: PathBuf,
+    thread_pool: ThreadPool,
+}
+
+impl MockExecutionContextCreator {
```
michaelmattig, line 18: what is the purpose of this creator? Where do you use it more than once to create multiple contexts? For the QueryContext I defined a trait and if you would do it like this here the MockExecutionContext could just own the threadpool. Would also be nice if the different Contexts would be implemented similarily

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
...
+    fn compute(&self, task: Task) {
+        self.global_queue.push(task);
+    }
+
+    // TODO: scope for threads
+}
+
+/// A computation context for a group that spawns tasks in a `ThreadPool`
+#[derive(Copy, Clone, Debug)]
+pub struct ThreadPoolContext<'pool> {
```
michaelmattig, line 132: why is this an extra context and not part of the ExecutionContext?

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
...
+#[derive(Debug)]
+pub struct ThreadPool {
+    degree_of_parallelism: usize,
+    global_queue: Arc<Injector<Task>>,
+    stealers: Vec<Stealer<Task>>,
+    threads: Vec<JoinHandle<()>>,
+    next_group_id: AtomicCell<usize>,
+}
+
+impl ThreadPool {
```
michaelmattig, line 58: wouldn't it make sense to make this a trait? Then we could have different schedulers depending on workload and software/hardware stack to be configurable.

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
+use std::sync::Arc;
+use std::thread::JoinHandle;
+use std::{iter, mem};
+
+use crossbeam::atomic::AtomicCell;
+use crossbeam::deque::{Injector, Steal, Stealer, Worker};
+use crossbeam::sync::WaitGroup;
+use crossbeam::utils::Backoff;
+use failure::_core::marker::PhantomData;
```
michaelmattig, line 9: use std::marker instead

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/// A worker thread pool for compute-heavy tasks
+///
+/// TODO: increase threads to a certain maximum number if current threads don't produce results fast enough?
+/// TODO: schedule tasks based on group context?
+///
+#[derive(Debug)]
+pub struct ThreadPool {
+    degree_of_parallelism: usize,
```
michaelmattig, line 51: `number_of_threads`

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
...
+impl<'pool> ThreadPoolContext<'pool> {
+    /// Create a new `ThreadPoolContext`
+    fn new(thread_pool: &'pool ThreadPool, task_group_id: TaskGroupId) -> Self {
+        Self {
+            thread_pool,
+            task_group_id,
+        }
+    }
+
+    /// What is the degree of paralellism that the `ThreadPool` provides?
```
michaelmattig, line 146: - typo parallelism
- I don't get this measurement, what is the unit? can't we just stick to number of threads?

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
...
+                // Extract the stolen task, if there is one.
+                .and_then(Steal::success)
+            });
+
+            if let Some(task) = task {
+                // TODO: recover panics
+                task.call_once();
+                backoff.reset();
+            } else {
+                // TODO: sleep instead of busy waiting?
```
michaelmattig, line 113: why don't you use channels where work comes in that can be awaited on?

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
...
+            "There must be at least one thread for the thread pool"
+        );
+
+        let worker_deques: Vec<Worker<Task>> =
+            (0..number_of_threads).map(|_| Worker::new_fifo()).collect();
+
+        let mut thread_pool = Self {
+            degree_of_parallelism: number_of_threads,
+            global_queue: Arc::new(Injector::new()),
+            stealers: worker_deques.iter().map(Worker::stealer).collect(),
```
michaelmattig, line 71: please explain (or point me to resources) of why you have a global AND worker-local queues. Why not use a global queue where all workers get work from?

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
...
+
+/// A scope in which you can execute tasks and it blocks until all tasks are finished
+#[derive(Debug)]
+pub struct Scope<'pool, 'scope> {
+    thread_pool_context: &'pool ThreadPoolContext<'pool>,
+    wait_group: WaitGroup,
+    _scope_marker: PhantomData<&'scope ()>,
+}
+
+impl<'pool, 'scope> Scope<'pool, 'scope> {
```
michaelmattig, line 189: what is this magic? did you invent this, or where is this scope idea and the particular implementation coming from? can't we also wait for specific group id to be finished?

operators/src/concurrency/thread_pool.rs
```diff
@@ -0,0 +1,339 @@
...
+                backoff.snooze();
+            }
+        }
+    }
+
+    pub fn create_context(&self) -> ThreadPoolContext {
+        ThreadPoolContext::new(self, self.next_group_id.fetch_add(1))
+    }
+
+    fn compute(&self, task: Task) {
```
michaelmattig, line 123: what about tasks that return something? Would be nice to get a future one can await on.

michaelmattig (changes requested):
so far I have a lot of questions. It would also be nice to see the pool actually used in an operator.

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: No, if we for instance decide later on, that we spawn more tasks if some get stuck on tasks for too long, then we still want to provide the operators who use it a degree of parallelism to decide on how to split the work.

E.g., our goal is to provide 1 thread per core, so 8 threads, but for reasons we spawn some more in the pool, we still don't want to create more than 8 tasks for computing items for an array.

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: Because all workers would poll from the same resource all the time, which means you have more lock conflicts. By doing this, you minimize the conflicts on one queue.

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: Hm, I would just need a notification, that there is a chance to get new work from the queue.

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: The idea is taken from https://docs.rs/crossbeam/0.8.0/crossbeam/fn.scope.html and is incredibly useful because of non-static lifetimes. I did a simple version than the one in `crossbeam` for now.

We could wait for the whole group, but that won't help you with lifetimes inside an operator since groups span over workflow graphs.

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: Yeah, that would be somewhat more complicated but I guess also possible. Would compute then return some handle to the yet-to-be-computed result? What would this handle be? Is this a common operation, where we spawn tasks that return something?

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: cf. other discussion

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: Sounds like premature overengineering, but I could do it if you really want it(?).

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: The thread pool does not need to know of the `ExecutionContext` in our query engine. It just reflects the idea to have some workflow-based scheduling in the future, if we have the means to group tasks somehow.

operators/src/engine/execution_context.rs
ChristianBeilschmidt: I just need something to give me a `ThreadPoolContext` with its lifetime. I guess I can't put the `ThreadPool` and `ThreadPoolContext` into the same struct since `ThreadPoolContext` contains a reference to `ThreadPool`.

The trait idea sounds okay, but I don't know if it is part of this PR. There are many questions regarding the `ExecutionContext`, e.g. raster paths and configurations.

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: 81a144e

operators/src/engine/execution_context.rs
michaelmattig: ok let's do a separate PR for this

operators/src/concurrency/thread_pool.rs
michaelmattig: So degree of parallelism would be the target number of threads but the actual number of threads may vary if we spawn additional ones because there is no progress in some tasks? Maybe call it `target_thread_count` and `current_thread_count` then, because `degree` is not really measurable here. We'd also need some parameter how many additional threads may be spawned but we can do it later.

operators/src/concurrency/thread_pool.rs
michaelmattig: yes, do you want to investigate that or leave it as to do?

operators/src/concurrency/thread_pool.rs
michaelmattig: Aren't we always interested in the result of a computation? ^^ In the tests you always pass a shared atomic reference for returning the result. Is this always feasible? E.g. when computing some geometry operation like buffer for a feature collection and each feature spawns a task.

ChristianBeilschmidt: ce760c0 

I'm not quite sure that `Future`s are the appropriate return type here since it mixed async and normal computations, but it works.

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: 1985630

operators/src/concurrency/thread_pool.rs
ChristianBeilschmidt: ef3bfce 

I renamed the variable but not the function. The variable name was misleading, but the function reflects this: https://en.wikipedia.org/wiki/Degree_of_parallelism

PR #63, opened 2020-11-30
Datafusion Example

ChristianBeilschmidt:


ChristianBeilschmidt:
[Data Table to other Geo Join Operator](https://trello.com/c/hsjIpRRb/5-data-table-to-other-geo-join-operator)

operators/tests/datafusion.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    ctx.register_table("foo", Box::new(provider));
+    let df = ctx
+        .table("foo")
+        .and_then(|df| df.filter(col("a").lt_eq(col("b"))))
+        .unwrap();
+
+    let result = df.collect().await.unwrap();
+
+    assert_eq!(result.len(), 1);
```
michaelmattig, line 157: check the actual data instead of checking the number of returned batches

operators/tests/datafusion.rs
```diff
@@ -0,0 +1,202 @@
...
```
michaelmattig, line 199: check the actual data instead of checking the number of returned batches

michaelmattig:
maybe we should create a playground repository for code like this which has no touching points with geo engine code

PR #84, opened 2020-12-09
Dev Docker

ChristianBeilschmidt:


ChristianBeilschmidt:
[Docker image for Dev](https://trello.com/c/iSVFeL4V/13-docker-image-for-dev)

services/src/util/config.rs
```diff
@@ -37,6 +33,31 @@ lazy_static! {
...
+/// test may run in subdirectory
+#[cfg(test)]
+fn retrieve_settings_dir() -> Result<PathBuf> {
+    use crate::error::Error;
+
+    let mut settings_dir = std::env::current_dir().context(error::MissingWorkingDirectory)?;
+
+    let mut unvisited = true;
+
+    while unvisited {
```
michaelmattig, line 45: is it not sufficient to check the current and the parent directory?

ChristianBeilschmidt: b2369c5 

For now, yes, but for adding tests to the binary, maybe not in the future.

PR #85, opened 2020-12-10
Vector Equi Join

ChristianBeilschmidt:


ChristianBeilschmidt:
[Data Table to other Geo Join Operator](https://trello.com/c/hsjIpRRb/5-data-table-to-other-geo-join-operator)

datatypes/src/collections/feature_collection.rs
```diff
@@ -127,6 +128,17 @@ pub trait FeatureCollectionModifications {
     /// This method fails if the columns do not match
     ///
     fn append(&self, other: &Self) -> Result<Self::Output>;
+
+    /// Rename column `old_column_name` to `new_column_name`.
+    fn rename_column(&self, old_column_name: &str, new_column_name: &str) -> Result<Self::Output> {
+        self.rename_columns(&[(old_column_name, new_column_name)])
+    }
+
+    /// Rename columns with (from, to) tuples.
```
michaelmattig, line 137: please specify whether all columns have to be renamed or only some of them

datatypes/src/collections/feature_collection.rs
```diff
@@ -496,6 +508,99 @@ where
             self.types.clone(),
         ))
     }
+
+    fn rename_columns<S1, S2>(&self, renamings: &[(S1, S2)]) -> Result<Self::Output>
+    where
+        S1: AsRef<str>,
+        S2: AsRef<str>,
+    {
+        let mut rename_map: HashMap<&str, &str> = HashMap::with_capacity(renamings.len());
```
michaelmattig, line 517: maybe also check:
 - no two `from`-names are identical
 - no two `to`-names are identical

operators/src/engine/query_processor.rs
```diff
@@ -231,3 +231,25 @@ impl TypedVectorQueryProcessor {
         }
     }
 }
+
+/// Maps a `TypedVectorQueryProcessor` to another `TypedVectorQueryProcessor` by calling a function on its variant.
+/// Call via `map_typed_vector_query_processor!(input, processor => function)`.
+#[allow(unused_macros)]
```
michaelmattig, line 237: why not macro_export instead?

operators/src/processing/vector_join/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+/// Define the type of join
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(tag = "type")]
+pub enum VectorJoinType {
+    /// A left equi join between a `GeoFeatureCollection` and a `DataCollection`
+    EquiLeft {
+        left_column: String,
+        right_column: String,
+        /// which prefix to use if columns have conflicting names?
+        right_column_prefix: String,
```
michaelmattig, line 37: make this an `Option` and have some default behavior like (1) and (2)? Would make usage of operator easier for user

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+};
+use geoengine_datatypes::primitives::{DataRef, FeatureDataRef, Geometry};
+use geoengine_datatypes::util::arrow::ArrowTyped;
+
+use crate::adapters::FeatureCollectionChunkMerger;
+use crate::engine::{QueryContext, QueryProcessor, QueryRectangle, VectorQueryProcessor};
+use crate::error::Error;
+use crate::util::Result;
+
+/// Implements a left inner join between a `GeoFeatureCollection` stream and a `DataCollection` stream.
```
michaelmattig, line 21: isn't a left join an outer join?

operators/src/processing/vector_join/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+    #[serde(flatten)]
+    join_type: VectorJoinType,
+}
+
+/// Define the type of join
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(tag = "type")]
+pub enum VectorJoinType {
+    /// A left equi join between a `GeoFeatureCollection` and a `DataCollection`
+    EquiLeft {
```
michaelmattig, line 33: If we allow other types of joins in the future the type should maybe reflect more precisely what it does. Maybe call it `EquiLeftGeoToData` because in the future we might also support Data to Geo or Geo to Geo or Data to Data joins

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+pub struct EquiLeftJoinProcessor<G> {
+    vector_type: PhantomData<FeatureCollection<G>>,
```
michaelmattig, line 23: do you need it? isn't it  bound by `left_processor`?

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+            left_column,
+            right_column,
+            right_column_prefix: right_prefix,
+        }
+    }
+
+    fn join<'c>(
+        &self,
+        left: &'c FeatureCollection<G>,
+        right: &DataCollection,
```
michaelmattig, line 58: why do left and right collection have different lifetimes?

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+        let matches_iter = left_join_values
+            .as_ref()
+            .iter()
+            .zip(left.time_intervals())
+            .map(move |(left_value, left_time_interval)| {
+                right_join_values
+                    .as_ref()
+                    .iter()
+                    .zip(right.time_intervals())
+                    .map(|(right_value, right_time_interval)| {
```
michaelmattig, line 106: why do you collect match bools instead of pairs of indexes where matches are. wouldn't need to skip so much later on.

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+                        (left_value == right_value)
+                            && left_time_interval.intersects(right_time_interval)
```
michaelmattig, line 108: would maybe be faster to collect the intersection as an Option because your are doint work twice if you first check for intersection and then compute it later on

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+
+                // add time
+                builder.push_time_interval(
+                    left.time_intervals()[left_feature_idx]
+                        .intersect(&right.time_intervals()[right_feature_idx])
+                        .expect("must intersect for join"),
+                )?;
+
+                // add left values
+                for column_name in left.column_names() {
```
michaelmattig, line 148: call the `column_names` function only once instead of every time.

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+            builder.add_column(column_name, column_type)?;
+        }
+
+        let mut builder = builder.finish_header();
+
+        for ((left_feature_idx, matches), geometry) in matches_iter
+            .enumerate()
+            .zip(left.geometries().map(Into::into))
+        {
+            // TODO: is it better to traverse column-wise?
```
michaelmattig, line 130: I think so too because you are performing row-based operations on a columnar data layout.

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+                }
+
+                builder.finish_row();
+            }
+        }
+
+        builder.build().map_err(Into::into)
+    }
+
+    /// Fix column by renaming column with duplicate keys
```
michaelmattig, line 172: maybe it would be wiser to directly use the correct column names (fixing them earlier) instead of creating a new collection here.

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+                    let data_query = self.right_processor.query(query, ctx);
+
+                    data_query
+                        .map(move |right_collection| {
+                            self.join(&left_collection, &right_collection?)
+                        })
+                        .boxed()
+                });
+
+        FeatureCollectionChunkMerger::new(result_stream.fuse(), ctx.chunk_byte_size).boxed()
```
michaelmattig, line 229: You do not only have to make sure the chunks are at least chunk_byte_size large but also that they are not (much) larger. The join result may be much larger than the individual collections.

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
michaelmattig, line 234: needs some more test
 - name collision
 - multiple matches in data collection for single geo feature
 - no matches at all

michaelmattig (changes requested):
cool stuff. please have a look at my comments

operators/src/engine/query_processor.rs
ChristianBeilschmidt: `macro_export` puts the function to the root of the crate, is this intended?

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: They are produced by different streams. I just made it as relaxed as possible.

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: How would I do this?

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: Why should one expect that only part of the inputs is incorporated?

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: It just iterates over the HashMap index, I don't see any improvement here. It does not create any Vec or other owned type.

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: The three tests are in 
7d53833

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: e633632

operators/src/processing/vector_join/mod.rs
ChristianBeilschmidt: b44c58d

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: becf1fd

operators/src/processing/vector_join/mod.rs
ChristianBeilschmidt: becf1fd

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: 27ee0c0

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: It is now quite different
7d53833

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: 3bafb42 

but is now quite different 


cf. 7d53833

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: 4794702

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: 7d53833

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: building the feature collection this way is quite row oriented, but we can try to batch-insert column-wise

dbfdf79

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: 7d53833

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: needed to introduce Iterator and random geo access

7d53833

datatypes/src/collections/feature_collection.rs
michaelmattig: I meant that it is not clear to the user of the function whether they have to specify a (from, to) pair for each existing column. Suggestion: "Rename selected columns with (from, to) tuples"

datatypes/src/collections/feature_collection.rs
michaelmattig: don't you still need to check the `rename_map`s `values` are unique?

operators/src/engine/query_processor.rs
michaelmattig: I think it's okay to put them at the root and also use them in the service crate.

operators/src/processing/vector_join/mod.rs
michaelmattig: using a hard coded english word makes it hard for i18n later on. That's why I would prefer something more universal like `(2)`, `(*)` suffix or something like that. Also `right` is a rather technical term which might not be obvious for non-database users. If it has to be an english prefix I would rather take something like `other` or `(conflict)`

operators/src/processing/vector_join/equi_data_join.rs
michaelmattig: why exactly was the random access necessary? Couldn't you store an iterator and the current feature when a chunk is full? The random access on features looks expensive to me, but I could be wrong.

ChristianBeilschmidt: It is suboptimal, but I can't store the collection and an iterator that references the collection in the same (iterator) struct. If you have an idea on how to accomplish this, we can try it out.

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: aa8e17f

datatypes/src/collections/feature_collection.rs
ChristianBeilschmidt: e8a91de

operators/src/engine/query_processor.rs
ChristianBeilschmidt: 302bab9

operators/src/processing/vector_join/equi_data_join.rs
michaelmattig: I created a moonshot task https://trello.com/c/WKrE6EyY

PR #92, opened 2020-12-22
add lld linker for linux builds

jdroenner:
using .cargo.config it should apply for the whole workspace. the alternative way is to put it into each crates cargo.toml.


jdroenner:
[use lld as linker on linux](https://trello.com/c/JUf9yc9N/134-use-lld-as-linker-on-linux)

michaelmattig:
what do I have to do in addition to `apt install lld`? I get
> error: linker `clang` not found
  |
  = note: No such file or directory (os error 2)

jdroenner:
all i did was ` apt install lld`

jdroenner:
we could keep it as it is and just add a note to the readme for everyone who wants lld.

ChristianBeilschmidt:
I merged the master and updated the readme file.

I also timed my debug build times (`cargo clean && cargo build -Z timings`):

Before:
`172.4s (2m 52.4s)`

After:
`157.1s (2m 37.1s)`

This is a performance improvement of **9%**.

How about you? What do you think?

PR #109, opened 2021-01-11
Make postgres backend an optional feature

michaelmattig:


michaelmattig:
[FeatureGate Postgres](https://trello.com/c/DzWMTD10/110-featuregate-postgres)

services/src/server.rs
```diff
@@ -73,16 +78,21 @@ pub async fn start_server(
...
-            let ctx = PostgresContext::new(
-                tokio_postgres::config::Config::from_str(
-                    &get_config_element::<config::Postgres>()?.config_string,
-                )?,
-                NoTls,
-            )
-            .await?;
-
-            start(shutdown_rx, static_files_dir, bind_address, ctx).await
+            #[cfg(feature = "postgres")]
```
ChristianBeilschmidt, line 81: Would it be possible to just opt-in/out the enum variant at all?

services/src/contexts/postgres.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "postgres")]
```
ChristianBeilschmidt, line 1: You exclude it once in `mod.rs` and once here, wouldn't be one in/exclusion enough?

ChristianBeilschmidt (commented):
What about the CI? Is it now included by default? Excluded? Do we have to check both?

michaelmattig:
> 
> 
> What about the CI? Is it now included by default? Excluded? Do we have to check both?

CI is run with `--all-features` so postgres is checked. It's a good idea to check both, though. Should I add an action that runs without features enabled? If we add more features in the future it might not be feasible to test all combinations of feature activation.

services/src/server.rs
michaelmattig: I would rather report an error if the user configured the backend to use postgres than silently use the in memory variant.

services/src/contexts/postgres.rs
michaelmattig: I guess it would be enough to exclude the whole file here. It wasn't enough to just do it in `mod.rs`. I guess it makes sense to exclude it in both places. If I only exclude it here, the file will (probably) still be accessed during compilation.

ChristianBeilschmidt:
It is okay from my side, maybe, since this is a major layout change, let @jdroenner look also into it.

services/src/contexts/postgres.rs
jdroenner: i guess you have to add cfg(feature = "postgres") in mod.rs before mod postgres.

services/src/projects/postgres_projectdb.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "postgres")]
```
jdroenner, line 1: put the cfg(feature = "postgres") into mod.rs before "pub mod postgres_xxxdb;"

services/src/users/postgres_userdb.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "postgres")]
```
jdroenner, line 1: put the cfg(feature = "postgres") into mod.rs before "pub mod postgres_xxxdb;"

services/src/workflows/postgres_workflow_registry.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "postgres")]
```
jdroenner, line 1: put the cfg(feature = "postgres") into mod.rs before "pub mod postgres_xxx;"

jdroenner:
my main concern is the "#![cfg(feature = "postgres")]" in the postgres files. I think thats not a good idea. It should be enough to add the gate in the mod.rs files before the includes (pub mod postgres_xxx).

services/src/users/postgres_userdb.rs
michaelmattig: 26511745f68be89453766afe268f7d49343fa48b

services/src/workflows/postgres_workflow_registry.rs
michaelmattig: 26511745f68be89453766afe268f7d49343fa48b

services/src/projects/postgres_projectdb.rs
michaelmattig: 26511745f68be89453766afe268f7d49343fa48b

services/src/contexts/postgres.rs
michaelmattig: 26511745f68be89453766afe268f7d49343fa48b

jdroenner:
looks good to me now

PR #115, opened 2021-01-13
Update layers in project

ChristianBeilschmidt:


ChristianBeilschmidt:
[Update layers in projects](https://trello.com/c/FvmqGFVD/159-update-layers-in-projects)

ChristianBeilschmidt:
[Update layers in projects](https://trello.com/c/FvmqGFVD/159-update-layers-in-projects)

services/src/projects/project.rs
```diff
@@ -253,10 +264,57 @@ pub struct UpdateProject {
...
-    pub layers: Option<Vec<Option<Layer>>>,
+    pub layers: Option<Vec<LayerUpdate>>,
     pub bounds: Option<STRectangle>,
 }
 
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub enum LayerUpdate {
+    #[serde(deserialize_with = "deserialize_layer_update_keep")]
+    Keep, // '/'
```
michaelmattig, line 275: `None` would make more sense.

services/src/projects/project.rs
```diff
@@ -53,16 +55,25 @@ impl Project {
...
+                    LayerUpdate::Delete => {}
                 }
             }
+
+            for layer_update in layer_updates {
+                if let LayerUpdate::Update(new_layer) = layer_update {
+                    project.layers.push(new_layer);
+                }
+
+                // TODO: throw errors on other cases?
```
michaelmattig, line 75: yes, please do

services/src/projects/project.rs
```diff
@@ -253,10 +264,57 @@ pub struct UpdateProject {
...
     pub description: Option<String>,
-    pub layers: Option<Vec<Option<Layer>>>,
+    pub layers: Option<Vec<LayerUpdate>>,
     pub bounds: Option<STRectangle>,
 }
 
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub enum LayerUpdate {
+    #[serde(deserialize_with = "deserialize_layer_update_keep")]
```
michaelmattig, line 274: why not just use native serialization? `-` and `/` are not intuitive either.

services/src/projects/project.rs
```diff
@@ -253,10 +264,57 @@ pub struct UpdateProject {
...
     pub name: Option<String>,
     pub description: Option<String>,
-    pub layers: Option<Vec<Option<Layer>>>,
+    pub layers: Option<Vec<LayerUpdate>>,
     pub bounds: Option<STRectangle>,
 }
 
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub enum LayerUpdate {
```
michaelmattig, line 273: Need third variant: `Add` because it doesn't seem natural to Add a (first) layer to a project by updating a non-existing one.

services/src/handlers/projects.rs
```diff
@@ -669,6 +671,183 @@ mod tests {
...
+
+        // add first layer
+        assert_eq!(
+            update_and_load_latest(
+                &ctx,
+                &session,
+                project,
+                UpdateProject {
+                    id: project,
+                    name: None,
```
michaelmattig, line 783: another problem with this approach of handling updates is: what about optional fields? How do we distinguish between "No Update" and "Set value to None". We need a different approach here. My suggestion would be to directly adress the fiels in the URI so "PATCH /project/uuid/name" and set a value. But for now we can do it like this if you want.

services/src/projects/project.rs
```diff
@@ -53,16 +55,25 @@ impl Project {
...
+            let mut layer_updates = layer_updates.into_iter();
+
+            for (layer, layer_update) in layers.zip(&mut layer_updates) {
+                match layer_update {
+                    LayerUpdate::Keep => project.layers.push(layer),
+                    LayerUpdate::Update(updated_layer) => project.layers.push(updated_layer),
+                    LayerUpdate::Delete => {}
                 }
             }
+
```
michaelmattig, line 69: what about layers that are neither addressed as keep, update or delete? should get an error then too.

ChristianBeilschmidt: This is an implicit deletion.

services/src/handlers/projects.rs
ChristianBeilschmidt: I did not change the way layer updates work, but layer list updates.

services/src/projects/project.rs
ChristianBeilschmidt: What is the advantage of that type? I guess we would only throw more errors without enhancing the functionality.

services/src/projects/project.rs
ChristianBeilschmidt: What is native to you? You can make a suggestion.

services/src/projects/project.rs
ChristianBeilschmidt: Currently, you didn't consider it to fail. The functionality is not enhanced by throwing, so should we really do this?

services/src/projects/project.rs
ChristianBeilschmidt: c9ccc31

services/src/projects/project.rs
michaelmattig: it makes it easier to catch bugs in the frontend where it wants to update layers that do not exist.

services/src/projects/project.rs
michaelmattig: I mean the default serde serialization. "None", "Delete" I guess

services/src/projects/project.rs
michaelmattig: it makes it easier to catch bugs in the frontend where it wants to update layers that do not exist. Also makes the API more intuitive. If you look at your test it says "add first layer" but creates an Update-object

services/src/handlers/projects.rs
michaelmattig: is there a difference?

services/src/projects/project.rs
michaelmattig: could also be implicit Keep... Neither is really intuitive which is why I would propose to throw error. Otherwise we have to document this behavior.

services/src/projects/project.rs
ChristianBeilschmidt: So you prefer
```
{
   "None": {}
}
```
```
{
   "Delete": {}
}
```
```
{
   "Update": {
        …
   }
}
```
?

This doesn't looks like a good API design, too, can't we do better?

services/src/projects/project.rs
ChristianBeilschmidt: I could rename it to upsert, because what is the difference between adding and updating after all? We store a new layer object in that place. My problem with it is that it does not provide real added value. Both variants do the same, just one crashes on one occasion. Maybe we will introduce it if we allow partial updates on layers some time.

services/src/handlers/projects.rs
ChristianBeilschmidt: Ah, I misread the code location. I did not change the `UpdateLayer` mechanism, just the layer portion. I guess optional fields are a separate task. As for layers, maybe a binary option is not enough.

services/src/projects/project.rs
ChristianBeilschmidt: The advantage here is that I can just set an arbitrary layer list of arbitrary length (all updates) and it works in all cases.
I would rather document it this way.

services/src/projects/project.rs
michaelmattig: I thought
```
{
  "name": null,
  ...
  "layers": [ "None", "Delete", { "Update": { ...}]
}
```
isn't the default json representation of an payload-less enum value just a string with the variant-name?

services/src/projects/project.rs
michaelmattig: ok

services/src/projects/project.rs
ChristianBeilschmidt: unfortunately not, but I can make it this way

services/src/projects/project.rs
ChristianBeilschmidt: https://github.com/geo-engine/geoengine/pull/115/commits/ec3fb0f6bac6bbf3e76339fc1ee4d2bf0c6f7c13#diff-81d6c97ee571fac79cd18261a47f66164451154aa8f637d1b60cbc9dacea210aR80

services/src/projects/project.rs
ChristianBeilschmidt: https://github.com/geo-engine/geoengine/pull/115/commits/ec3fb0f6bac6bbf3e76339fc1ee4d2bf0c6f7c13#diff-81d6c97ee571fac79cd18261a47f66164451154aa8f637d1b60cbc9dacea210aR280

services/src/projects/project.rs
ChristianBeilschmidt: https://github.com/geo-engine/geoengine/commit/ec3fb0f6bac6bbf3e76339fc1ee4d2bf0c6f7c13#diff-81d6c97ee571fac79cd18261a47f66164451154aa8f637d1b60cbc9dacea210aR45

PR #116, opened 2021-01-13
fix request parsing to use milliseconds

jdroenner:


jdroenner:
[Request parsing shouuld return milli seconds](https://trello.com/c/nq7zb2nU/167-request-parsing-shouuld-return-milli-seconds)

PR #118, opened 2021-01-14
unbreak non postgres build

jdroenner:
build without postgres feature failed because ResultExt is required for .context() in `impl FromStr for SpatialReference {`


jdroenner:
[Unbreak build](https://trello.com/c/qHJ6K1Fv/169-unbreak-build)

PR #119, opened 2021-01-15
add code coveralls to ci

jdroenner:


jdroenner:
[Code coverage mit coveralls](https://trello.com/c/cLAV1WVr/176-code-coverage-mit-coveralls)

ChristianBeilschmidt:
In my opinion, we should include actions into our CI that can fail so that you automatically assess certain parts of a PR.
For code coverage, we just build it, so it cannot fail from the PRs side.

Having said that, we could discuss if we would like to have a lower limit for code coverage that we want to assess in a PR.
Alternatively, we can create a new GitHub workflow that creates the metric on a commit to the master branch.

jdroenner:
the task runs very log so i guess it would be ok if we run it once per day.
However, the CI integration could check if the coverage drops. This could also be a good trigger for a CI fail.

michaelmattig:
- You say the coverage tool runs very long, but it says here run in 17 min while the regular CI took 19 min.
- did you run the coverage manually this time? Because it appears to me as a regular check like the other ones.
- I think it's ok to run the coverage every time even if it takes longer, because you can see the whether the other checks completed already when they complete and you don't have to wait for all to complete first
- the badge says to me "coverage unknown". Does it only work once it is in master or is there something missing still?

jdroenner:
The regular CI run does: 1x run clippy, 1x run fmt, 2x run build, 1x run test. In the same time the coverage job does 1x run build & 1x run test. It is fine for me if we want to run it for each PR & merge. However, it must be a separate job.

We can define a threshold (i guess 80%) as a lower bound in coveralls and a decrease threshold (to discuss). Then, the CI run fails if one of the conditions is true.

ChristianBeilschmidt:
1. Okay, so it is fine for me to run it for every PR if we see some status.
 * Like here: https://github.com/kulp/mangling/pull/4, where you can see **+/-**
 * or even better: integrated into the checks as Clippy does.
 
    Is this possible? Then, one can just assess the coverage of this PR.

2. I doubt saying it must not drop below 80% works, because all the people do shitty PRs and then one guy cannot get his PR accepted. Furthermore, if we say 1%, then you can write less-tested code the bigger the project gets. Furthermore, I doubt tarpaulin has the parameter to fail if it drops x%.

3. If we do it in every PR, why is it necessary to have it as a separate task? We could save installing GDAL twice and spare compile-time if we just append the _tarpauling_ command to the end.

jdroenner:
1.
yes this works: 
image
and you can also integrate it into the checks (have to find out how)

2. i dont know what a good parameter would be. However, you can set it at coveralls (see image) then the ci fails.

3. i would say we split jobs into essential and nice to have. This way we know when the test are done the code works. maybe they can also run in parallel? To save installing gdal we should create a docker image

michaelmattig:
is relative coverage decrease really a sensible measure for a large code base? If we allow x% decrease only PRs that add more than x% of lines to the code base can even fail. It would have to check the coverage of the added code instead but I don't think that's possible.

A minimum coverage of the total code base seems to make more sense to me, but as Chris said this may punish a PR author for lack of tests in the current master.

So I'd say either
- run if for every PR but don't take it as an absolute criterion for merging. Whether we split the checks into essential and nice to have, or just ignore a failing check for merge is another question. Is it even possible to have multiple check groups?
- run it only on master on merge and not as a check in the PR. Then we have the badge and if the coverage degrades we make a new task to improve it

ChristianBeilschmidt:
From our discussion: Please update coveralls to post a message into the PR with the stats.
Then, we can merge it.

PR #123, opened 2021-01-19
Project and layer parameter extension

ChristianBeilschmidt:


jdroenner:
I guess the code is fine. However, do we want to store the visibility of a workflow in different views in a single struct?

PR #132, opened 2021-01-26
Dataset loadinginfo

michaelmattig:
https://trello.com/c/XPbaEk9W


datatypes/src/dataset.rs
```diff
@@ -0,0 +1,62 @@
+use crate::identifier;
+use serde::{Deserialize, Serialize};
+
+identifier!(DataSetProviderId);
+
+identifier!(InternalDataSetId);
+
+identifier!(StagingDataSetId);
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Deserialize, Serialize)]
```
michaelmattig, line 10: > ChristianBeilschmidt commented on 18.01.2021
How is this serialized? I guess we don't want to say:
{
	"operator": "source",
	"dataset": {
		"Internal": "abcd-1234-abcd-1234"
	}
}

> jdroenner commented on 19.01.2021
the alternative is "internal:abcd-1234-abcd-1234". I like the prefix / KV as it enables us to have different handlers.

I don't see a problem with the native serialization but we could always change it if there is an advantage

datatypes/src/util/identifiers.rs
```diff
@@ -0,0 +1,77 @@
+pub trait Identifier: Sized {
+    /// Create a new (random) identifier
+    fn new() -> Self;
+}
+
+#[macro_export]
```
michaelmattig, line 6: >jdroenner commented on 20.01.2021
why do you need the macro? can't you do default implementations on the trait and/or implement it for T:xxx ?

because every identifier is a different type and they have nothing in common. I can't see a different way than using a macro

operators/src/engine/execution_context.rs
```diff
@@ -1,46 +1,163 @@
...
-pub struct ExecutionContext<'pool> {
-    pub raster_data_root: PathBuf,
-    pub thread_pool: ThreadPoolContext<'pool>,
-    pub tiling_specification: TilingSpecification,
+/// A context that provides certain utility access during operator initialization
+pub trait ExecutionContext:
+    Send
+    + Sync
+    + LoadingInfoProvider<MockDataSetDataSourceLoadingInfo, VectorResultDescriptor>
+    + LoadingInfoProvider<OgrSourceDataset, VectorResultDescriptor>
```
michaelmattig, line 20: > jdroenner commented on 20.01.2021
We talked about this. I don't think it is a good idea to add all loading provider requirements here. Just require them in the operator and implement them for the context?
>
> this also requires to change the box in initialize to T. Then in an operator we can do where T: LoadingInfoProvider<MeData, MeResult>

I see it differently. Any impl of ExecutionContext needs to be LoadingInfoProvider for all Operators because it has to give LoadingInfos to all operators. That's why I demand it here. If a operator needs only one specific LoadingInfoProvider it can ask for this one, but any implementation of ExecutionContext must provide them all

operators/src/engine/execution_context.rs
```diff
@@ -1,46 +1,163 @@
...
+
+impl<T, U> Clone for Box<dyn LoadingInfo<T, U>> {
+    fn clone(&self) -> Box<dyn LoadingInfo<T, U>> {
+        self.box_clone()
     }
 }
 
-impl Default for MockExecutionContextCreator {
+pub struct MockExecutionContext {
+    pub raster_data_root: PathBuf,
```
michaelmattig, line 45: > ChristianBeilschmidt commented on 18.01.2021
Why do you make this pub when you provide an add method to add LoadingInfo?

it's only for mocking purposes so why not make it as flexible as possible?

operators/src/engine/execution_context.rs
```diff
@@ -1,46 +1,163 @@
...
+    fn tiling_specification(&self) -> TilingSpecification;
 }
 
-/// Create a provider for execution contexts in test environments
-pub struct MockExecutionContextCreator {
-    raster_data_root: PathBuf,
-    thread_pool: ThreadPool,
-    tiling_specification: TilingSpecification,
+pub trait LoadingInfoProvider<T, U> {
+    fn loading_info(&self, data_set: &DataSetId) -> Result<Box<dyn LoadingInfo<T, U>>>;
```
michaelmattig, line 28: >ChristianBeilschmidt commented on 18.01.2021
You could say something that implements LoadingInfo since you've already got two generics.

>jdroenner commented on 20.01.2021
yes, then you don't have to put the Box<dyn< into the trait. However, it will work if you implement the trait for it.

operators/src/engine/execution_context.rs
```diff
@@ -1,46 +1,163 @@
...
-        ExecutionContext {
-            raster_data_root: self.raster_data_root.clone(),
-            thread_pool: self.thread_pool.create_context(),
-            tiling_specification: self.tiling_specification,
-        }
+pub trait LoadingInfo<T, U>: Debug + Send + Sync {
+    fn get(&self, query: QueryRectangle) -> Result<T>;
+    fn meta(&self) -> Result<U>;
+
+    fn box_clone(&self) -> Box<dyn LoadingInfo<T, U>>;
```
michaelmattig, line 35: > ChristianBeilschmidt commented on 18.01.2021
Coulnd't you expect it to be Clone?

traits can't be clone because they are not sized.

operators/src/engine/query.rs
```diff
@@ -8,10 +8,22 @@ pub struct QueryRectangle {
...
-/// A collection of meta data for processing a query
-#[derive(Copy, Clone, Debug)]
-pub struct QueryContext {
-    // TODO: resolution, profiler, user session, ...
-    // TODO: determine chunk size globally or dynamically from workload? Or global Engine Manager instance that gives that info
+pub trait QueryContext: Send + Sync {
+    fn chunk_byte_size(&self) -> usize;
+}
+
+pub struct MockQueryContext {
```
michaelmattig, line 15: >jdroenner commented on 20.01.2021
why not call it something like SimpleQueryContext or something likt this?

because it should only ever be used in tests

operators/src/source/ogr_source.rs
```diff
@@ -207,24 +187,7 @@ pub enum OgrSourceErrorSpec {
...
-    where
-        G: Geometry,
-    {
-        if let Some(v) = self.0.clone() {
-            Ok(Some(G::try_from(v)?))
-        } else {
-            Ok(None)
-        }
-    }
+    dataset_information: Box<dyn LoadingInfo<OgrSourceDataset, VectorResultDescriptor>>,
```
michaelmattig, line 190: > jdroenner commented on 20.01.2021
can'T we use a generic here and throw a box in it if we need one?

operators/src/source/ogr_source.rs
```diff
@@ -396,19 +261,19 @@ pub struct OgrSourceProcessor<G>
...
     _collection_type: PhantomData<FeatureCollection<G>>,
 }
 
 impl<G> OgrSourceProcessor<G>
 where
     G: Geometry + ArrowTyped,
 {
-    pub fn new(dataset_information: Arc<OgrSourceDataset>, default_geometry: Option<G>) -> Self {
+    pub fn new(
+        dataset_information: Box<dyn LoadingInfo<OgrSourceDataset, VectorResultDescriptor>>,
```
michaelmattig, line 273: > jdroenner commented on 20.01.2021
maybe this could also be a generic. then we just use it with Box if we implement the trait for Box.

operators/src/source/ogr_source.rs
```diff
@@ -715,17 +579,23 @@ where
...
         dataset_information: &OgrSourceDataset,
-        default_geometry: &Option<G>,
         data_types: &HashMap<String, FeatureDataType>,
         query_rectangle: &QueryRectangle,
         time_extractor: &dyn Fn(&Feature) -> Result<TimeInterval>,
         chunk_byte_size: usize,
     ) -> Result<FeatureCollection<G>> {
         let mut builder = feature_collection_builder.finish_header();
 
+        // TODO: cleaner way of handling the default geometry that fails earlier on wrongt data type and avoids cloning
```
michaelmattig, line 589: >ChristianBeilschmidt commented on 18.01.2021
You deleted that previously

I meant a type safe way `enum { Skip, Error, Keep (DefaultGeometry)}`

operators/src/engine/execution_context.rs
michaelmattig: But what if a provider wants to return different implementations of LoadingInfo depending on the data set? IIRC impl means it always has to be the same result type

operators/src/source/ogr_source.rs
michaelmattig: the execution context gives us a `Box<dyn LoadingInfo<...>>`. How shall I convert this to a concrete type? I tried making `OgrState` generic over D where `D: LoadingInfo<OgrSourceDataset, VectorResultDescriptor>` but when I put the box into it it says 
> doesn't have a size known at compile-time

operators/src/source/ogr_source.rs
michaelmattig: but then we don't even avoid the indirections but hide them?!

operators/src/mock/mock_data_set_data_source.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+pub type MockDataSetDataSource = SourceOperator<MockDataSetDataSourceParams>;
+
+#[typetag::serde]
+impl VectorOperator for MockDataSetDataSource {
+    fn initialize(
+        self: Box<Self>,
+        context: &dyn ExecutionContext,
+    ) -> Result<Box<InitializedVectorOperator>> {
```
jdroenner, line 98: dyn?

operators/src/source/ogr_source.rs
jdroenner: you could try +?Sized. I did something like this in the playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5e6550d77a93ce386ea2745502e41adf

operators/src/source/ogr_source.rs
```diff
@@ -234,66 +197,18 @@ pub type InitializedOgrSource =
...
-                reason: "Default geometry must only be specified on error type `keep`".to_string(),
-            });
-        }
+        let info: Box<dyn LoadingInfo<OgrSourceDataset, VectorResultDescriptor>> =
+            context.loading_info(&self.params.data_set)?;
 
         Ok(InitializedOgrSource::new(
             self.params,
-            result_descriptor,
+            info.meta()?,
```
jdroenner, line 207: what is a meta?

operators/src/source/ogr_source.rs
jdroenner: but it looks nicer

operators/src/source/ogr_source.rs
```diff
@@ -420,12 +285,15 @@ where
...
-    fn query(&self, query: QueryRectangle, ctx: QueryContext) -> BoxStream<Result<Self::Output>> {
+    fn query<'a>(
+        &'a self,
+        query: QueryRectangle,
+        ctx: &'a dyn QueryContext,
+    ) -> BoxStream<'a, Result<Self::Output>> {
         OgrSourceStream::new(
-            self.dataset_information.clone(),
-            self.default_geometry.clone(),
+            self.dataset_information.get(query).unwrap(), // TODO: handle error
```
jdroenner, line 294: new task to return Result<Stream>?

datatypes/src/dataset.rs
ChristianBeilschmidt: Can't we just have UUIDs and the system resolves if it is internal or is it advantageous to define it in the workflow?

operators/src/engine/execution_context.rs
ChristianBeilschmidt: There are pros and cons:

Pro:
When you implement an ExecutionContext, you know what to implement.

Cons:
Currently, you can add a new operator by just having it in the program and type-tagged. Now, you are restricted to a few loading providers.
You could do it in the `Any` fashion and try to resolve it.
This makes it extensible, but you can't type check it.

operators/src/engine/query.rs
ChristianBeilschmidt: The name is fine for me :shrug:

operators/src/source/ogr_source.rs
ChristianBeilschmidt: I would say if we can't optimize (generify) it for our core engine, we could leave it as it is then.

operators/src/concurrency/thread_pool.rs
```diff
@@ -140,6 +140,17 @@ impl ThreadPool {
     }
 }
 
+impl Default for ThreadPool {
```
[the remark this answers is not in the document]
ChristianBeilschmidt: I think it is quite nice this way.

operators/src/error.rs
```diff
@@ -135,9 +136,23 @@ pub enum Error {
         source: chrono::format::ParseError,
     },
 
+    LoadingInfo {
+        reason: String, // cannot nest error because it's from service crate
```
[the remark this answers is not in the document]
ChristianBeilschmidt: Box<dyn Error>? But there were some errors doing this, right?

operators/src/source/ogr_source.rs
ChristianBeilschmidt: Metadata? Yeah, I guess the naming is suboptimal.

operators/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(async_closure)]
+#![feature(available_concurrency)]
```
ChristianBeilschmidt, line 2: Meh, it is required to activate some nightly feature for it? Are there alternatives?

ChristianBeilschmidt (changes requested):

datatypes/src/dataset.rs
michaelmattig: possible advantages:
- wokflow reflects where the data resides without looking up further meta data
- implementation allows requiring a specific type of id, e.g. `fn import` requires a `StagedId`
- saves the extra step to lookup the location of the data set because it is encoded in the id

on the other hand one id type makes for a cleaner interface.

datatypes/src/util/identifiers.rs
michaelmattig: if it's ok for you then I propose we leave it like it is for now

operators/src/engine/execution_context.rs
michaelmattig: I get the extensiblity argument. However, the `initialize` method of `Operator` only requires an `ExecutionContext`. How does the `GdalSource` e.g. know that the `&dyn ExecutionContext` also happens to be a `&dyn LoadingInfoProvider<Gdal..>`?

operators/src/engine/query.rs
michaelmattig: Can you think of any concrete usages other than testing for this? If there are any we can rename it, otherwise we should leave it like it is and change it if there is ever the need.

operators/src/source/ogr_source.rs
michaelmattig: this would go against rust's phliosophy. They added the `dyn` keyword exactly for the reason to make it apparent when there is cost to something.

operators/src/engine/execution_context.rs
```diff
@@ -1,46 +1,163 @@
...
 
-impl Default for MockExecutionContextCreator {
+pub struct MockExecutionContext {
+    pub raster_data_root: PathBuf,
+    pub thread_pool: ThreadPool,
+    pub loading_info: HashMap<DataSetId, Box<dyn Any + Send + Sync>>,
+    pub tiling_specification: TilingSpecification,
+}
+
+impl Default for MockExecutionContext {
```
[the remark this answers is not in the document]
michaelmattig: I think it's convention to implement Default whereever possible

operators/src/mock/mock_data_set_data_source.rs
michaelmattig: please elaborate

operators/src/source/ogr_source.rs
michaelmattig: https://trello.com/c/Q3jlJKiT/249-refactor-query-processor-trait-to-return-resultboxstream

operators/src/concurrency/thread_pool.rs
```diff
@@ -140,6 +140,17 @@ impl ThreadPool {
...
+    fn default() -> Self {
+        Self::new(
+            std::thread::available_concurrency()
+                .map(std::num::NonZeroUsize::get)
+                .unwrap_or(1),
+        )
+    }
+}
+
+// TODO: include/merge into ExecutionContext
```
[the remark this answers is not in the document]
michaelmattig: c7d763e6d7c040b39022e5a2893d05d8f9e3d082

operators/src/lib.rs
michaelmattig: https://crates.io/crates/num_cpus

operators/src/concurrency/thread_pool.rs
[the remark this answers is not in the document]
michaelmattig: I can't think of a better default for our use case where we have a single global threadpool

operators/src/error.rs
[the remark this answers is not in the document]
michaelmattig: Maybe something like?
```
    LoadingInfo {
        reason: Box<dyn std::error::Error + Send + Sync>, // cannot nest error because it's from service crate
    },
```
but I can't really test it right now because the service crate where this error shall be user does not yet use it because I splitted the PRs. Shall I leave it like this or remove the error since it's not yet used.

operators/src/source/ogr_source.rs
michaelmattig: If i understand you correctly this would again just hide the fact that the T is boxed, so I vote against it.

operators/src/engine/execution_context.rs
ChristianBeilschmidt: You wouldn't, you just try to get it.

operators/src/error.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: I would at least make a TODO there with this solution. But I guess if you manage to use the Box then it will work in the future.

services/src/handlers/wms.rs
```diff
@@ -147,9 +147,10 @@ async fn get_map<C: Context>(
     let thread_pool = ThreadPool::new(1); // TODO: use global thread pool
 
     let config_tiling_spec = get_config_element::<config::TilingSpecification>()?;
-    let execution_context = ExecutionContext {
+    let execution_context = MockExecutionContext {
```
ChristianBeilschmidt, line 150: What is the reason to use a mock context in our WMS/WFS?

operators/src/engine/execution_context.rs
```diff
@@ -1,46 +1,163 @@
...
-    pub thread_pool: ThreadPoolContext<'pool>,
-    pub tiling_specification: TilingSpecification,
+/// A context that provides certain utility access during operator initialization
+pub trait ExecutionContext:
+    Send
+    + Sync
+    + LoadingInfoProvider<MockDataSetDataSourceLoadingInfo, VectorResultDescriptor>
+    + LoadingInfoProvider<OgrSourceDataset, VectorResultDescriptor>
+{
+    fn thread_pool(&self) -> ThreadPoolContext;
```
ChristianBeilschmidt, line 22:
```suggestion
    fn thread_pool(&self) -> &ThreadPoolContext;
```
I guess it's sufficient to hand out a reference to it. You can clone a context (currently) but I guess you don't want to all the time.

operators/src/lib.rs
ChristianBeilschmidt: Yeah, I've used this one in another project. I guess using it is one step closer to stable.
You should add a todo, though, that covers your feature that will come :tm: .

operators/src/mock/mock_data_set_data_source.rs
jdroenner: i thought InitializedVectorOperator is a trait so it should be a Box<dyn InitializedVectorOperator>

operators/src/source/ogr_source.rs
jdroenner: please find a better name. In this case its a ResultDescriptor?

datatypes/src/dataset.rs
jdroenner: i still think it would be a good idea to be able to add more prefixes. I would suggest to leave it as it is.

datatypes/src/util/identifiers.rs
jdroenner: ok leave it like this

operators/src/engine/execution_context.rs
jdroenner: Well i guess it does not matter if we add all the types here. To avoid a long discussion i guess we should add them. However, i guess we loose type safety anyway. The current approach already uses downcast, as i understand it, there is no way to have the compiler derive all the needed implementations? So you can implement a new source with a special provider without adding the LoadingInfoProvider<X,Y> to ExecutionContext.

operators/src/error.rs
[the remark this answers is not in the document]
jdroenner: yes just add a todo

operators/src/mock/mock_data_set_data_source.rs
michaelmattig: it is a trait but it is a type alias:
`pub type InitializedVectorOperator =
    dyn InitializedOperator<VectorResultDescriptor, TypedVectorQueryProcessor>;`

so the dyn is hidden here. If I try to add it I get
> Expected trait, found type alias `InitializedVectorOperator`

operators/src/source/ogr_source.rs
michaelmattig: I renamed a bunch of stuff, I think it fits much better now fe0f3712025f7fc28e9c5434e3cb5dca2444a775

operators/src/error.rs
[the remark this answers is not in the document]
michaelmattig: 3ef00f432b46f288cdcee28a2d2bc23f01f520c5

operators/src/lib.rs
michaelmattig: 87f62856d6517d1b2533690aeed904d4d9b1a43a

services/src/handlers/wms.rs
michaelmattig: because the real context will be the next step (second part of the dataset PR that I splitted)

operators/src/engine/execution_context.rs
michaelmattig: but who owns the contexts if I only hand out a reference? There can be multiple in each execution context or not?

operators/src/engine/execution_context.rs
michaelmattig: I only use downcast in the mock execution context to make a convenient wrapper.

Can we resolve this discussion (for now)?

jdroenner: yes keep it as it is for now

operators/src/mock/mock_data_set_data_source.rs
jdroenner: ah well then we keep it like this

operators/src/source/ogr_source.rs
jdroenner: ok

operators/src/engine/execution_context.rs
ChristianBeilschmidt: The goal is to have a single context for each execution unit, e.g., computation of a workflow.
Then, each operator can make use of this joint context and schedule work. This allows us to later on schedule differently. For now, we know that all work items share the same "group id" in the ThreadPoolContext.

michaelmattig: Do we really want one group per ExecutionContext? Maybe we want one per user or user group.  Or one for anonymous user and one for registered users. The execution context could abstract that.

If we decide to hand out only references, then the ExecutionContext has to own the ThreadPoolContext which introduces a lifetime to the trait. Also for the Mock-implementation, the struct has to own the ThreadPool AND the ThreadPoolContext (which references the ThreadPool) in order to allow for convenient Default-constructor. As you know referencing a field of the struct itself is not possible. I could turn the ThreadPoolContext into a trait and have an OwningThreadPoolContext which is a single ThreadPoolContext that owns the thrad pool, which might be sufficient in tests. But then the ExecutionContext has to hand out a &dyn ThreadPoolContext a fat pointer you two don't like.

Shall I do it or do you have other ideas to solve this?

ChristianBeilschmidt: > Do we really want one group per ExecutionContext? Maybe we want one per user or user group. Or one for anonymous user and one for registered users. The execution context could abstract that.

It is not clear yet, but that doesn't change that we don't necessarily want to create a thread context on each call.

> If we decide to hand out only references, then the ExecutionContext has to own the ThreadPoolContext which introduces a lifetime to the trait.

Actually, the implementation would need one, but not the trait.

> Also for the Mock-implementation, the struct has to own the ThreadPool AND the ThreadPoolContext (which references the ThreadPool) in order to allow for convenient Default-constructor. As you know referencing a field of the struct itself is not possible.

Yeah, that was the reason for the implementation before you changed it in this PR.

> I could turn the ThreadPoolContext into a trait and have an OwningThreadPoolContext which is a single ThreadPoolContext that owns the thrad pool, which might be sufficient in tests. But then the ExecutionContext has to hand out a &dyn ThreadPoolContext a fat pointer you two don't like.

That is not the only solution. We could also think about…
 * … having a creator like before,
 * … introducing an Arc of the pool inside the context to get rid of the lifetime
 * …

> Shall I do it or do you have other ideas to solve this?

If it's not easy to find a good solution, we should discuss what we want or know for now and then do it.

michaelmattig: The ThreadPoolContext is a ThreadPool-reference and a group id. I agree that multiple users of the thread-pool in the scope of a request may share the same group id but
- I'm not sure that it's always the same group id
- the overhead of having multiple individual contexts with the same group id is really really negligible
- owning the context allows for a cleaner design because it takes the ExecutionContext out of the picture. Operators want to keep the threadPoolContext longer than the execution context, which is only needed in the `initialize` method

I vote for keeping the trait like it is or creating an extra task for this.

ChristianBeilschmidt (approved):

PR #133, opened 2021-01-27
Use spatial ref information

jdroenner:
This PR
- adds checks in wms and wfs services: a) fail if request has no crs, b) fail if request and workflow crs are not equal
- renames wgs84 to epsg_4326
- enables crs reading in the gdal_source


services/src/error.rs
```diff
@@ -66,6 +67,13 @@ pub enum Error {
 
     InvalidNamespace,
 
+    InvalidSpatialReference,
+    #[snafu(display("SpatialReferenceMissmatch {} != {}", a, b))]
+    SpatialReferenceMissmatch {
+        a: SpatialReferenceOption,
+        b: SpatialReferenceOption,
+    },
+
```
ChristianBeilschmidt, line 76: You could maybe also rename the error.

And I would say `expected: {}, found: {}` or does the order not matter for the error and the user?

services/src/handlers/wfs.rs
```diff
@@ -189,6 +189,19 @@ async fn get_feature<C: Context>(
         .initialize(&execution_context)
         .context(error::Operator)?;
 
+    // handle request and workflow crs matching
+    let spatial_ref = initialized.result_descriptor().spatial_reference();
+    // TODO: use a default spatial reference if it is not set?
+    snafu::ensure!(request.srs_name.is_some(), error::InvalidSpatialReference);
```
ChristianBeilschmidt, line 195: 4326 by default?

services/src/handlers/wfs.rs
```diff
@@ -189,6 +189,19 @@ async fn get_feature<C: Context>(
...
+    // TODO: inject projection Operator?
```
ChristianBeilschmidt, line 196:
```suggestion
    // TODO: inject projection Operator
```
:+1:

services/src/handlers/wms.rs
```diff
@@ -169,6 +170,19 @@ async fn get_map<C: Context>(
         .initialize(&execution_context)
         .context(error::Operator)?;
 
+    // handle request and workflow crs matching
+    let spatial_ref = initialized.result_descriptor().spatial_reference();
+    // TODO: use a default spatial reference if it is not set?
+    snafu::ensure!(request.crs.is_some(), error::InvalidSpatialReference);
+    // TODO: inject projection Operator?
```
ChristianBeilschmidt, line 177:
```suggestion
    // TODO: inject projection Operator
```
:+1:

services/src/handlers/wms.rs
```diff
@@ -169,6 +170,19 @@ async fn get_map<C: Context>(
         .initialize(&execution_context)
         .context(error::Operator)?;
 
+    // handle request and workflow crs matching
+    let spatial_ref = initialized.result_descriptor().spatial_reference();
+    // TODO: use a default spatial reference if it is not set?
```
ChristianBeilschmidt, line 175: also here, default to 4326

services/src/ogc/wms/request.rs
```diff
@@ -48,7 +51,7 @@ pub struct GetMap {
     #[serde(alias = "LAYERS")]
     pub layers: String,
     #[serde(alias = "CRS")]
-    pub crs: String, // TODO: parse CRS
+    pub crs: Option<SpatialReference>,
```
ChristianBeilschmidt, line 54: Doesn't `SpatialReferenceOption` work?
We could use no sref if it works instead of a default?

ChristianBeilschmidt (changes requested):

services/src/handlers/wms.rs
jdroenner: The GeoServer docs state crs is a required parameter. So it makes no sense to allow it to be empty. 
https://docs.geoserver.org/stable/en/user/services/wms/reference.html

services/src/handlers/wfs.rs
jdroenner: i really don't know. i guess the default is to use the projection of the dataset or in our case the workflow

services/src/ogc/wms/request.rs
jdroenner: yeah well it was like this in wfs so i just adapted wms. I can change it if you prefer it.

services/src/error.rs
jdroenner: done

services/src/handlers/wms.rs
jdroenner: new task

services/src/handlers/wfs.rs
jdroenner: new task

services/src/error.rs
jdroenner: 29a9e1e

services/src/handlers/wfs.rs
ChristianBeilschmidt: Can you add it as a task somewhere? There are multiple tasks regarding projections (story?). Then, I can close it.

ChristianBeilschmidt: https://trello.com/c/ue8BMo64/255-allow-users-to-re-project-data

ChristianBeilschmidt (approved):

PR #137, opened 2021-02-02
coordinate pixel access for raster tile

ChristianBeilschmidt:


datatypes/src/raster/raster_tile.rs
```diff
@@ -187,11 +189,128 @@ where
...
 
+impl<D, A, P> CoordinatePixelAccess<P> for RasterTile<D, P>
+where
+    D: GridSize + GridSpaceToLinearSpace<IndexArray = A> + GridBounds<IndexArray = A>,
+    A: AsRef<[isize]> + Into<GridIdx<A>> + Clone,
+    P: Pixel,
+    Self: GridIndexAccess<P, GridIdx2D>,
+{
+    fn pixel_value_at_coord(&self, coordinate: Coordinate2D) -> Result<P> {
+        // TODO: lots of computations for each coordinate, can we do better?
```
jdroenner, line 200: we could still add the local_geo_transform calculation to raster_tile like this:
```
    pub fn tile_geo_transform(&self) -> GeoTransform {
        let global_upper_left_idx =
            self.tile_position * [self.grid_array.axis_size_y(), self.grid_array.axis_size_x()];

        let tile_upper_left_coord = self
            .global_geo_transform
            .grid_idx_to_coordinate_2d(global_upper_left_idx);

        GeoTransform::new(
            tile_upper_left_coord,
            self.global_geo_transform.x_pixel_size,
            self.global_geo_transform.y_pixel_size,
        )
    }
```
then maybe add # inline to help the compiler this way, you only have to throw the coordinate at the local geo_transform

jdroenner (commented):
i have a suggestion for an improvement but i guess we could also do this later

datatypes/src/raster/raster_tile.rs
ChristianBeilschmidt: I changed it here: 
a7863be 

I also added some inline's to give the compiler the option to do it.

PR #138, opened 2021-02-02
Dataset context

michaelmattig:
https://trello.com/c/nNttwt8X


services/src/datasets/listing.rs
```diff
@@ -0,0 +1,120 @@
...
+use geoengine_operators::source::OgrSourceDataset;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct DataSetListing {
+    pub id: DataSetId,
+    pub name: String,
+    pub description: String,
+    pub tags: Vec<String>,
+    pub source_operator: String, // TODO: enum?
```
ChristianBeilschmidt, line 20: Hm, this is unclear, an enum would be bad for extensibility.
I guess a string is okay for now, maybe a type id?…

services/src/datasets/listing.rs
```diff
@@ -0,0 +1,120 @@
...
+                                 // TODO: meta data like bounds, crs, resolution
```
ChristianBeilschmidt, line 21: I guess at least CRS is important since Johannes just recently enforced it.

services/src/datasets/listing.rs
```diff
@@ -0,0 +1,120 @@
...
+pub struct DataSetFilter {
+    pub name: String,
+    // TODO: tags, ..
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone, Hash)]
+pub enum OrderBy {
+    NameAsc,
+    NameDesc,
+}
```
ChristianBeilschmidt, line 50: I guess we have or will have redundant ordering enums. Do you think we should merge them?

services/src/datasets/listing.rs
```diff
@@ -0,0 +1,120 @@
...
+    async fn list(
+        &self,
+        user: UserId,
+        options: Validated<DataSetListOptions>,
+    ) -> Result<Vec<DataSetListing>>;
+}
+
+pub enum TypedDataSetProvider {
+    Mock(MockDataSetProvider), // wcs, ...
+}
```
ChristianBeilschmidt, line 69: What do you think about also using typetag here for extensibility?…

services/src/datasets/storage.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub name: String,
+    pub description: String,
+    pub data_type: LayerInfo, // TODO: custom type instead of reusing existing one?
+    pub source_operator: String, // TODO: enum?
+}
+
+impl From<AddDataSet> for DataSet {
+    fn from(value: AddDataSet) -> Self {
+        Self {
+            id: DataSetId::Internal(InternalDataSetId::new()),
```
ChristianBeilschmidt, line 27: maybe add a TODO for using staging later on

services/src/datasets/storage.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub name: String,
+    pub description: String,
+    pub data_type: LayerInfo,
+    pub source_operator: String,
+}
+
+impl From<ImportDataSet> for DataSet {
+    fn from(value: ImportDataSet) -> Self {
+        DataSet {
+            id: DataSetId::Internal(InternalDataSetId::new()),
```
ChristianBeilschmidt, line 62: TODO external?

services/src/datasets/storage.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub name: String,
+    pub description: String,
+    // more meta data (number of data sets, ...)
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub enum AddDataSetProvider {
+    AddMockDataSetProvider(AddMockDataSetProvider),
+    // TODO: geo catalog, wcs, ...
+}
```
ChristianBeilschmidt, line 90: Also here, maybe consider typetag?

services/src/datasets/storage.rs
```diff
@@ -0,0 +1,188 @@
...
+        todo!()
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct DataSetProviderListOptions {
+    // TODO: filter
+    pub offset: u32,
+    pub limit: u32,
+}
```
ChristianBeilschmidt, line 108: I guess pagination options are also quite redundant in services or will the filter be specific?

services/src/datasets/storage.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub data_set: InternalDataSetId,
+    pub permission: DataSetPermission,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone, Hash)]
+pub enum DataSetProviderPermission {
+    Read,
+    Write,
+    Owner,
+}
```
ChristianBeilschmidt, line 181: This also looks redundant, I guess we should think about creating a util for that.

services/src/util/mod.rs
```diff
@@ -1,12 +1,25 @@
...
+}
+
+impl<T> IdResponse<T>
+where
+    T: Identifier,
+{
+    pub fn from_id(id: T) -> Self {
+        Self { id }
+    }
+}
```
ChristianBeilschmidt, line 21: Can't you implement `From`?

services/src/util/mod.rs
```diff
@@ -1,12 +1,25 @@
...
-mod identifiers;
-pub use self::identifiers::{IdResponse, Identifier};
 pub mod config;
 mod string_token;
 pub mod user_input;
 
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+pub struct IdResponse<T: Identifier> {
+    pub id: T,
+}
```
ChristianBeilschmidt, line 12: How is this serialized? Is it just wrapping an id?

ChristianBeilschmidt (changes requested):
Lot's of todos, here are my comments :wink:.

Settings-default.toml
```diff
@@ -13,7 +13,10 @@ config_string = "postgresql://geoengine:geoengine@localhost:5432"
...
-    origin_coordinate_y = 0.0
-    tile_shape_pixels_x = 600
-    tile_shape_pixels_y = 600
+origin_coordinate_x = 0.0
+origin_coordinate_y = 0.0
+tile_shape_pixels_x = 600
+tile_shape_pixels_y = 600
+
+[query_context]
+chunk_byte_size = 1048576
```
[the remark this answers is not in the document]
michaelmattig: it's enough for 256x256 f64 raster tiles. For now it's just an arbitrary number until we have some data to select a better size.

services/src/contexts/in_memory.rs
```diff
@@ -63,4 +85,20 @@ impl Context for InMemoryContext {
...
+        Ok(QueryContextImpl {
+            // TODO: load config only once
+            chunk_byte_size: config::get_config_element::<config::QueryContext>()?.chunk_byte_size,
+        })
+    }
+
+    fn execution_context(&self) -> Result<Self::ExecutionContext> {
+        Ok(ExecutionContextImpl::<HashMapDataSetDB> {
+            data_set_db: self.data_set_db.clone(),
+            thread_pool: self.thread_pool.clone(),
```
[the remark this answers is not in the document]
michaelmattig: I could, but we don't always need one. E.g. if the user logs in we create a `Context` but we don't yet need an `ExecutionContext`

services/src/contexts/in_memory.rs
```diff
@@ -63,4 +85,20 @@ impl Context for InMemoryContext {
...
+            // user: self.session()?.user.id,
```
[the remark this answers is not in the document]
michaelmattig: this is the code that I actually want to use when WMS/WFS use the authorization, that's why I kept it

services/src/contexts/mod.rs
```diff
@@ -37,7 +57,132 @@ pub trait Context: 'static + Send + Sync + Clone {
...
+    async fn data_set_db_ref(&self) -> RwLockReadGuard<Self::DataSetDB>;
+    async fn data_set_db_ref_mut(&self) -> RwLockWriteGuard<Self::DataSetDB>;
+
     fn session(&self) -> Result<&Session>;
 
     fn set_session(&mut self, session: Session);
+
+    fn query_context(&self) -> Result<Self::QueryContext>;
+
+    fn execution_context(&self) -> Result<Self::ExecutionContext>;
```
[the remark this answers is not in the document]
michaelmattig: in order to send out references we first have to create the data structures. And I don't do it by default because they are not alway needed.

services/src/contexts/postgres.rs
```diff
@@ -294,6 +299,18 @@ where
...
+    }
+
+    async fn data_set_db_ref(&self) -> RwLockReadGuard<'_, Self::DataSetDB> {
+        todo!()
+    }
+
+    async fn data_set_db_ref_mut(&self) -> RwLockWriteGuard<'_, Self::DataSetDB> {
+        todo!()
+    }
+
```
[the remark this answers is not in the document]
michaelmattig: I added it as a todo item

services/src/datasets/listing.rs
michaelmattig: so you mean https://doc.rust-lang.org/std/any/struct.TypeId.html ? it doesn't seem to be serializable

services/src/datasets/listing.rs
michaelmattig: Merging is difficult, because different types have different attributes to sort by. We would need something like
```
struct Ordering {
  attribute: Attribute,
  order: Order,
}
```
but ideally it should still serialize to a string. I guess it's more of an "Idea" task

services/src/datasets/listing.rs
michaelmattig: with a enum the backend can use `match` to handle the different providers. If we use a trait then we have to somehow put the logic into the trait implementations. Not sure how to realize different behavior for different backends then.

services/src/datasets/storage.rs
michaelmattig: the idea is that added data sets (`AddDataSet`) always turn into internal data sets. For staging you would first `stage` and then `import` but this will be a different PR

services/src/datasets/storage.rs
michaelmattig: imported data sets always become internal data sets

services/src/contexts/mod.rs
```diff
@@ -37,7 +57,132 @@ pub trait Context: 'static + Send + Sync + Clone {
...
+
+    fn raster_data_root(&self) -> std::result::Result<PathBuf, geoengine_operators::error::Error> {
+        Ok(get_config_element::<GdalSource>()
+            .map_err(|_| geoengine_operators::error::Error::RasterRootPathNotConfigured)?
+            .raster_data_root_path)
+    }
+
+    fn tiling_specification(&self) -> TilingSpecification {
+        // TODO: load only once and handle error
+        let config_tiling_spec = get_config_element::<config::TilingSpecification>().unwrap();
```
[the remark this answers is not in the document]
michaelmattig: the method above is obsolete anyhow once GdalSource uses LoadingInfo, so no need for a TODO

services/src/contexts/mod.rs
```diff
@@ -37,7 +57,132 @@ pub trait Context: 'static + Send + Sync + Clone {
...
+    }
+}
+
+// TODO: use macro(?) for delegating meta_data function to DataSetDB to avoid redundant code
+impl<D> MetaDataProvider<MockDataSetDataSourceLoadingInfo, VectorResultDescriptor>
+    for ExecutionContextImpl<D>
+where
+    D: DataSetDB + MetaDataProvider<MockDataSetDataSourceLoadingInfo, VectorResultDescriptor>,
+{
+    // TODO: make async
```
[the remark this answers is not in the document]
michaelmattig: it would require the `initialize` method of Operator to be async (or use blocking there instead). So it would be quite a bigger change.

services/src/contexts/mod.rs
```diff
@@ -37,7 +57,132 @@ pub trait Context: 'static + Send + Sync + Clone {
...
+                    .map_err(|e| geoengine_operators::error::Error::LoadingInfo {
+                        source: Box::new(e),
+                    })?
+                    .meta_data(data_set),
+            }
+        })
+    }
+}
+
+// TODO: use macro(?) for delegating meta_data function to DataSetDB to avoid redundant code
```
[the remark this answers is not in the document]
michaelmattig: I think it would just be a proxy. I tried to implement it using generics but there are no appropriate bounds to use for generic implementation because the DataSetProvider only implements very specific generics which have nothing in common. Not sure if auto-impl would work.

services/src/datasets/listing.rs
michaelmattig: 74bb6bae607bd98f778a77ea862b32fc132ae3f3

services/src/util/mod.rs
michaelmattig: 9198d00ef59c7e0d24fc3a82cf46730c23a9aca1

services/src/util/mod.rs
michaelmattig: Yes just a wrapper for a more clean output than just a string

services/src/contexts/mod.rs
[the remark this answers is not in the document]
michaelmattig: the method above is obsolete anyhow once GdalSource uses LoadingInfo, so no need for a TODO

services/src/datasets/storage.rs
michaelmattig: yeah we could reuse a pagination struct but that would mean one more nesting level while not really saving much code

services/src/datasets/storage.rs
michaelmattig: we could think about a macro once it's all working

services/src/datasets/storage.rs
michaelmattig: it might make sense here but I would add it as a TODO because I am not sure yet how to implement the providers at all. I would address it in the PR that actually implements the backend.

Settings-default.toml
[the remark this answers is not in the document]
ChristianBeilschmidt: Okay, so we should add a TODO I guess.

services/src/util/mod.rs
ChristianBeilschmidt: Maybe I get this wrong, but an id is an id. Do you want to say
```json
{
  "id": {
    "id": "abc"
  }
}
```
…or is it for working in the code?

ChristianBeilschmidt (commented):
As discussed, the global context shouldn't have the session, but rather the created execution context.

ChristianBeilschmidt:
[GE Context erweitern, um ExecutionContext und QueryContext zu erstellen](https://trello.com/c/nNttwt8X/264-ge-context-erweitern-um-executioncontext-und-querycontext-zu-erstellen)

services/src/util/mod.rs
michaelmattig: The `Identifier`s serialize to a simple String containing a UUID. APIs that create something and return the ID like `POST user` would then just output `"9d15d404-dfe7-490a-99ab-d48493368eac"`

Using the wrapper yields
```
{
  "id": "9d15d404-dfe7-490a-99ab-d48493368eac"
}
```
which seems better to me.

michaelmattig:
> 
> 
> As discussed, the global context shouldn't have the session, but rather the created execution context.

I added a new task to the sprint because this concerns more than just the Execution and QueryContext: https://trello.com/c/XjfiZKNL

ChristianBeilschmidt (approved):

PR #140, opened 2021-02-03
vector-attributes-in-descriptor

ChristianBeilschmidt:


ChristianBeilschmidt:
[Vektor-Attribute im ResultDescriptor](https://trello.com/c/Kus16UbI/205-vektor-attribute-im-resultdescriptor)

operators/src/engine/result_descriptor.rs
```diff
@@ -69,10 +71,25 @@ impl ResultDescriptor for RasterResultDescriptor {
...
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 pub struct VectorResultDescriptor {
     pub data_type: VectorDataType,
     pub spatial_reference: SpatialReferenceOption,
+    pub columns: HashMap<String, FeatureDataType>,
+}
+
+impl VectorResultDescriptor {
+    /// Create a new `VectorResultDescriptor` by only using new `columns`
+    pub fn map_columns<F>(&self, f: F) -> Self
```
michaelmattig, line 83: why not consume self like the other map functions on result descriptor?

operators/src/engine/result_descriptor.rs
```diff
@@ -69,10 +71,25 @@ impl ResultDescriptor for RasterResultDescriptor {
...
-#[derive(Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 pub struct VectorResultDescriptor {
     pub data_type: VectorDataType,
     pub spatial_reference: SpatialReferenceOption,
+    pub columns: HashMap<String, FeatureDataType>,
+}
+
+impl VectorResultDescriptor {
+    /// Create a new `VectorResultDescriptor` by only using new `columns`
```
michaelmattig, line 82: why is columns in tick when there is no parameter named columns? Also why do they all have to be new? Then I wouldn't use map but new

operators/src/mock/mock_raster_point_join.rs
```diff
@@ -91,23 +92,40 @@ impl VectorOperator for MockRasterPointJoinOperator {
...
+
+        let raster_sources = self
+            .raster_sources
+            .into_iter()
+            .map(|o| o.initialize(context))
+            .collect::<Result<Vec<_>>>()?;
+
+        let result_descriptor = {
+            let mut columns = vector_sources[0].result_descriptor().columns.clone();
+            columns.insert(self.params.feature_name.clone(), FeatureDataType::Number);
```
michaelmattig, line 109: check if column with that name already exists?

operators/src/mock/mock_raster_point_join.rs
```diff
@@ -91,23 +92,40 @@ impl VectorOperator for MockRasterPointJoinOperator {
...
+
+            VectorResultDescriptor {
+                spatial_reference: vector_sources.get(0).map_or_else(
```
michaelmattig, line 112: I think you can just unwrap here because you used the first vector source already so it has to exist

operators/src/mock/mock_raster_point_join.rs
```diff
@@ -91,23 +92,40 @@ impl VectorOperator for MockRasterPointJoinOperator {
...
+            .collect::<Result<Vec<_>>>()?;
+
+        let raster_sources = self
+            .raster_sources
+            .into_iter()
+            .map(|o| o.initialize(context))
+            .collect::<Result<Vec<_>>>()?;
+
+        let result_descriptor = {
+            let mut columns = vector_sources[0].result_descriptor().columns.clone();
```
michaelmattig, line 108: check if a raster source exists

operators/src/mock/mock_raster_point_join.rs
```diff
@@ -91,23 +92,40 @@ impl VectorOperator for MockRasterPointJoinOperator {
...
+            columns.insert(self.params.feature_name.clone(), FeatureDataType::Number);
+
+            VectorResultDescriptor {
+                spatial_reference: vector_sources.get(0).map_or_else(
+                    || SpatialReferenceOption::Unreferenced,
+                    |o| o.result_descriptor().spatial_reference,
+                ),
+                data_type: VectorDataType::MultiPoint,
```
michaelmattig, line 116: do we need to check that first vector source is actually MultiPoint?

operators/src/processing/raster_vector_join/mod.rs
```diff
@@ -86,14 +88,40 @@ impl VectorOperator for RasterVectorJoin {
         // TODO: check for column clashes earlier with the result descriptor
         // TODO: update result descriptor with new column(s)
```
michaelmattig, line 89: remove

operators/src/processing/raster_vector_join/mod.rs
```diff
@@ -86,14 +88,40 @@ impl VectorOperator for RasterVectorJoin {
...
+                    AggregationMethod::First => {
+                        match raster_sources[i].result_descriptor().data_type {
+                            RasterDataType::U8
+                            | RasterDataType::U16
+                            | RasterDataType::U32
+                            | RasterDataType::U64
+                            | RasterDataType::I8
+                            | RasterDataType::I16
+                            | RasterDataType::I32
+                            | RasterDataType::I64 => FeatureDataType::Decimal,
```
michaelmattig, line 110: remark: I still find it confusing that our integer types are named Decimal when e.g. in Postgres Decimal is a variable precision data type with a decimal point.

operators/src/processing/vector_join/util.rs
```diff
@@ -0,0 +1,22 @@
+use std::collections::{HashMap, HashSet};
+
+/// Create a translation table to resolve name conflicts in the `DataCollection`
+pub(super) fn translation_table<'i>(
```
michaelmattig, line 4: would be a good idea to test this function standalone, especially with the edge cases of suffixeing stuff multiple times as long as collisions occur

services/src/handlers/workflows.rs
```diff
@@ -49,6 +52,37 @@ async fn load_workflow<C: Context>(id: Uuid, ctx: C) -> Result<impl warp::Reply,
     Ok(warp::reply::json(&wf).into_response())
 }
 
+pub(crate) fn get_workflow_metadata_handler<C: Context>(
+    ctx: C,
+) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
+    warp::get()
+        .and(warp::path!("workflow" / "metadata" / Uuid))
```
michaelmattig, line 59: wouldn't it be more RESTful to call /workflow/UUid/metdata ?
and maybe /workflow/UUid/definition for the actual workflow? then /workflow/UUid could deliver both..

services/src/handlers/workflows.rs
```diff
@@ -49,6 +52,37 @@ async fn load_workflow<C: Context>(id: Uuid, ctx: C) -> Result<impl warp::Reply,
...
+    ctx: C,
+) -> Result<impl warp::Reply, warp::Rejection> {
+    // TODO: check for rights to access workflow?
+    let workflow = ctx
+        .workflow_registry_ref()
+        .await
+        .load(&WorkflowId(id))
+        .await?;
+    let operator = workflow.operator.get_vector().context(error::Operator)?;
+    // TODO: use meaningful context to initialize
```
michaelmattig, line 75: you can tet it from `ctx` now

services/src/handlers/workflows.rs
```diff
@@ -49,6 +52,37 @@ async fn load_workflow<C: Context>(id: Uuid, ctx: C) -> Result<impl warp::Reply,
...
+    let operator = operator
+        .initialize(&MockExecutionContext::default())
+        .context(error::Operator)?;
+
+    let result_descriptor = operator.result_descriptor();
```
michaelmattig, line 80: we should think about caching this. maybe we can calculate it once on workflow creation? It doesn't depend on the query rectangle.

operators/src/engine/result_descriptor.rs
ChristianBeilschmidt: Because you now get result descriptors by reference and then you would always need to say clone and then map. Then you've cloned data and discard it, e.g. a hash map, so I think it makes sense this way.

operators/src/processing/raster_vector_join/mod.rs
ChristianBeilschmidt: True, maybe we can sit together and think of reworks to the feature data types for a small story.

operators/src/engine/result_descriptor.rs
ChristianBeilschmidt: ba5480b 

You don't have to, you only get a reference of the hash map.

operators/src/mock/mock_raster_point_join.rs
ChristianBeilschmidt: 1b9537d

operators/src/mock/mock_raster_point_join.rs
ChristianBeilschmidt: 1b9537d

operators/src/mock/mock_raster_point_join.rs
ChristianBeilschmidt: 1b9537d

operators/src/processing/raster_vector_join/mod.rs
ChristianBeilschmidt: 1d3aea5

operators/src/processing/vector_join/util.rs
ChristianBeilschmidt: 5477481

services/src/handlers/workflows.rs
ChristianBeilschmidt: 7075bd4

services/src/handlers/workflows.rs
ChristianBeilschmidt: I added a todo

7075bd4 

caching it would mean we compiled it ;-)

services/src/handlers/workflows.rs
ChristianBeilschmidt: afbbacb

operators/src/mock/mock_raster_point_join.rs
ChristianBeilschmidt: 69e3441

PR #144, opened 2021-02-05
Store data sets in memory and list them

michaelmattig:
https://trello.com/c/YN9Tmih6

### DIFF
--- a/services/src/datasets/storage.rs
+++ b/services/src/datasets/storage.rs
@@ -1,10 +1,11 @@
-use crate::datasets::listing::DataSetProvider;
+use crate::datasets::listing::{DataSetListing, DataSetProvider};
 use crate::error::Result;
 use crate::projects::project::LayerInfo;
 use crate::users::user::UserId;
 use crate::util::user_input::{UserInput, Validated};
 use async_trait::async_trait;
 use geoengine_datatypes::dataset::{DataSetId, DataSetProviderId, InternalDataSetId};
+use geoengine_datatypes::spatial_reference::SpatialReferenceOption;
 use geoengine_datatypes::util::Identifier;
 use geoengine_operators::mock::MockDataSetDataSourceLoadingInfo;
 use geoengine_operators::source::OgrSourceDataset;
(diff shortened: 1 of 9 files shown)


michaelmattig:
[In-Memory Dataset DB](https://trello.com/c/YN9Tmih6/263-in-memory-dataset-db)

services/src/datasets/storage.rs
```diff
@@ -18,7 +19,21 @@ pub struct DataSet {
...
+    pub source_operator: String,
+    pub spatial_reference: SpatialReferenceOption,
+}
+
+impl DataSet {
+    pub fn listing(&self) -> DataSetListing {
+        DataSetListing {
+            id: self.id.clone(),
+            name: self.name.clone(),
+            description: self.description.clone(),
```
ChristianBeilschmidt, line 31: is it actually necessary to clone all this stuff or could we just give out &str? We serialize it anyway.

services/src/datasets/storage.rs
```diff
@@ -133,7 +152,10 @@ pub enum VectorLoadingInfo {
 
 /// Handling of data sets provided by geo engine internally, staged and by external providers
 #[async_trait]
-pub trait DataSetDB: DataSetProvider + Send + Sync {
+pub trait DataSetDB: DataSetProvider + DataSetProviderDB + Send + Sync {}
+
+#[async_trait]
+pub trait DataSetProviderDB {
```
ChristianBeilschmidt, line 158: Maybe you can add some info on public traits and types

services/src/datasets/storage.rs
```diff
@@ -159,6 +179,20 @@ pub trait DataSetDB: DataSetProvider + Send + Sync {
     ) -> Result<&dyn DataSetProvider>;
 }
 
+pub trait DataSetStorer: Send + Sync {
+    type D: Send + Sync;
+}
```
ChristianBeilschmidt, line 184: if `D` is the meta data type, maybe you should call it
```suggestion
pub trait DataSetStorer: Send + Sync {
    type MetaData: Send + Sync;
}
```

ChristianBeilschmidt (changes requested):

services/src/datasets/storage.rs
michaelmattig: it's not necessary but I use this function only in the in memory DataSetDB. In real deployments using Postgres backend I would create the DataSetListing directly out of data from the db and then the Listing actually has to own the data

services/src/datasets/storage.rs
michaelmattig: a0c4a7488f489f31a35aab8b045816f567874a98

services/src/datasets/storage.rs
michaelmattig: a0c4a7488f489f31a35aab8b045816f567874a98

ChristianBeilschmidt (approved):

PR #145, opened 2021-02-05
detach session from Context

michaelmattig:
https://trello.com/c/XjfiZKNL

### DIFF
--- a/services/src/handlers/projects.rs
+++ b/services/src/handlers/projects.rs
@@ -3,6 +3,7 @@ use crate::projects::project::{
     CreateProject, LoadVersion, ProjectId, ProjectListOptions, UpdateProject, UserProjectPermission,
 };
 use crate::projects::projectdb::ProjectDB;
+use crate::users::session::Session;
 use crate::util::user_input::UserInput;
 use crate::util::IdResponse;
 use uuid::Uuid;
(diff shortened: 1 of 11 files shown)


michaelmattig:
[Context vs Session](https://trello.com/c/XjfiZKNL/280-context-vs-session)

services/src/handlers/projects.rs
```diff
@@ -13,21 +14,23 @@ pub(crate) fn create_project_handler<C: Context>(
 ) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
     warp::post()
         .and(warp::path("project"))
-        .and(authenticate(ctx))
+        .and(authenticate(ctx.clone()))
+        .and(warp::any().map(move || ctx.clone()))
```
ChristianBeilschmidt, line 18: Is this okay, because you rely on that C is an `Arc` (cheap to clone) or would it be advantageous to pass references to the actual thing you need, for instance, `&mut ProjectDB`?

michaelmattig: As far as I seet it I can't pass a reference here because the filter has to capture the variable in order to execute the functions multiple times for multiple different (web) calls.

PR #148, opened 2021-02-08
First Plot Operator

ChristianBeilschmidt:


ChristianBeilschmidt:
[Ersten Plot erstellen (stats)](https://trello.com/c/YqJHvUkl/300-ersten-plot-erstellen-stats)

operators/src/engine/query_processor.rs
```diff
@@ -82,6 +83,24 @@ where
...
 
+/// An instantiation of a plot operator that produces a stream of vector results for a query
+pub trait PlotQueryProcessor: Sync + Send {
+    type PlotType;
+
+    fn plot_query<'a>(
+        &'a self,
+        query: QueryRectangle,
+        ctx: &'a dyn QueryContext,
+    ) -> BoxFuture<'a, Result<Self::PlotType>>;
```
michaelmattig, line 94: is it safe to assume that there is always exactly one plot resulting from a workflow?

I could think of
- Multiple views of the same data (paginated in frontend)
 - could also be realized with a special compund plot type
- plot that is split into multiple plots for contiguous time spans
 - query a decade, get a plot for each year
 
not sure if that makes sense but we should at least think about this.

operators/src/engine/query_processor.rs
```diff
@@ -248,6 +267,31 @@ impl TypedVectorQueryProcessor {
     }
 }
 
+/// An enum that contains all possible query processor variants
+#[allow(clippy::pub_enum_variant_names)]
+pub enum TypedPlotQueryProcessor {
```
michaelmattig, line 272: The json type I get but do we really have to commit to PNG here? What if we want to deliver plots as .webp or SVG? Do we introduce more types? Does the client have to care about the image format?

operators/src/engine/result_descriptor.rs
```diff
@@ -120,6 +120,34 @@ impl ResultDescriptor for VectorResultDescriptor {
     }
 }
 
+/// A `ResultDescriptor` for plot queries
+#[derive(Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Serialize, Deserialize)]
+pub struct PlotResultDescriptor {}
+
+impl ResultDescriptor for PlotResultDescriptor {
+    type DataType = (); // TODO: maybe distinguish between image, interactive plot, etc.
```
michaelmattig, line 128: The basic distinction would be between backend and frontend rendered plots. And the type also has to correspond to the `TypedPlotQueryProcessor`. But how is that relevant for the client? Why does he need to know how the plot is rendered? It's not like for other layers where we need the info because they are used as input for operators.

operators/src/util/number_statistics.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    pub fn max(&self) -> f64 {
+        if self.value_count > 0 {
+            self.max_value
+        } else {
+            f64::NAN
+        }
+    }
+
+    pub fn mean(&self) -> f64 {
```
michaelmattig, line 83: why return NAN and not Option<f64> ?

operators/src/engine/query_processor.rs
ChristianBeilschmidt: I guess for now it is since we currently don't know what to do if there is more than one output.
We have only one HTTP reply for now, so I guess we should go with it and then have some ideas for the future.
For instance, I could think of updating plots, e.g. continuing to draw the line for new results… but it is very unclear for now.
Going from Future to Stream in the future with a good concept is possible.

operators/src/engine/query_processor.rs
ChristianBeilschmidt: We could either go with new types or bytes and an image flag. At some point, however, or service needs to reply with the correct mime type. I guess we can refactor this if we know more about our plot ambitions. PNG and JSON are the mapping responses.

operators/src/engine/result_descriptor.rs
ChristianBeilschmidt: Vega rendering, for instance, is some kind of special JSON file. The client would need to render it differently. But I guess this will be more clear if we try to put it together with the front end. For now, it's empty.

operators/src/plot/statistics.rs
```diff
@@ -0,0 +1,261 @@
...
+impl PlotQueryProcessor for StatisticsQueryProcessor {
+    type PlotType = serde_json::Value;
+
+    fn plot_query<'a>(
+        &'a self,
+        query: QueryRectangle,
+        ctx: &'a dyn QueryContext,
+    ) -> BoxFuture<'a, Result<Self::PlotType>> {
+        let mut queries = Vec::with_capacity(self.rasters.len());
+        for (i, raster_processor) in self.rasters.iter().enumerate() {
```
michaelmattig, line 94: maybe it would make sense to process the childs in parallel. Always continuing with the source that has the next tile available the earliest using the select macro

operators/src/util/number_statistics.rs
ChristianBeilschmidt: It only occurs on very little data, I'm not sure, do you think it is worthwhile distinguishing?

operators/src/plot/statistics.rs
ChristianBeilschmidt: I agree, but we first have to find a way to incorporate our pool in queries :smile:

operators/src/plot/statistics.rs
```diff
@@ -0,0 +1,261 @@
...
+        for &value in &raster_tile.grid_array.data {
+            number_statistics.add(value);
+        }
+    }
+}
+
+/// The statistics summary output type for each raster input
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+struct StatisticsOutput {
+    pub count: usize,
```
michaelmattig, line 148: pixel_count would be a more descriptive name

operators/src/util/number_statistics.rs
michaelmattig: I think for rasters it will occure on a lot of data. I think for the API it's fine to ouput NAN. Not sure what's better in backend code. We can leave it if you think it's the correct semantics

michaelmattig:
Please merge master

ChristianBeilschmidt:
> Please merge master

15078af  and 15078af

operators/src/plot/statistics.rs
ChristianBeilschmidt: 4b963e9

operators/src/util/number_statistics.rs
ChristianBeilschmidt: cf. our conversation, we will leave it for now like this.

PR #150, opened 2021-02-10
Plot endpoint

ChristianBeilschmidt:


ChristianBeilschmidt:
[Plot endpoint](https://trello.com/c/QhmReSow/299-plot-endpoint)

services/src/handlers/plots.rs
```diff
@@ -0,0 +1,225 @@
...
+use crate::error;
+use crate::handlers::authenticate;
+use crate::ogc::util::{parse_bbox, parse_time};
+use crate::users::session::Session;
+use crate::util::parsing::parse_spatial_resolution;
+use crate::workflows::registry::WorkflowRegistry;
+use crate::workflows::workflow::WorkflowId;
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub(crate) struct GetPlot {
```
michaelmattig, line 20: isn't this a QueryRectangle?

michaelmattig, line 20: can't you just use `QueryRectangle`?

services/src/handlers/plots.rs
```diff
@@ -0,0 +1,225 @@
...
+                    spatial_reference: SpatialReference::epsg_4326().into(),
+                    measurement: Measurement::Unitless,
+                },
+            },
+        }
+        .boxed()
+    }
+
+    #[tokio::test]
+    async fn json() {
```
michaelmattig, line 143: maybe use Lutz's `check_allowed_http_methods` to ensure anything other than GET requests yield a 405

services/src/handlers/plots.rs
ChristianBeilschmidt: I don't know since I am deserializing from URL parameters. I guess a custom deserializer for the string value in one field for a whole qrect is quite complicated and doesn't make the API better. If we would deserialize from JSON, I would agree.

services/src/handlers/plots.rs
ChristianBeilschmidt: I'm not quite sure why you want this, but here you are:

8012f57

PR #157, opened 2021-02-16
Histogram plot

ChristianBeilschmidt:


ChristianBeilschmidt:
[Histogram](https://trello.com/c/IB3g5DCS/329-histogram)

datatypes/src/plots/mod.rs
```diff
@@ -6,27 +6,40 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, Clone, Deserialize, PartialEq, Serialize)]
+pub struct PlotData {
     pub vega_string: String,
-    pub metadata: M,
+    pub metadata: PlotMetaData,
+}
+
+#[derive(Debug, Clone, Deserialize, PartialEq, Serialize)]
+#[serde(untagged)]
+pub enum PlotMetaData {
```
michaelmattig, line 29: What is the scope of this? Right now you use it for specifying whether (and how) the plot is interactable, but I think this data type is not really extendable, is it? What if there are multiple different interaction types in the future or other meta data like additional text to display. Ithink meta data should be a struct and the interaction a field. But we can also tackle this later on.

operators/src/engine/query_processor.rs
```diff
@@ -87,6 +88,8 @@ where
 pub trait PlotQueryProcessor: Sync + Send {
     type PlotType;
 
+    fn plot_name(&self) -> &'static str;
```
michaelmattig, line 91: is this the correct place to get the name? I mean, layers also have names but what does the query processor have to do with that?

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+///
+/// For vector inputs, it calculates the histogram on one of its attributes.
+///
+pub type Histogram = Operator<HistogramParams>;
+
+/// The parameter spec for `Histogram`
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct HistogramParams {
+    /// Name of the (numeric) attribute to compute the histogram on. Ignored for operation on rasters.
+    pub column_name: Option<String>,
```
michaelmattig, line 35: is this info really necessary here? And if so, also the unit of the column would be interesting. Still it's a bit unelegant to have this field also in raster histograms.

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+}
+
+string_token!(Data, "data");
+
+/// Let the bounds either be computed or given.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum HistogramBounds {
+    Data(Data),
+    Values { min: f64, max: f64 },
```
michaelmattig, line 52: these values are only interesting for drawing the plot, aren't they? Why does the operator need them?

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+
+#[typetag::serde]
+impl PlotOperator for Histogram {
+    fn initialize(
+        self: Box<Self>,
+        context: &dyn ExecutionContext,
+    ) -> Result<Box<InitializedPlotOperator>> {
+        ensure!(
+            self.vector_sources.len() + self.raster_sources.len() == 1,
+            error::InvalidNumberOfVectorInputs {
```
michaelmattig, line 64: this error doesn't really fit here

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+                    .context(error::InvalidOperatorSpec {
+                        reason: "Histogram on vector input is missing `column_name` field"
+                            .to_string(),
+                    })?;
+
+            let vector_result_descriptor = vector_sources[0].result_descriptor();
+
+            match vector_result_descriptor.columns.get(column_name) {
+                None => {
+                    return Err(Error::ColumnDoesNotExist {
```
michaelmattig, line 89: histograms on strings columns are also interesting. Of course only if there are few different values and we might consider that Categorical data, but then we need a convenient way for users to use the operator that way..

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+                        column: column_name.to_string(),
+                    });
+                }
+                Some(FeatureDataType::Categorical) | Some(FeatureDataType::Text) => {
+                    // TODO: incorporate categorical data
```
michaelmattig, line 94: histograms on strings columns are also interesting. Of course only if there are few different values and we might consider that Categorical data, but then we need a convenient way for users to use the operator that way..

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+                    .column_name
+                    .as_ref()
+                    .context(error::InvalidOperatorSpec {
+                        reason: "Histogram on vector input is missing `column_name` field"
+                            .to_string(),
+                    })?;
+
+            let vector_result_descriptor = vector_sources[0].result_descriptor();
+
+            match vector_result_descriptor.columns.get(column_name) {
```
michaelmattig, line 87: could also be expressed as an ensure. Not mandatory though.

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+        )
+    }
+}
+
+impl HistogramRasterQueryProcessor {
+    fn preprocess<'p>(
+        &'p self,
+        query: QueryRectangle,
+        ctx: &'p dyn QueryContext,
+    ) -> BoxFuture<'p, Result<HistogramMetadata>> {
```
michaelmattig, line 229: why not make function async?

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+pub type Histogram = Operator<HistogramParams>;
+
+/// The parameter spec for `Histogram`
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct HistogramParams {
+    /// Name of the (numeric) attribute to compute the histogram on. Ignored for operation on rasters.
+    pub column_name: Option<String>,
+    /// The bounds (min/max) of the histogram.
+    pub bounds: HistogramBounds,
+    /// If the bounds are empty, it is derived from the square-root choice rule.
```
michaelmattig, line 38: "If the number of buckets are undefined" ?

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+impl HistogramRasterQueryProcessor {
+    fn preprocess<'p>(
+        &'p self,
+        query: QueryRectangle,
+        ctx: &'p dyn QueryContext,
+    ) -> BoxFuture<'p, Result<HistogramMetadata>> {
+        fn process_metadata<T: Pixel>(
+            input: BoxStream<Result<RasterTile2D<T>>>,
+            metadata: HistogramMetadataOptions,
+        ) -> BoxFuture<'_, Result<HistogramMetadata>> {
```
michaelmattig, line 233: async?

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+impl HistogramMetadataInProgress {
+    #[inline]
+    fn add_raster_batch<T: Pixel>(&mut self, values: &[T], no_data: Option<T>) {
+        if let Some(no_data) = no_data {
+            for &v in values {
+                if v == no_data {
+                    continue;
+                }
+
+                self.n += 1;
```
michaelmattig, line 430: an optimization would be to compute min/max and bucket number separately, so if users specifies bucket count but not min/max you don't compute bucket count in vein. Maybe something for moonshots.

operators/src/plot/histogram.rs
```diff
@@ -0,0 +1,841 @@
...
+        })
+    }
+
+    async fn process<'p>(
+        &'p self,
+        metadata: HistogramMetadata,
+        query: QueryRectangle,
+        ctx: &'p dyn QueryContext,
+    ) -> Result<<HistogramRasterQueryProcessor as PlotQueryProcessor>::PlotType> {
+        let mut histogram = geoengine_datatypes::plots::Histogram::builder(
```
michaelmattig, line 266: From<Histogrammetadata> for Histogram would be nice

operators/src/engine/query_processor.rs
michaelmattig: I see that it's not plot name but rather plot type. Anyhow my question remains. If the function remains it should be called plot_type

services/src/handlers/plots.rs
```diff
@@ -189,14 +206,87 @@ mod tests {
...
+                    "mean": 3.5,
+                    "stddev": 1.707_825_127_659_933
+                }]
+            })
+            .to_string()
+        );
+    }
+
+    #[tokio::test]
+    async fn chart() {
```
michaelmattig, line 226: I think we have to be very careful with upgrading the vega library because we don't test how the plots actually look.

michaelmattig:
I got the following error:
> thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code: disallowed in param', operators\src\plot\histogram.rs:474:17

for this plot query 

> {"type":"Plot","operator":{"type":"Histogram","params":{"column_name":"featurecla","bounds":"data"},"raster_sources":[],"vector_sources":[{"type":"OgrSource","params":{"data_set":{"Internal":"ac34c4d3-fdaa-48e3-84b3-cef81657d683"},"attribute_projection":null}}]}}

operators/src/plot/histogram.rs
ChristianBeilschmidt: I guess that is a different task, since it is not easy to to it for strings, e.g., if each feature has a different string, you'll get into memory problems.

services/src/handlers/plots.rs
ChristianBeilschmidt: We actually do since the definition is checked in `datatypes`.

datatypes/src/plots/mod.rs
ChristianBeilschmidt: I don't know, you mean it should just be a box of some JSON Value?

operators/src/plot/histogram.rs
ChristianBeilschmidt: e91ddcc

operators/src/plot/histogram.rs
ChristianBeilschmidt: e91ddcc

operators/src/plot/histogram.rs
ChristianBeilschmidt: It has three different options, so I think it is fine this way :shrug: .

operators/src/engine/query_processor.rs
ChristianBeilschmidt: f0638b7

operators/src/plot/histogram.rs
ChristianBeilschmidt: From<(HistogramMetadata, Measurement)>? I'm not sure if that is really better.

operators/src/plot/histogram.rs
ChristianBeilschmidt: 445537d

operators/src/plot/histogram.rs
ChristianBeilschmidt: 0b80abb

operators/src/plot/histogram.rs
ChristianBeilschmidt: 0b80abb

operators/src/plot/histogram.rs
ChristianBeilschmidt: fd3417b 

I guess there are lots of improvements, also for the actual calculation.

ChristianBeilschmidt:
> I got the following error:
> 
> > thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code: disallowed in param', operators\src\plot\histogram.rs:474:17
> 
> for this plot query
> 
> > {"type":"Plot","operator":{"type":"Histogram","params":{"column_name":"featurecla","bounds":"data"},"raster_sources":[],"vector_sources":[{"type":"OgrSource","params":{"data_set":{"Internal":"ac34c4d3-fdaa-48e3-84b3-cef81657d683"},"attribute_projection":null}}]}}
> 
> on the ports data set

The error was because the result descriptor outputted wrong data types.
I fixed it here 
eeadda1  and here 
8deb171 .

datatypes/src/plots/mod.rs
michaelmattig: No, but rather a struct because it can be more easily extended.
````
pub struct PlotMetaData {
   is_interactive: bool,
````

File: operators/src/engine/query_processor.rs (hunk 2 above)
Comment (michaelmattig) [reply operators/src/engine/query_processor.rs]: looks good but what do you think about: output_format: "json/plain" for statistics and "json/vega" for histogram?

File: services/src/handlers/plots.rs (hunk 14 above)
Comment (michaelmattig) [reply services/src/handlers/plots.rs:226]: we only check how the vega string looks like but not how the rendered plot looks like.

File: datatypes/src/plots/mod.rs (hunk 1 above)
Comment (ChristianBeilschmidt) [reply datatypes/src/plots/mod.rs:29]: But then you don't get the name of the interaction component that you need for accessing it from the vega plot.

File: services/src/handlers/plots.rs (hunk 14 above)
Comment (ChristianBeilschmidt) [reply services/src/handlers/plots.rs:226]: Yes, but we can't do this before we haven't figured out how to get a png server-wise.

**Comment by ChristianBeilschmidt** [comment]
> I guess the plot_query method can also be async?

I would like to, but
> functions in traits cannot be declared `async` E0706 Note: `async` trait functions are not currently supported Note: consider using the `async-trait` crate: https://crates.io/crates/async-trait

File: operators/src/engine/query_processor.rs (hunk 2 above)
Comment (ChristianBeilschmidt) [reply operators/src/engine/query_processor.rs]: bf88b3d  6328dea 26e9f77

File: datatypes/src/plots/mod.rs (hunk 1 above)
Comment (michaelmattig) [reply datatypes/src/plots/mod.rs:29]: then rename `PlotMetaData` to `InteractionType` and make
```
pub struct PlotMetaData {
   interaction: InteractionType,
```

but like I said we can also do this later on. maybe the interaction will be the only plot meta data we will ever need.

operators/src/engine/query_processor.rs
michaelmattig: >use vega json spec directly instead of vega lite crate

why do you do this? Seems much less typesafe as before

michaelmattig:
> 
> 
> > I guess the plot_query method can also be async?
> 
> I would like to, but
> 
> > functions in traits cannot be declared `async` E0706 Note: `async` trait functions are not currently supported Note: consider using the `async-trait` crate: https://crates.io/crates/async-trait

We use async_trait everywhere else, so why not here?

operators/src/engine/query_processor.rs
ChristianBeilschmidt: First of all, for the other plot, I had to do it manually, because the crate won't let me do it, but the plot is valid.
Second, the crate provides no IDE support.
Third, I could remove `ndarray` as a dependency.
Fourth, the plot stays valid and you can check it with https://vega.github.io/editor/.

I also thought it was a good thing, but I am quite disappointed.

ChristianBeilschmidt:
> > > I guess the plot_query method can also be async?
> > 
> > 
> > I would like to, but
> > > functions in traits cannot be declared `async` E0706 Note: `async` trait functions are not currently supported Note: consider using the `async-trait` crate: https://crates.io/crates/async-trait
> 
> We use async_trait everywhere else, so why not here?


14b9f60

datatypes/src/plots/mod.rs
ChristianBeilschmidt: I would vote for this. We change it once we have the second type and the current implementation seems odd.

PR #159, opened 2021-02-19
refactor GdalSource to get loading info from meta data provided by ex…

michaelmattig:
…ecution context

https://trello.com/c/tqDJQMal


michaelmattig:
[GdalSource auf MetaData und LoadingInfo umstellen](https://trello.com/c/tqDJQMal/326-gdalsource-auf-metadata-und-loadinginfo-umstellen)

operators/src/source/gdal_source.rs
```diff
@@ -445,75 +287,65 @@ where
...
 
     ///
     /// A stream of `RasterTile2D`
     ///
     pub fn tile_stream(
         &self,
-        bbox: BoundingBox2D,
-        time_interval: TimeInterval,
-        spatial_resolution: SpatialResolution,
+        query: QueryRectangle,
```
jdroenner, line 304: if we use "QueryRectangle" also as STBoundingBox then we should rename it.

operators/src/source/gdal_source.rs
```diff
@@ -445,75 +287,65 @@ where
...
-            TilingStrategy::new_with_tiling_spec(self.tiling_specification, x_signed, y_signed);
+        let tiling_strategy = self.tiling_specification.strategy(x_signed, y_signed);
 
-        stream::iter(self.time_tile_iter(tiling_strategy, time_interval, bbox))
-            .map(move |(time, tile)| {
-                (
-                    self.gdal_params.clone(),
-                    self.dataset_information.clone(),
-                    time,
+        // TODO: avoid cloning loading_info?
```
jdroenner, line 329: where is the clone?

operators/src/source/gdal_source.rs
```diff
@@ -522,8 +354,14 @@ where
         query: crate::engine::QueryRectangle,
         _ctx: &'a dyn crate::engine::QueryContext,
     ) -> BoxStream<Result<RasterTile2D<T>>> {
-        self.tile_stream(query.bbox, query.time_interval, query.spatial_resolution)
-            .boxed() // TODO: handle query, ctx, remove one boxed
+        let meta_data = self.meta_data.loading_info(query).unwrap(); // TODO: change when query signature returns Result
+
+        // TODO: more efficient chaining of streams
```
jdroenner, line 359: what is not efficient here?

operators/src/source/gdal_source.rs
```diff
@@ -445,75 +287,65 @@ where
...
+        stream::iter(tiling_strategy.tile_information_iterator(query.bbox))
+            .map(move |tile| {
+                Ok(RasterTile2D::new_with_tile_info(
+                    info.time,
                     tile,
-                )
-            })
-            .then(|(gdal_params, dataset_information, time, tile)| {
-                Self::load_tile_data_async(gdal_params, dataset_information, time, tile)
+                    Self::load_tile_data(
```
jdroenner, line 335: why not use the async version?

operators/src/source/gdal_source.rs
```diff
@@ -680,19 +504,66 @@ where
...
+
+    if !current_path.ends_with("operators") {
+        current_path = current_path.join("operators");
+    }
+
+    current_path = current_path.join("test-data/raster");
+    current_path
+}
+
+// TODO: move test helper somewhere else?
```
jdroenner, line 523: add a util module or something similar

operators/src/source/gdal_source.rs
```diff
@@ -820,771 +691,281 @@ mod tests {
...
-        let time_interval = TimeInterval::new_unchecked(0, 1000);
+        assert_eq!(x.data.len(), 64);
 
-        let x = gdal_source
-            .load_tile_data(time_interval, tile_information)
-            .unwrap();
-
-        assert_eq!(x.tile_information(), tile_information);
-        assert_eq!(x.time, time_interval);
+        // TODO: compare actual data
```
jdroenner, line 803: add the test

operators/src/source/gdal_source.rs
```diff
@@ -820,771 +691,281 @@ mod tests {
...
-            .collect();
+        let x = load_ndvi_jan_2014(output_shape, output_bounds).unwrap();
 
-        let ndvi_center_pixel_values = vec![
-            19, 255, 255, 43, 76, 17, 255, 255, 255, 145, 255, 255, 255, 255, 255, 255, 255, 255,
-        ];
+        assert_eq!(x.data.len(), 64);
 
-        assert_eq!(upper_left_pixels, ndvi_center_pixel_values);
+        // TODO: compare actual data
```
jdroenner, line 816: add test

operators/src/source/gdal_source.rs
```diff
@@ -820,771 +691,281 @@ mod tests {
...
-        let gdal_source = GdalSourceProcessor {
-            dataset_information: dataset_information_provider,
-            gdal_params,
-            tiling_specification: TilingSpecification {
-                origin_coordinate: Coordinate2D::new(0., 0.),
-                tile_size_in_pixels: GridShape2D::from(tile_size_in_pixels),
-            },
-            phantom_data: PhantomData,
-        };
+    fn test_load_tile_data_overlaps_dataset_bounds() {
```
jdroenner, line 808: is this a test or a helper? please move helpers to the top

operators/src/source/gdal_source.rs
```diff
@@ -820,771 +691,281 @@ mod tests {
...
-            raster_data_root: "../operators/test-data/raster".into(),
-        };
-
-        let gdal_source = GdalSourceProcessor {
-            dataset_information: dataset_information_provider,
-            gdal_params,
-            tiling_specification: TilingSpecification {
-                origin_coordinate: Coordinate2D::new(0., 0.),
-                tile_size_in_pixels: GridShape2D::from(tile_size_in_pixels),
+    async fn query_gdal_source(
```
jdroenner, line 820: test or helper?

operators/src/source/gdal_source.rs
```diff
@@ -820,771 +691,281 @@ mod tests {
...
-        let time_interval_provider = TimeIntervalInformation {
-            start_time: TimeInstance::from_millis(0),
-            time_step: TimeStep {
-                granularity: TimeGranularity::Seconds,
-                step: 1,
-            },
-        };
+    #[tokio::test]
+    async fn test_query_single_time_slice() {
+        let mut exe_ctx = MockExecutionContext::default();
```
jdroenner, line 860: why dues this require a mut exe_ctx?

operators/src/source/gdal_source.rs
```diff
@@ -820,771 +691,281 @@ mod tests {
...
-            file_name_with_time_placeholder: "MOD13A2_M_NDVI_%%%_START_TIME_%%%.TIFF".into(),
-            time_format: "%Y-%m-%d".into(),
-            base_path: "../modis_ndvi".into(),
-            data_type: RasterDataType::U8,
-            geo_transform: dataset_geo_transform,
-            grid_shape: global_size_in_pixels.into(),
-            spatial_ref: SpatialReference::epsg_4326().into(),
-            measurement: Measurement::Unitless,
-        };
+    async fn test_query_multi_time_slices() {
```
jdroenner, line 908: how many time slices does this test?

services/src/contexts/mod.rs
```diff
@@ -89,18 +89,13 @@ impl<D> ExecutionContext for ExecutionContextImpl<D>
 where
     D: DataSetDb
         + MetaDataProvider<MockDataSetDataSourceLoadingInfo, VectorResultDescriptor>
-        + MetaDataProvider<OgrSourceDataset, VectorResultDescriptor>,
+        + MetaDataProvider<OgrSourceDataset, VectorResultDescriptor>
+        + MetaDataProvider<GdalLoadingInfo, RasterResultDescriptor>,
```
jdroenner, line 93: just a thought: we could also think about implementing the actual MEtaDataProviders as separate structs which get a database object. then we can composition them together (or wrap them in an enum or something).

services/src/util/tests.rs
```diff
@@ -124,6 +129,31 @@ pub async fn register_workflow_helper<C: Context>(ctx: &C) -> (Workflow, Workflo
     (workflow, id)
 }
 
+pub async fn add_ndvi_to_datasets(ctx: &InMemoryContext) -> DataSetId {
```
jdroenner, line 132: i think there is already a method with the same/similar name?

jdroenner (commented):
Well, like i said i have some issues with the design choice but we decided to go this way. 
I still have some questions.
Do we want the have the json parser back to import datasets from a description (to a database) at some time?

In general i think we could change some details of the database <-> provider design. Maybe we cold implement a struct for each provider GdalMetaDataProvider<D: Database> and then pass a D to the provider. Then we can put all the providers in a wrapper struct, type list or an enum. Do you think that could work?

michaelmattig:
> 
> 
> Well, like i said i have some issues with the design choice but we decided to go this way.
> I still have some questions.
> Do we want the have the json parser back to import datasets from a description (to a database) at some time?

Yes we have a task for that https://trello.com/c/4OSv3m3r

> In general i think we could change some details of the database <-> provider design. Maybe we cold implement a struct for each provider GdalMetaDataProvider<D: Database> and then pass a D to the provider. Then we can put all the providers in a wrapper struct, type list or an enum. Do you think that could work?

I don't think we need that. In my mind the DataSetDB or DataSetProvider knows how to create the loading infos in the queried interval. E.g. by creating them from a reference time and regular intervals, or by retrieving individual infos from a database for non regular time series. It can create these infos and give them to the Source which doesn't need to know about which type of database/provider created the info.

operators/src/error.rs
```diff
@@ -153,6 +153,12 @@ pub enum Error {
     InvalidDataSetId,
     DataSetLoadingInfoProviderMismatch,
     UnknownDataSetId,
+
+    // TODO: this error should not be propagated to user
+    #[snafu(display("Could not open gdal data set for file path {:?}", file_path))]
```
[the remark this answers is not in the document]
michaelmattig: For a GdalDataSet = a single file

operators/src/source/gdal_source.rs
```diff
@@ -44,163 +42,148 @@ use geoengine_datatypes::{
...
-            buffered_writer,
-            &self.dataset_information,
-        )?)
-    }
+/// Meta data for a regular time series that begins (is anchored) at `start` with multiple gdal data
+/// sets `step` time apart. The `placeholder` in the file path of the data set is replaced with the
+/// queried time in specified `time_format`.
+// TODO: `start` is actually more a reference time, because the time series also goes in
+//        negative direction. Maybe it would be better to have a real start and end time, then
+//        everything before start and after end is just one big nodata raster instead of many
```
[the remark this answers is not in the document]
michaelmattig: I think this is orthogonal to my point. If a regular time series has a start then we can answer everything before the start with nodata. Right now we would create many no data rasters in the regular interval defined by `step`. But yes, EmptyTiles tiles absolutely make sense

operators/src/source/gdal_source.rs
```diff
@@ -44,163 +42,148 @@ use geoengine_datatypes::{
...
-        self.raster_data_root
-            .clone()
-            .join(Self::DEFINITION_SUBPATH)
-            .join(&self.dataset_information.base_path)
-    }
-    fn data_type(&self) -> RasterDataType {
-        self.dataset_information.data_type
-    }
+        let snapped_interval =
+            TimeInterval::new_unchecked(snapped_start, query.time_interval.end()); // TODO: snap end?
```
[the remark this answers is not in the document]
michaelmattig: I think we have to snap to the start of the next interval. so something like `snapped_start + step`. It just makes sense to get the correct validity e.g. for caching

operators/src/source/gdal_source.rs
```diff
@@ -44,163 +42,148 @@ use geoengine_datatypes::{
...
-            lower_right,
-        )
+        let time_interval_iterator = time_iterator
+            .try_as_intervals(self.step)
+            .map(|ti| ti.expect("is a valid time"));
+
+        let info = time_interval_iterator
+            .map(|time| GdalLoadingInfoPart {
+                time,
+                params: self.params.clone().replace_time_placeholder(
```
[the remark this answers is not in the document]
michaelmattig: f7a124f5f7d56e91057afe38d28c04368f3a6f3a

operators/src/source/gdal_source.rs
michaelmattig: I don't use it like this, I need the resolution here, too.

operators/src/source/gdal_source.rs
michaelmattig: comment was no longer valid 55c169fd986cfd5262d748a8e1e5337abc54f7da

operators/src/source/gdal_source.rs
michaelmattig: I wasn't sure whether the streams are created lazily. If they are I can delete the comment.

operators/src/source/gdal_source.rs
michaelmattig: 4c1c5e5ab8f123cf7c8db87f80ed7d381a525ca8

operators/src/source/gdal_source.rs
michaelmattig: 56ec12c9d69d2d3d4457e6a272d0ce91f058b78e

services/src/util/tests.rs
michaelmattig: there ist `create_` which just creates the metadata and `add_` which creates the metadata and adds it to the context

services/src/contexts/mod.rs
michaelmattig: it's possible but then each provider needs e.g. a separate db pool reference etc. even though they all belong to the same database. Also makes the in memory implementation harder imho.

operators/src/source/gdal_source.rs
michaelmattig: b321150062eb7c995d424d0eecbbb09e626a1f49

operators/src/source/gdal_source.rs
michaelmattig: b321150062eb7c995d424d0eecbbb09e626a1f49

operators/src/source/gdal_source.rs
michaelmattig: because I add the data set info to it after creation, so the GdalSource can access it using the meta_data function

operators/src/source/gdal_source.rs
michaelmattig: two slices. Jan/Feb 2014 each with 4 tiles

operators/src/source/gdal_source.rs
michaelmattig: I am not sure how to verfy the results. They are slightly differnent in QGIS but looks good to me 🤓

operators/src/source/gdal_source.rs
michaelmattig: I am not sure how to verify the results. They are slightly differnent in QGIS but looks good to me 🤓

operators/src/source/gdal_source.rs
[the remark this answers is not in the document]
jdroenner: yes two points. Currently we only have regular tiles. We need to see what happens when we produce a very large one.

operators/src/source/gdal_source.rs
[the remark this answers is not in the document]
jdroenner: this only works for regular intervals with a fix step size.

operators/src/source/gdal_source.rs
jdroenner: how can we figure this out?

operators/src/source/gdal_source.rs
jdroenner: well then we need a task to generate valid tests

operators/src/source/gdal_source.rs
jdroenner: well then we need a task to generate valid tests

jdroenner:
> > Well, like i said i have some issues with the design choice but we decided to go this way.
> > I still have some questions.
> > Do we want the have the json parser back to import datasets from a description (to a database) at some time?
> 
> Yes we have a task for that https://trello.com/c/4OSv3m3r
> 
> > In general i think we could change some details of the database <-> provider design. Maybe we cold implement a struct for each provider GdalMetaDataProvider<D: Database> and then pass a D to the provider. Then we can put all the providers in a wrapper struct, type list or an enum. Do you think that could work?
> 
> I don't think we need that. In my mind the DataSetDB or DataSetProvider knows how to create the loading infos in the queried interval. E.g. by creating them from a reference time and regular intervals, or by retrieving individual infos from a database for non regular time series. It can create these infos and give them to the Source which doesn't need to know about which type of database/provider created the info.
> 
> Or do you mean something else?

no i mean the same thing. i just want to put the providers into extra modules to make it more modular. Then we could put them in a typed list (like in config) and then we could change the context to not be a provider but to have a method with get<T>.

operators/src/source/gdal_source.rs
michaelmattig: I traced it:
> running 1 test
create tile stream
load tile
load tile
load tile
load tile
create tile stream
load tile
load tile
load tile
load tile
test source::gdal_source::tests::test_query_multi_time_slices ... ok

so it seems fine to me. I removed the todos 51d34bd866d428217421543782d1298306833141

operators/src/source/gdal_source.rs
michaelmattig: https://trello.com/c/nH4Rw9zK

operators/src/source/gdal_source.rs
michaelmattig: https://trello.com/c/nH4Rw9zK

michaelmattig:
> no i mean the same thing. i just want to put the providers into extra modules to make it more modular. Then we could put them in a typed list (like in config) and then we could change the context to not be a provider but to have a method with get<T>.

sounds interesting. maybe you can make a PR once this PR is through.

operators/src/source/gdal_source.rs
jdroenner: you should validate the looks good values. it is better then no test.

michaelmattig: 2fb88fa6eae13d7e3c6a90eedc19c8ff5dcc5d35

operators/src/source/gdal_source.rs
michaelmattig: 2fb88fa6eae13d7e3c6a90eedc19c8ff5dcc5d35

PR #162, opened 2021-02-24
Temporal Raster Mean Plot

ChristianBeilschmidt:


ChristianBeilschmidt:
[Temporal Area Mean Plot](https://trello.com/c/T1f6sfmr/330-temporal-area-mean-plot)

datatypes/src/primitives/time_interval.rs
```diff
@@ -341,6 +343,14 @@ impl PartialOrd for TimeInterval {
     }
 }
 
+impl TryFrom<TimeInstance> for TimeInterval {
```
michaelmattig, line 346: Why TryFrom? How can this ever fail?

operators/src/plot/temporal_raster_mean_plot.rs
```diff
@@ -0,0 +1,438 @@
...
+use futures::stream::BoxStream;
+use futures::StreamExt;
+use geoengine_datatypes::plots::{AreaLineChart, Plot, PlotData};
+use geoengine_datatypes::primitives::{Measurement, TimeInstance, TimeInterval};
+use geoengine_datatypes::raster::{Pixel, RasterTile2D};
+use serde::{Deserialize, Serialize};
+use snafu::ensure;
+use std::collections::BTreeMap;
+
+pub const TEMPORAL_RASTER_MEAN_PLOT_NAME: &str = "Temporal Area Mean Plot";
```
michaelmattig, line 19: I don't really like the name
- Why did you decide to use an area plot here in the first place? I think it could be optional to fill the area under the line
- The operator only works for rasters, do we want to extend it in the future to handle feature collections? Then we have the Option<FeatureAttribute> again. Otherwise the word raster should be mentioned so users can find it better. Also: do we want to call every plot XyZPlot? Clippy would be furious. ;) I would suggest "Mean Raster Pixel Values over Time"

operators/src/plot/temporal_raster_mean_plot.rs
```diff
@@ -0,0 +1,438 @@
...
+
+/// A plot that shows the mean values of rasters over time as an area plot.
```
michaelmattig, line 21: Why stop at mean btw? Do we want to create separate operators for min/max/median over time? Maybe it would be nice to have a StatisticsOverTime plot where users can choose from various statistics, and the selected ones are all plotted into a line chart? This is of course beyond the scope of this PR but a general decision we have to make. Lots of plots, or few rich ones.

operators/src/plot/temporal_raster_mean_plot.rs
```diff
@@ -0,0 +1,438 @@
...
+pub type TemporalRasterMeanPlot = Operator<TemporalRasterMeanPlotParams>;
+
+/// The parameter spec for `TemporalRasterMeanPlot`
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct TemporalRasterMeanPlotParams {
+    /// Where should the x-axis (time) tick be positioned?
+    /// At either time start, time end or in the center.
+    time_position: TemporalRasterMeanPlotTimePosition,
```
michaelmattig, line 29: if we'd want to be correct we'd have at least 2 data points for each raster time slice at beginning and end because in essence we have a stepping function of valid data points. it's a bit dangerous to connect the points like you do because it implies there are values inbetween

operators/src/plot/temporal_raster_mean_plot.rs
```diff
@@ -0,0 +1,438 @@
...
+        for (timestamp, mean_calculator) in means {
+            timestamps.push(timestamp);
+            values.push(mean_calculator.mean());
+        }
+
+        AreaLineChart::new(timestamps, values, measurement).map_err(Into::into)
+    }
+}
+
+struct MeanCalculator {
```
michaelmattig, line 180: why not reuse the stuff from the Statistics operator?

operators/src/plot/temporal_raster_mean_plot.rs
```diff
@@ -0,0 +1,438 @@
...
+        position: TemporalRasterMeanPlotTimePosition,
+    ) -> Result<BTreeMap<TimeInstance, MeanCalculator>> {
+        let mut means: BTreeMap<TimeInstance, MeanCalculator> = BTreeMap::new();
+
+        while let Some(tile) = tile_stream.next().await {
+            let tile = tile?;
+
+            let time = Self::time_interval_projection(tile.time, position);
+
+            let mean = means.entry(time).or_default();
```
michaelmattig, line 142: maybe initialize the Calculators once and then just expect them here. Might save millions of checks whether it's there. Or maybe the complier is smart enough.

operators/src/plot/temporal_raster_mean_plot.rs
ChristianBeilschmidt: It calculates more than the mean. In the last PR you made me split up a special case where it was possible to do less work, so… :shrug:.

operators/src/plot/temporal_raster_mean_plot.rs
ChristianBeilschmidt: I have to do this at least for every time step and I don't know them in advance. So I can't pull it out of the loop. I have to check it at least once for every tile and this is done here.

operators/src/plot/temporal_raster_mean_plot.rs
ChristianBeilschmidt: Well, the user has to know the data. We can't make it correct here for every use case. For the ebv case, for instance, it seemed to be right this way. This way you can see a trend if you like. A bar chart for every time step is another plot type to consider.

operators/src/plot/temporal_raster_mean_plot.rs
ChristianBeilschmidt: You have several options, also to compare multiple datasets in one plot. I would say we extend the operator based on demands :shrug:. In the end we want to be able to combine stuff, but I guess we can't do it prematurely without use case because of other functionality.

operators/src/plot/temporal_raster_mean_plot.rs
ChristianBeilschmidt: > I don't really like the name
> 
>     * Why did you decide to use an area plot here in the first place? I think it could be optional to fill the area under the line
Because it is the demand of a use case.

>     * The operator only works for rasters, do we want to extend it in the future to handle feature collections? Then we have the Option again. Otherwise the word raster should be mentioned so users can find it better. Also: do we want to call every plot XyZPlot? Clippy would be furious. ;) I would suggest "Mean Raster Pixel Values over Time"
Yes, I would like to have something for features, but this is much more complicated, since each feature has its own validity. We have to specify buckets/steps there.

>     * The disctinction between name and plot is still a bit fuzzy imho. What do we show to the user und how do we distinguish plot types?
I guess we can use a user friendly name for that.

datatypes/src/primitives/time_interval.rs
ChristianBeilschmidt: 408f387

operators/src/plot/temporal_raster_mean_plot.rs
michaelmattig: True, but we also have to think about the future proofness of the operators. Once there are workflows using a plot we have to keep it around forever. And we can't just add min/max values to a plot named "Mean Plot". Then we'd have at some point a general StatisticsOverTime and a special Mean plot.

Maybe this is a separate task "Polish API" for a 1.0 release though and until then we don't care too much about it?

ChristianBeilschmidt: Yeah, I think we aren't there yet. We could also deprecate operators and switch all occurrences of the old one with a spec of the new one.

PR #168, opened 2021-02-26
query method now returns a result

ChristianBeilschmidt:


ChristianBeilschmidt:
[Refactor query processor trait to return `Result<BoxStream<..>>`](https://trello.com/c/Q3jlJKiT/249-refactor-query-processor-trait-to-return-resultboxstream)

operators/src/processing/vector_join/equi_data_join.rs
```diff
@@ -358,7 +358,10 @@ where
                         Err(e) => return stream::once(async { Err(e) }).boxed(),
                     };
 
-                    let data_query = self.right_processor.query(query, ctx);
+                    let data_query = match self.right_processor.query(query, ctx) {
```
michaelmattig, line 361: So the left stream produced a valid tile but the right stream could not be created. I think the whole `vector_query` should fail instead of creating a single error tile.

operators/src/processing/expression.rs
```diff
@@ -241,12 +241,13 @@ where
         &'b self,
         query: QueryRectangle,
         ctx: &'b dyn QueryContext,
-    ) -> BoxStream<'b, Result<RasterTile2D<TO>>> {
+    ) -> Result<BoxStream<'b, Result<RasterTile2D<TO>>>> {
```
michaelmattig, line 244: wouldn't it makes ense to fail here now if e.g. the CL program could not be run or parameters could not be set? Instead of panic.

operators/src/source/ogr_source.rs
```diff
@@ -289,13 +289,13 @@ where
         &'a self,
         query: QueryRectangle,
         ctx: &'a dyn QueryContext,
-    ) -> BoxStream<'a, Result<Self::Output>> {
-        OgrSourceStream::new(
-            self.dataset_information.loading_info(query).unwrap(), // TODO: handle error
+    ) -> Result<BoxStream<'a, Result<Self::Output>>> {
+        Ok(OgrSourceStream::new(
+            self.dataset_information.loading_info(query)?, // TODO: handle error
```
michaelmattig, line 294: remove todo

operators/src/processing/vector_join/equi_data_join.rs
ChristianBeilschmidt: You can either fail in the method or within the stream.

We can't do the first thing, since we create multiple query processors within the stream.
Each of them could fail and we can't create them in advance.

When we are in the stream, we can't suddenly hop out and let `vector_query` crash.

operators/src/processing/expression.rs
ChristianBeilschmidt: Also here, that is inside the stream and has nothing to do with `query`.

operators/src/source/ogr_source.rs
ChristianBeilschmidt: 2decdde

PR #171, opened 2021-03-01
handle no_data in gdal_source

jdroenner:


jdroenner:
[GDALSource setzt kein NO DATA](https://trello.com/c/9314zlwE/360-gdalsource-setzt-kein-no-data)

operators/src/source/gdal_source.rs
```diff
@@ -254,6 +254,11 @@ where
         let rasterband: GdalRasterBand =
             dataset.rasterband(data_set_params.rasterband_channel as isize)?;
 
+        // get the no_data_value. TODO: move to dataset parameters? Keep in mind, that this can change for each time step.
```
michaelmattig, line 257: I think we should add No Data value to GdalParameters s.t. we have all the meta data in one place.

operators/src/source/gdal_source.rs
```diff
@@ -254,6 +254,11 @@ where
...
+        // We also need to ge metadata from the dataset (for each tile) e.g. scale + offset.
```
michaelmattig, line 258: what do you mean by that? To turn raw data into a measurement? Why don't we use the expression operator?

Also type: "ge" -> "get"

operators/src/source/gdal_source.rs
```diff
@@ -931,11 +938,13 @@ mod tests {
...
+        let tile_1 = &c[0];
+
         assert_eq!(
-            c[0].time,
+            tile_1.time,
             TimeInterval::new_unchecked(1_385_856_000_000, 1_388_534_400_000)
         );
 
-        assert!(!c[0].grid_array.data.iter().any(|p| *p != 0)); // TODO: use actual no data value
+        assert!(!tile_1.grid_array.data.iter().any(|p| *p != 0)); // TODO: the NDVI data has no NO DATA value. Currently, we fill empty areas with 0 if there is no NO DATA. We have to add a strategy to fix this since we meight add empty areas through the tiling approach.
```
michaelmattig, line 948: with our tiling approach we will actually almost certainly have no data values because the outer tile have always pixels outside of the valid CRS bounds. I think we need to enforce a no data value for all data sets. This might mean to use the next bigger data type if all the values are used up by the measurement.

Also: typo: "meight" -> might

operators/src/source/gdal_source.rs
```diff
@@ -254,6 +254,11 @@ where
         let rasterband: GdalRasterBand =
             dataset.rasterband(data_set_params.rasterband_channel as isize)?;
 
+        // get the no_data_value. TODO: move to dataset parameters? Keep in mind, that this can change for each time step.
+        // We also need to ge metadata from the dataset (for each tile) e.g. scale + offset.
+        let no_data_value = rasterband.no_data_value().map(T::from_);
+        let fill_value = no_data_value.unwrap_or_else(T::zero);
```
michaelmattig, line 260: I think we should enforce a no data value. cf. my other comment

operators/src/source/gdal_source.rs
jdroenner: yes but what happens if the no-data value changes over time?

operators/src/source/gdal_source.rs
jdroenner: yes i don't want the source to do the scaling but e.g. for meteosat there are offset and scale values which are different for each timestep. will we store them in the database too?

operators/src/source/gdal_source.rs
jdroenner: how would we do this? we need to open the dataset to check if there is a no-data-value. but to open the dataset we need the type...

operators/src/source/gdal_source.rs
jdroenner: yes but i dno'T know how to do this at the moment.

operators/src/source/gdal_source.rs
michaelmattig: with "over time" you mean you have different no-data values for different geotiffs? Well, you get one `GdalDataSetParameters` for each file and the meta data provider has to know the value somehow. If there are situations where there is no way of knowing it other than reading the tiff meta data, then we maybe need to enhance the `GdalDataSetParameters` to have all parameters as Options, and if they are None they are read from the tiff meta data.

operators/src/source/gdal_source.rs
michaelmattig: that's an interesting case I wasn't aware of. First of all I would transform the time series once into the measurement instead of doing it every time. But of course we want to be able to do that in geo engine. We either need to enhance the raster tile meta data to carry this meta data and use it in the expression or we need some raster - datacollections join where we get the scaling info from somewhere and use it in the operator.

operators/src/source/gdal_source.rs
michaelmattig: In my opinion the datasetdb should know the no-data value for every data set in the system. At some point it gets added/imported and then we can specify and store it.

operators/src/source/gdal_source.rs
michaelmattig: yeah, just create a New Idea task for it.

operators/src/source/gdal_source.rs
jdroenner: that could work

operators/src/source/gdal_source.rs
jdroenner: i think that is not a good idea. First, this will blow up the required storage space. We will never store all the transformed MSG data. The wavelet encoded original data is already ~100TB large. And second: sometimes i want the real raw data and somteimes i want to transform it.

The RasterTile needs a map for metadata that is correct. Do you want to store the metadata in the database? Where will it come from? Also if you want to transform the data once, we need to read and transform it... So we have to read the metadata in any case and we need to handle this somewhere.

operators/src/source/gdal_source.rs
jdroenner: where is it imported? What kind of operator does this? Is this a geoengine2 which can handle this process?

operators/src/source/gdal_source.rs
jdroenner: ok

operators/src/source/gdal_source.rs
michaelmattig: I would store them in the database if they are unique for each tiff file and otherwise compute them in the `loading_info` method based on some meta data in the database.

operators/src/source/gdal_source.rs
michaelmattig: I'll let you decide if you implement it now or create a new task for it

operators/src/source/gdal_source.rs
michaelmattig: at some point you have to say here is a raster time series whose files lie in path x and this is how to get from a query time to a file name. That's where we can also say the no data value is always x. or the no data value is x before date y and after that it's y.

operators/src/source/gdal_source.rs
michaelmattig: and link it here so I can resolve the discussion ;)

jdroenner: https://trello.com/c/pjchxCTQ

operators/src/source/gdal_source.rs
jdroenner: we need to add a gdal metadata reader. so we should do it as a new task. https://trello.com/c/uddqQmtx

operators/src/source/gdal_source.rs
jdroenner: but who inserts them into the database?

operators/src/source/gdal_source.rs
jdroenner: i guess we will need to scan the files and a) use the metadata from that scan directly or b) store them in the database.

operators/src/source/gdal_source.rs
jdroenner: da3aee9

ChristianBeilschmidt:
You guys both missed that missed:
https://github.com/geo-engine/geoengine/blob/da3aee9fda623f6396ae30db4bae111493073e64/operators/src/source/gdal_source.rs#L235-L246

jdroenner:
> You guys both missed that missed:
> 
> https://github.com/geo-engine/geoengine/blob/da3aee9fda623f6396ae30db4bae111493073e64/operators/src/source/gdal_source.rs#L235-L246

778af17

PR #172, opened 2021-03-01
first step to a simple InitializedOperator

jdroenner:
remove params from InitializedOperatorImpl

### DIFF
--- a/operators/src/processing/point_in_polygon.rs
+++ b/operators/src/processing/point_in_polygon.rs
@@ -1,5 +1,6 @@
 use futures::stream::BoxStream;
 use futures::{StreamExt, TryStreamExt};
+use serde::{Deserialize, Serialize};
 use snafu::ensure;
 
 use geoengine_datatypes::collections::{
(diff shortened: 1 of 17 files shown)


jdroenner:
[Refactor InitializedOperatorImpl](https://trello.com/c/ZgQvJKhw/177-refactor-initializedoperatorimpl)

operators/src/processing/point_in_polygon.rs
```diff
@@ -77,7 +76,7 @@ impl VectorOperator for PointInPolygonFilter {
     }
 }
 
-pub type InitializedPointInPolygonFilter = InitializedOperatorImpl<(), VectorResultDescriptor, ()>;
+pub type InitializedPointInPolygonFilter = InitializedOperatorImpl<VectorResultDescriptor, ()>;
```
ChristianBeilschmidt, line 79: I guess we should use the moment to introduce an empty dummy struct here as state.

ChristianBeilschmidt:
I guess that is an improvement :+1:. However, the task states that we tried to move from *Lambda function to builder pattern*. This is not in this PR, yet.

ChristianBeilschmidt (changes requested):

jdroenner:
i know but i had to start somewhere.

jdroenner:
will add builder later

ChristianBeilschmidt (approved):

PR #173, opened 2021-03-01
prevent ogr source use of Feature iterator after finish

michaelmattig:


michaelmattig:
[prevent ogr source use of Feature iterator after finish](https://trello.com/c/klbtdsfR/374-prevent-ogr-source-use-of-feature-iterator-after-finish)

PR #175, opened 2021-03-02
add data set definition for polygon collection

michaelmattig:


michaelmattig:
[Germany boundaries as polygon dataset](https://trello.com/c/C0FHkAeW/377-germany-boundaries-as-polygon-dataset)

ChristianBeilschmidt:
Ok, but what about https://github.com/geo-engine/geoengine/blob/master/operators/test-data/vector/germany_polygon.json?

michaelmattig:
what about it? do you not use it in the operators crate? Shall I delete it?

PR #179, opened 2021-03-10
replace layer info with symbology

michaelmattig:


michaelmattig:
[Raster- und Vektor-Symbology im Backend speichern und Frontend anpassen](https://trello.com/c/guxEfO4E/421-raster-und-vektor-symbology-im-backend-speichern-und-frontend-anpassen)

datatypes/src/operations/image/colorizer.rs
```diff
@@ -265,6 +268,7 @@ impl Colorizer {
             Self::Palette {
                 colors,
                 no_data_color,
+                default_color: _,
```
ChristianBeilschmidt, line 271: Isn't it meaningless to have a `default_color` that isn't used?

services/src/projects/project.rs
```diff
@@ -213,21 +215,116 @@ pub enum LayerType {
...
+pub struct TextSymbology {
+    pub attribute: String,
+    pub fill_color: ColorParam,
+    pub stroke: StrokeParam,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
+pub struct PointSymbology {
+    pub radius: NumberParam,
+    pub clustered: bool,
```
ChristianBeilschmidt, line 250: I'm not sure if we should have clustered here, since the change of symbology changes the processing (?).
I guess clustering can be expressed as derived radius by attribute and text by attribute.

ChristianBeilschmidt (changes requested):
just 2 things

datatypes/src/operations/image/colorizer.rs
michaelmattig: I think a default color makes sense, because there may be values that are neither defined in the palette nor the no data value. Thus the Platte has to be adjusted, should I do it as part of this task though?

services/src/projects/project.rs
michaelmattig: That's a neat idea but currently I only support linear scaling of a feature attribute to the radius. You use some logarithmic transformation in the clustering algorithm right? Either we need to add the radius as an additional attribute then, or need a more sophisticated attribute to radius derivation.

datatypes/src/operations/image/colorizer.rs
ChristianBeilschmidt: It isn't about having a default color, but this is ignored here and thus, it is useless. I guess you should incorporate it into the palette.

services/src/projects/project.rs
ChristianBeilschmidt: No, the algorithm outputs pixel widths directly, thus it is a linear mapping of the value.

datatypes/src/operations/image/colorizer.rs
michaelmattig: 362b10755e22f0af0999cec496d7e1bc379832a6

services/src/projects/project.rs
michaelmattig: f1cadf4a3d888b01e3c63a66ad55fb940b508e78

datatypes/src/operations/image/colorizer.rs
ChristianBeilschmidt: I somehow can't refer to it properly, but you now output the default color rather then the no data color color when you have NaN as palette value.

michaelmattig: cf our discussion

PR #180, opened 2021-03-10
add open options for csv files

michaelmattig:


michaelmattig:
[Test OgrSource with CSV file](https://trello.com/c/x1b46cuS/438-test-ogrsource-with-csv-file)

operators/src/source/ogr_source.rs
```diff
@@ -361,6 +364,50 @@ where
         }
     }
 
+    fn open_gdal_dataset(dataset_info: &OgrSourceDataset) -> Result<Dataset> {
+        // TODO: reliably detect CSV files or allow defining them as such in params
+        if let Some(ext) = dataset_info.file_name.extension() {
+            if ext == "csv" || ext == "tsv" {
+                let columns = dataset_info
+                    .columns
```
ChristianBeilschmidt, line 372: Maybe we can split that into two functions to make it better readable and then do something like this:
```rust
match dataset_info.file_name.extension().and_then(OsStr::to_str) {
    Some("csv") | Some("tsv") => open_csv_dataset(dataset_info),
    _ => Ok(Dataset::open(&dataset_info.file_name)?),
}
```

operators/src/source/ogr_source.rs
```diff
@@ -361,6 +364,50 @@ where
...
+                        None,
+                    )?);
+                }
+
+                return Ok(Dataset::open_ex(
+                    &dataset_info.file_name,
+                    None,
+                    None,
+                    Some(&[
+                        &("GEOM_POSSIBLE_NAMES=".to_owned() + &columns.x),
```
ChristianBeilschmidt, line 400: I think you should use `format!` instead.

ChristianBeilschmidt (changes requested):
Looks really good, I have some issues regarding readability when I'm thinking at debugging in the future :smile:.

operators/src/source/ogr_source.rs
michaelmattig: 992dbbe0c995dc1ba9aef4947614ba1dbd93eb96

operators/src/source/ogr_source.rs
michaelmattig: 992dbbe0c995dc1ba9aef4947614ba1dbd93eb96

ChristianBeilschmidt (approved):

PR #189, opened 2021-03-31
auto detect loading info for uploaded dataset

michaelmattig:


michaelmattig:
[Import GeoJSON](https://trello.com/c/PEQWEkHp/48-import-geojson)

datatypes/src/collections/data_types.rs
```diff
@@ -26,6 +28,25 @@ pub enum VectorDataType {
     MultiPolygon,
 }
 
+impl VectorDataType {
+    pub fn try_from_ogr_type_code(code: u32) -> Result<Self> {
+        Ok(match code {
+            OGRwkbGeometryType::wkbPoint | OGRwkbGeometryType::wkbMultiPoint => {
```
ChristianBeilschmidt, line 34: if it is called try_from_…, you should implement `TryFrom`

services/src/datasets/storage.rs
```diff
@@ -129,8 +131,28 @@ pub struct CreateDataSet {
...
+}
+
+impl UserInput for AutoCreateDataSet {
+    fn validate(&self) -> Result<()> {
+        // TODO: more sophisticated input validation
+        ensure!(
+            !self.main_file.contains('/') && !self.main_file.contains(".."),
+            error::InvalidUploadFileName
+        );
+
```
ChristianBeilschmidt, line 149: you can add that `dataset_name`, as well as `main_file`, shouldn't be empty.
Of course, more sophisticated checks can come later.

services/src/handlers/datasets.rs
```diff
@@ -41,6 +58,31 @@ async fn list_datasets<C: Context>(
     Ok(warp::reply::json(&list))
 }
 
+pub(crate) fn get_dataset_handler<C: Context>(
```
ChristianBeilschmidt, line 61: Maybe you can briefly add what this handler does.

services/src/handlers/datasets.rs
```diff
@@ -97,8 +139,189 @@ fn adjust_user_path_to_upload_path(meta: &mut MetaDataDefinition, upload: &Uploa
...
+    let create = create.validated()?.user_input;
+
+    let main_file_path = upload.id.root_path()?.join(&create.main_file);
+    let meta_data = auto_detect_dataset(&main_file_path)?;
+
+    let properties = AddDataSet {
+        id: None,
+        name: create.dataset_name,
+        description: create.dataset_description,
+        source_operator: "OgrSource".to_owned(),
```
ChristianBeilschmidt, line 174: I guess this should also be output of `auto_detect_dataset`.

services/src/handlers/datasets.rs
```diff
@@ -97,8 +139,189 @@ fn adjust_user_path_to_upload_path(meta: &mut MetaDataDefinition, upload: &Uploa
...
+    Ok(warp::reply::json(&IdResponse::from(id)))
+}
+
+fn auto_detect_dataset(main_file_path: &Path) -> Result<MetaDataDefinition> {
+    let mut dataset = Dataset::open(&main_file_path).context(error::Gdal)?;
+    let layer = {
+        if let Ok(layer) = dataset.layer(0) {
+            layer
+        } else {
+            // TODO: handle Raster datasets as well
```
ChristianBeilschmidt, line 192: Don't raster datasets have layers as well? They have no features, but layers I guess.

services/src/handlers/datasets.rs
```diff
@@ -97,8 +139,189 @@ fn adjust_user_path_to_upload_path(meta: &mut MetaDataDefinition, upload: &Uploa
...
+            return Err(crate::error::Error::DataSetHasNoAutoImportableLayer);
+        }
+    };
+    let vector_type = detect_vector_type(&layer)?;
```
ChristianBeilschmidt, line 196: Coudln't you prepare a branching, e.g. is_feature_collection or is_raster or is_something_different?
And then implement only the first and return Err on the others.

services/src/handlers/datasets.rs
```diff
@@ -97,8 +139,189 @@ fn adjust_user_path_to_upload_path(meta: &mut MetaDataDefinition, upload: &Uploa
...
+
+    Ok(MetaDataDefinition::OgrMetaData(StaticMetaData {
+        loading_info: OgrSourceDataset {
+            file_name: main_file_path.into(),
+            layer_name: layer.name(),
+            data_type: Some(vector_type),
+            time: OgrSourceDatasetTimeType::None, // TODO: auto detect time type and corresponding columns
+            columns: Some(OgrSourceColumnSpec {
+                x: "".to_owned(), // TODO: for csv-files: try to find wkt/xy columns
+                y: None,
```
ChristianBeilschmidt, line 213: Does Ogr output geometry columns in general?

services/src/handlers/datasets.rs
```diff
@@ -97,8 +139,189 @@ fn adjust_user_path_to_upload_path(meta: &mut MetaDataDefinition, upload: &Uploa
...
+
+struct Columns {
+    decimal: Vec<String>,
+    numeric: Vec<String>,
+    textual: Vec<String>,
+}
+
+fn detect_columns(layer: &Layer) -> HashMap<String, FeatureDataType> {
+    let mut columns = HashMap::default();
+    // TODO: avoid reading ALL of the data
```
ChristianBeilschmidt, line 249: Couldn't you use the [`defn`](https://docs.rs/gdal/0.7.2/gdal/vector/struct.Layer.html#method.defn) method that returns  [`Defn`](https://docs.rs/gdal/0.7.2/gdal/vector/struct.Defn.html), which has a method [`fields`](https://docs.rs/gdal/0.7.2/gdal/vector/struct.Defn.html#method.fields)?

services/src/handlers/datasets.rs
```diff
@@ -97,8 +139,189 @@ fn adjust_user_path_to_upload_path(meta: &mut MetaDataDefinition, upload: &Uploa
...
+                        columns.insert(name, FeatureDataType::Number);
+                    }
+                }
+                gdal::vector::FieldValue::StringValue(_) => {
+                    columns.insert(name, FeatureDataType::Text);
+                }
+                gdal::vector::FieldValue::DateValue(_)
+                | gdal::vector::FieldValue::DateTimeValue(_) => {
+                    // TODO: ensure all feature's field by this name are a date/datetime
+                    // TODO: auto detect if this is start/end date/datetime
```
ChristianBeilschmidt, line 278: I guess we don't. We should just assume it and then we have our error handling, for instance, set null.

ChristianBeilschmidt (changes requested):

datatypes/src/collections/data_types.rs
michaelmattig: TryFrom<u32> doesn't make much sense though.

services/src/handlers/datasets.rs
michaelmattig: it gets a dataset by id... I'd like to do proper documentation of the API in a separate task

services/src/handlers/datasets.rs
michaelmattig: I thought about it but I think it makes more sense to output only the DataSetId, because we not always need the dataset after creation and then we would needlessly send out this information. It's also more in line with how the other API calls work

services/src/handlers/datasets.rs
michaelmattig: well the layer() method calls some OGR_ method internally so I figure you only use this for vector datasets and use rasterbands/channels for rasters

services/src/handlers/datasets.rs
michaelmattig: it does exactly that already

services/src/handlers/datasets.rs
michaelmattig: that is a hard to find method 😅 yes, makes absolutely sense c8e6fe123441d3cfe993329a1edf7663afe85fbd

services/src/handlers/datasets.rs
michaelmattig: c8e6fe123441d3cfe993329a1edf7663afe85fbd

services/src/datasets/storage.rs
michaelmattig: f302348cdda31a2953801ac4e9dd025a2e559586

services/src/handlers/datasets.rs
ChristianBeilschmidt: But now, you just assume that after `auto_detect_dataset` it is always OgrSource.

services/src/handlers/datasets.rs
ChristianBeilschmidt: Maybe you can use the information in https://docs.rs/gdal/0.7.2/gdal/vector/struct.Defn.html#method.geom_fields to get the columns.

services/src/handlers/datasets.rs
michaelmattig: 73b4e8430add8c17b2e5670d7a011c6fbc3fa5e4

services/src/handlers/datasets.rs
michaelmattig: 73b4e8430add8c17b2e5670d7a011c6fbc3fa5e4

services/src/handlers/datasets.rs
michaelmattig: 73b4e8430add8c17b2e5670d7a011c6fbc3fa5e4

PR #192, opened 2021-04-07
BoT/EoT chrono adaption

ChristianBeilschmidt:
* We now don't run into `chrono` panics when using TimeInterval::default()
* I had to fix the time snap when it ran out of bounds
* The raster-vector-join now emits nulls when querying with points that have TimeInterval::default(). It is quite slow because the GdalSource seems to collect every time interval from BoT to EoT. The aggregator, however, returns early if everything is null.
* I changed `From<i64>` to `TryFrom<i64>` for `TimeInstance`. This is more correct but somewhat unergonomic.


ChristianBeilschmidt:
[Begin of time und EoT auf Chrono anpassen](https://trello.com/c/J4Kd9IlK/52-begin-of-time-und-eot-auf-chrono-anpassen)

datatypes/src/primitives/time_interval.rs
```diff
@@ -35,8 +37,8 @@ impl Default for TimeInterval {
     /// ```
     fn default() -> Self {
         Self {
-            start: i64::min_value().into(),
-            end: i64::max_value().into(),
+            start: chrono::MIN_DATETIME.into(),
```
jdroenner, line 40: can we use the min max from TimeInstancE?

jdroenner:
i think it is fine. please unify the source of  max min values.

datatypes/src/primitives/time_interval.rs
ChristianBeilschmidt: 5a4d480

ChristianBeilschmidt:
0fc1dff changes the Vec to Iterator of tile parts such that the CI runs through.

PR #196, opened 2021-04-14
add metadata suggestion

michaelmattig:


michaelmattig:
[Geführter Upload-Dialog](https://trello.com/c/swZsuCeQ/66-gef%C3%BChrter-upload-dialog)

ChristianBeilschmidt (approved):

operators/src/source/ogr_source.rs
```diff
@@ -708,6 +708,7 @@ where
                 FeatureDataType::Text => {
                     let text_option = match field {
                         Ok(FieldValue::IntegerValue(v)) => Some(v.to_string()),
+                        Ok(FieldValue::Integer64Value(v)) => Some(v.to_string()),
```
ChristianBeilschmidt, line 711: Wouldn't it be possible to also map datetime to string if it is set?
That was one of the problems I guess.

services/src/handlers/datasets.rs
```diff
@@ -282,7 +287,10 @@ fn auto_detect_meta_data_definition(main_file_path: &Path) -> Result<MetaDataDef
         result_descriptor: VectorResultDescriptor {
             data_type: vector_type,
             spatial_reference: spatial_reference.into(),
-            columns: columns_map,
+            columns: columns_map
+                .into_iter()
+                .filter_map(|(k, v)| v.try_into().map(|v| (k, v)).ok())
```
ChristianBeilschmidt, line 292: maybe add a comment that you ignore all columns here that don't have a corresponding GE type

services/src/handlers/datasets.rs
```diff
@@ -331,7 +339,7 @@ fn detect_time_type(layer: &Layer, columns: &Columns) -> OgrSourceDatasetTimeTyp
 
     let mut start = None;
     let mut end = None;
-    for column in &(columns.textual) {
+    for column in columns.textual.iter().chain(&(columns.date)) {
```
ChristianBeilschmidt, line 342: This is strange since if you have a date column, you know that this is a date.
But here, it can fail (and it does?).

I guess you should at least see if `into_string` transforms dates into rfc3339 dates. Otherwise, it will always fail.

services/src/handlers/datasets.rs
```diff
@@ -392,46 +400,71 @@ struct Columns {
...
-                numeric.push(k.clone());
-            }
-            FeatureDataType::Text => {
-                textual.push(k.clone());
-            }
+            ColumnDataType::Decimal => decimal.push(k.clone()),
+            ColumnDataType::Number => numeric.push(k.clone()),
+            ColumnDataType::Text => textual.push(k.clone()),
+            ColumnDataType::Date => date.push(k.clone()),
+            _ => {}
```
ChristianBeilschmidt, line 459: I guess the only one left is Unknown. If you leave it here, we can extend it and see, that we have to do something here. The wildcard would match everything else.

ChristianBeilschmidt (changes requested):

ChristianBeilschmidt:
Is it possible to specify the time manually for the hlhl test data? I specified `time_from`, `time_to` but I can't see any data (it doesn't fail).

operators/src/source/ogr_source.rs
michaelmattig: it is possible, but if GDAL detects your chemical element name as a date then I think it's better to output nothing (or maybe an error-string) but not a date if we present it as a text column to the user.

services/src/handlers/datasets.rs
michaelmattig: only because Gdal thinks the column is a datetime it doesn't mean that it is an ISO time. And I only suggest columns containing ISO datetime as a start or end column because I don't know the format String I would have to provide to the user in the OgrSourceTimeFormat

michaelmattig:
> Is it possible to specify the time manually for the hlhl test data? I specified `time_from`, `time_to` but I can't see any data (it doesn't fail). 

it might not be possible because the time_from time_to are not actual fields/properties of the data but inherent datetime properties that e.b. are used in the Ogr time filter. But we have a task for finding that out in trello

operators/src/source/ogr_source.rs
michaelmattig: it is possible, but if GDAL detects your chemical element name as a date then I think it's better to output nothing (or maybe an error-string) but not a date if we present it as a text column to the user.

services/src/handlers/datasets.rs
michaelmattig: only because Gdal thinks the column is a datetime it doesn't mean that it is an ISO time. And I only suggest columns containing ISO datetime as a start or end column because I don't know the format String I would have to provide to the user in the OgrSourceTimeFormat

services/src/handlers/datasets.rs
michaelmattig: c5ffd6ec608bf584f64d7f98f94be3c48e545821

services/src/handlers/datasets.rs
michaelmattig: c5ffd6ec608bf584f64d7f98f94be3c48e545821

services/src/handlers/datasets.rs
ChristianBeilschmidt: Are dates then really parsed by that format string or is it just the date used then? In the latter case, we could do it anyway.

michaelmattig: we always parse dates ourselves and don't use the Gdal parsing. https://github.com/geo-engine/geoengine/blob/master/operators/src/source/ogr_source.rs#L491

ChristianBeilschmidt (approved):

PR #199, opened 2021-04-16
ogc_reprojection

michaelmattig:


michaelmattig:
[Auto-Projektion im UI](https://trello.com/c/jBQAJLob/88-auto-projektion-im-ui)

datatypes/src/primitives/time_interval.rs
```diff
@@ -299,6 +299,10 @@ impl TimeInterval {
     pub fn duration_ms(&self) -> u64 {
         self.end.inner().wrapping_sub(self.start.inner()) as u64
     }
+
+    pub fn is_instant(&self) -> bool {
```
jdroenner, line 303: plase add a test (thats easy)

services/src/handlers/wfs.rs
```diff
@@ -185,25 +188,36 @@ async fn get_feature<C: Context>(
...
-        error::SpatialReferenceMissmatch {
-            found: request_spatial_ref,
-            expected: workflow_spatial_ref,
-        }
-    );
+    let request_spatial_ref: SpatialReference = request
+        .srs_name
+        .ok_or(error::Error::InvalidSpatialReference)?;
+
+    // perform reprojection if necessary
```
jdroenner, line 205: we should add a todo to avoid initializing twice. However, i see that it is not trivial using the current implementation of the projection.

services/src/handlers/wms.rs
```diff
@@ -150,33 +152,48 @@ async fn get_map<C: Context>(
...
-        workflow_spatial_ref == request_spatial_ref,
-        error::SpatialReferenceMissmatch {
-            found: request_spatial_ref,
-            expected: workflow_spatial_ref,
-        }
-    );
+    let request_spatial_ref: SpatialReference =
+        request.crs.ok_or(error::Error::InvalidSpatialReference)?;
+
+    // perform reprojection if necessary
```
jdroenner, line 168: we should add a todo to avoid initializing twice. However, i see that it is not trivial using the current implementation of the projection.

jdroenner (changes requested):
i think  we should add a few todos and add tests. The impleemntation is finee for now. why did you select shapefile instead of json or gpkg?

services/src/handlers/wms.rs
michaelmattig: b0e52e6416e7548ded6ed9f6ff1e5ec46d2c4909

services/src/handlers/wfs.rs
michaelmattig: b0e52e6416e7548ded6ed9f6ff1e5ec46d2c4909

datatypes/src/primitives/time_interval.rs
michaelmattig: 97b55797f554028b6691365d5d753fc2500c3ff6

michaelmattig:
> why did you select shapefile instead of json or gpkg?

I used the largest data set which just happened to be stored as shapefile

PR #202, opened 2021-04-19
rename number and decimal to float and int

ChristianBeilschmidt:


ChristianBeilschmidt:
[remark: I still find it confusing that our integer types are named Decimal when e.g. in Postgres Decimal is a variable precision data type with a decimal point.](https://trello.com/c/DH7t11Wf/24-remark-i-still-find-it-confusing-that-our-integer-types-are-named-decimal-when-eg-in-postgres-decimal-is-a-variable-precision-da)

PR #209, opened 2021-04-24
msg-integration

jdroenner:
- proj string is not the same as a display string
- add simple MSG projection handling
- avoid cloning of tile info in msg_source
- add buffer in msg source (needs a config entry and default)


jdroenner:
[Spatial_reference ennhancement  + gdal_source uses buffered + msg spec](https://trello.com/c/DXFHYIyP/114-spatialreference-ennhancement-gdalsource-uses-buffered-msg-spec)

services/test-data/dataset_defs/msg_fulldisk_ch2_netcdf.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "metaData": {
+    "GdalMetaDataRegular": {
+      "resultDescriptor": {
+        "dataType": "I16",
+        "spatialReference": "SR-ORG:81",
+        "measurement": "unitless"
+      },
+      "params": {
+        "filePath": "NETCDF:\"operators/test-data/raster/msg_netcdf/W_XX-EUMETSAT-Darmstadt,VIS+IR+HRV+IMAGERY,MSG4+SEVIRI_C_EUMG_%%%_START_TIME_%%%10.nc\":ch2",
```
ChristianBeilschmidt, line 16: You haven't checked-in the data.
Is it possible to have a test data set?

operators/src/source/gdal_source.rs
```diff
@@ -416,15 +431,8 @@ where
...
-            .then(async move |(tile, info)| {
-                Ok(RasterTile2D::new_with_tile_info(
-                    info.time,
-                    tile,
-                    Self::load_tile_data_async(info.params.clone(), tile).await?,
-                ))
-            })
-            .boxed()
+            .map(move |tile| Self::load_tile_async(info.params.clone(), tile, info.time))
+            .buffered(4) // TODO: find a good default and / or add to config.
```
ChristianBeilschmidt, line 435: Either remove buffering or leave a buffer of one with a comment.
But 4 is too arbitrary for me.

ChristianBeilschmidt (changes requested):

operators/src/source/gdal_source.rs
jdroenner: changed it to 1

services/test-data/dataset_defs/msg_fulldisk_ch2_netcdf.json
jdroenner: nope the data is not really public. the def is now here: https://github.com/geo-engine/dataset-defs/blob/main/msg_seviri_netcdf_ch2.json

ChristianBeilschmidt (approved):

PR #213, opened 2021-04-27
add MultiLineChart and temporal vector line plot

michaelmattig:


michaelmattig:
[Temporal Line Plot for Features](https://trello.com/c/uCPv5wU0/125-temporal-line-plot-for-features)

datatypes/src/plots/multi_line_plot.rs
```diff
@@ -0,0 +1,110 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn serialization() {
+        let chart = MultiLineChart::new(
+            vec![
+                ("S0".to_owned(), TimeInstance::from_millis_unchecked(0), 0.).into(),
+                ("S1".to_owned(), TimeInstance::from_millis_unchecked(0), 2.).into(),
+                ("S0".to_owned(), TimeInstance::from_millis_unchecked(0), 1.).into(),
```
ChristianBeilschmidt, line 98: Please use a different timestamp for the second data point.

datatypes/src/plots/multi_line_plot.rs
```diff
@@ -0,0 +1,110 @@
...
+            ],
+            Measurement::Unitless,
+        );
+        assert_eq!(
+            chart.to_vega_embeddable(false).unwrap(),
+            PlotData {
+                vega_string: r#"{"$schema":"https://vega.github.io/schema/vega-lite/v4.17.0.json","data":{"values":[{"x":"1970-01-01T00:00:00+00:00","y":0.0,"series":"S0"},{"x":"1970-01-01T00:00:00+00:00","y":2.0,"series":"S1"},{"x":"1970-01-01T00:00:00+00:00","y":1.0,"series":"S0"}]},"description":"Multi Line Chart","encoding":{"x":{"field":"x","title":"Time","type":"temporal"},"y":{"field":"y","title":"","type":"quantitative"},"color":{"field":"series","scale":{"scheme":"category20"}}},"mark":{"type":"line","line":true,"point":true}}"#.to_owned(),
```
ChristianBeilschmidt, line 105: In this case, it shows something like Friday, 3rd as x label(s).
Is it possible to show it more accurately, does it make sense to do so?
I guess it behaves okay when you have more points.

datatypes/src/plots/multi_line_plot.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn from(p: (String, TimeInstance, f64)) -> Self {
+        Self {
+            series: p.0,
+            time: p.1,
+            value: p.2,
+        }
+    }
+}
+
+pub struct MultiLineChart {
```
ChristianBeilschmidt, line 21: Please briefly describe what this plot is about.

operators/src/plot/temporal_vector_line_plot.rs
```diff
@@ -0,0 +1,360 @@
...
+            error::ColumnDoesNotExist {
+                column: self.params.value_column.clone()
+            }
+        );
+
+        let id_type = columns.get(&self.params.id_column).expect("checked");
+        let value_type = columns.get(&self.params.value_column).expect("checked");
+
+        ensure!(
+            id_type == &FeatureDataType::Text || id_type == &FeatureDataType::Int,
```
ChristianBeilschmidt, line 91: I guess categorical would be optimal, here, but I guess a TODO sufficies.

operators/src/plot/temporal_vector_line_plot.rs
```diff
@@ -0,0 +1,360 @@
...
+            error::InvalidFeatureDataType,
+        );
+
+        ensure!(
+            value_type != &FeatureDataType::Text,
```
ChristianBeilschmidt, line 96: Maybe we need a boolean indicator for numeric/non-numeric on our types.

operators/src/plot/temporal_vector_line_plot.rs
```diff
@@ -0,0 +1,360 @@
...
+                                    .ok_or(Error::FeatureDataValueMustNotBeNull)?,
+                                FeatureDataValue::Text(v) => v.clone(),
+                                FeatureDataValue::NullableText(v) => {
+                                    v.ok_or(Error::FeatureDataValueMustNotBeNull)?
+                                }
+                                FeatureDataValue::Category(v) => v.to_string(),
+                                FeatureDataValue::NullableCategory(v) => {
+                                    v.map(|v| v.to_string())
+                                        .ok_or(Error::FeatureDataValueMustNotBeNull)?
+                                }
```
ChristianBeilschmidt, line 182: What about a method `as_string()`?

I guess we will need it elsewhere and it improves readability.

operators/src/plot/temporal_vector_line_plot.rs
```diff
@@ -0,0 +1,360 @@
...
+                            };
+
+                            let value: f64 = match values.get_unchecked(i) {
+                                FeatureDataValue::Int(v) => v as f64,
+                                FeatureDataValue::NullableInt(v) => v.map_or(0.0, |v| v as f64), // TODO: NAN better default?
+                                FeatureDataValue::Float(v) => v,
+                                FeatureDataValue::NullableFloat(v) => v.unwrap_or(0.0),
+                                FeatureDataValue::Category(v) => f64::from(v),
+                                FeatureDataValue::NullableCategory(v) => v.map_or(0.0, f64::from),
+                                _ => return Err(Error::InvalidFeatureDataType),
```
ChristianBeilschmidt, line 192: What about a method `as_f64()` or `as_float()`?

I guess we will need it elsewhere and it improves readability.

operators/src/plot/temporal_vector_line_plot.rs
```diff
@@ -0,0 +1,360 @@
...
+        V: Into<TemporalValue>,
+    {
+        let len = self.values.len();
+
+        match self.values.entry(id) {
+            Occupied(mut entry) => entry.get_mut().push(value.into()),
+            Vacant(entry) => {
+                if len < LENGTH {
+                    entry.insert(vec![value.into()]);
+                }
```
ChristianBeilschmidt, line 245: So you just ignore it in the other case?
You could either return a Result or doc-comment the add function.

operators/src/plot/temporal_vector_line_plot.rs
```diff
@@ -0,0 +1,360 @@
...
+impl From<(TimeInterval, f64)> for TemporalValue {
+    fn from(value: (TimeInterval, f64)) -> Self {
+        Self {
+            time: value.0,
+            value: value.1,
+        }
+    }
+}
+
+#[derive(Default)]
```
ChristianBeilschmidt, line 228: I guess you should implement default yourself and instantiate the map `.with_capacity(LENGTH)`.

operators/src/plot/temporal_vector_line_plot.rs
```diff
@@ -0,0 +1,360 @@
...
+            .flat_map(|(id, values)| {
+                values.into_iter().map(move |value| DataPoint {
+                    series: id.clone(),
+                    time: value.time.start(),
+                    value: value.value,
+                })
+            })
+            .collect::<Vec<_>>();
+
+        data.sort_by(|a, b| match a.series.cmp(&b.series) {
```
ChristianBeilschmidt, line 263: Maybe [`sort_unstable`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_unstable)?

ChristianBeilschmidt (changes requested):

datatypes/src/plots/multi_line_plot.rs
```diff
@@ -0,0 +1,110 @@
...
+                    "title": x_axis_label,
+                    "type": "temporal"
+                },
+                "y": {
+                    "field": "y",
+                    "title": y_axis_label,
+                    "type": "quantitative"
+                },
+                "color": {
+                    "field": "series",
```
ChristianBeilschmidt, line 67: Maybe we should call the field id because we "select an id"

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: I don't think Category would be optimal. We really would need a way of flagging an attribute as an id (at source level, and operators would need to ensure that it stays an id or remove the flag). A categorical value does not necessarily uniquely identify a feature over time. Rather than a todo we would need to add a card.

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: yeah. Plus  it would really be nice to more easily process over int/float in a generic fashion

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: tbh I much rather would like a way of iterating over a column as a certain type. So turning a FeatureDataRef into an Iterator that automatically converts int to float or int to string etc. I created a card https://trello.com/c/MSNIEFxd

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: tbh I much rather would like a way of iterating over a column as a certain type. So turning a FeatureDataRef into an Iterator that automatically converts int to float or int to string etc. I created a card https://trello.com/c/MSNIEFxd

datatypes/src/plots/multi_line_plot.rs
michaelmattig: naming it `id` makes it appear to be the id of the DataPoint, but it's the id of feature or in plot terminology: the time series.

datatypes/src/plots/multi_line_plot.rs
michaelmattig: 745b8bf65d6fcc83aa594ac0759c9d472ed21214

datatypes/src/plots/multi_line_plot.rs
michaelmattig: 1de04c56901a6259b4f40ce4bac73c21a1fe4b38

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: dfece38832b7ccb1ba2e73f1de6b619818b93afc

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: f32607d32756a3e28ac482b3f342e27d738eb449

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: 315e96057f60a753ea553f16eb211d92d3c0ef78

datatypes/src/plots/multi_line_plot.rs
michaelmattig: I'd too like to have the axis label exactly on the time of the data points. However we can't assume that the data is always regular like in the raster-value-extraction where all points have values at the exact same times. So we would need to have a special type of plots for this I think.

operators/src/plot/temporal_vector_line_plot.rs
ChristianBeilschmidt: Hm, that's true, but we could offer an aggregate over a category.

operators/src/plot/temporal_vector_line_plot.rs
ChristianBeilschmidt: Could you extract a function then to improve the readability of the function and then put a link from the trello card to this code block such that we know what to replace once we created the functionality?

operators/src/plot/temporal_vector_line_plot.rs
ChristianBeilschmidt: Could you extract a function then to improve the readability of the function and then put a link from the trello card to this code block such that we know what to replace once we created the functionality?

operators/src/plot/temporal_vector_line_plot.rs
ChristianBeilschmidt: Could you create such a method?

operators/src/plot/temporal_vector_line_plot.rs
```diff
@@ -0,0 +1,369 @@
...
+                                FeatureDataValue::Category(v) => v.to_string(),
+                                FeatureDataValue::NullableCategory(v) => {
+                                    v.map(|v| v.to_string())
+                                        .ok_or(Error::FeatureDataValueMustNotBeNull)?
+                                }
+                            };
+
+                            let value: f64 = match values.get_unchecked(i) {
+                                FeatureDataValue::Int(v) => v as f64,
+                                FeatureDataValue::NullableInt(v) => v.map_or(0.0, |v| v as f64), // TODO: NAN better default?
```
ChristianBeilschmidt, line 187: You are right, I guess we should leave all null-values out?

Could you create a test for a plot where null-values where left out and one for duplicate values?

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: 626feaf41576cd7e04f859ea9701a2714356ad71

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: 817ad66a0e36eff863e2511991254ae7f6bea144

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: 92801e562d20b900275516583b8b8ab4b228334b

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: 92801e562d20b900275516583b8b8ab4b228334b

operators/src/plot/temporal_vector_line_plot.rs
michaelmattig: I handle nulls now: 92801e562d20b900275516583b8b8ab4b228334b
not sure about duplicate values. Should I check the whole vector if the exact timestamp with the same value already exists? Should I keep the vector sorted then?

ChristianBeilschmidt: I guess you don't have to do anything if vega handles it somehow. That is why you could create a test and then we see the outcome. My guess is that it's fine for now.

michaelmattig: I only test the generated vega code (which of course will contain duplicates) and not the resulting diagram. How do you suggest to do that?

ChristianBeilschmidt: I just paste the resulting definition to https://vega.github.io/editor/ and see if this works out.
If so, having the json definition as our test result is the same as saying "checked values with QGis".

michaelmattig: ok, I did that and the plots looks the same 7bf399c3fbc05ba4b2f755ea38a9004a4fb57280

PR #216, opened 2021-04-28
always box warp filters

michaelmattig:


michaelmattig:
[Make warp use less filters or change it.](https://trello.com/c/1FewFyF7/93-make-warp-use-less-filters-or-change-it)

PR #220, opened 2021-04-30
fix upload permissions in docker

jdroenner:


jdroenner:
[Docker-Nutzer braucht IO-Schreibrechte](https://trello.com/c/r07AbWoI/138-docker-nutzer-braucht-io-schreibrechte)

dev.Dockerfile
```diff
@@ -37,7 +37,15 @@ COPY docker/dev/Settings-dev.toml /app/Settings.toml
 COPY docker/dev/service.sh /etc/service/geoengine/run
 RUN chmod +x /etc/service/geoengine/run \
     && \
-    adduser --disabled-password --gecos "" geoengine
+    adduser --disabled-password --gecos "" geoengine 
+
+# Create upload folder and chown it to the geoengine user
+RUN mkdir -p /app/upload \
+    && \
+    chown -R geoengine:geoengine /app/upload \
```
michaelmattig, line 45: what happens if you mount a host folder/volume under this path (like we did it in the terranova demo for the data definitions)? Will it try to chown the host folder? Please test it.

dev.Dockerfile
```diff
@@ -37,7 +37,15 @@ COPY docker/dev/Settings-dev.toml /app/Settings.toml
...
+    && \
+    chown -R geoengine:geoengine /app/operators/test-data/vector/data
+# the last line is a work-around to allow the ogr-source to read gpkg files from the test-data folder. The gpkg driver needs RW rights.
```
michaelmattig, line 48: have you tried opening the data set as read only? Or, alternatively, is it possible to modify the gpkg s.t. it doesn't need write access?

michaelmattig (changes requested):
two comments

dev.Dockerfile
jdroenner: i don' think that works

dev.Dockerfile
jdroenner: this is run at image creation. it will not change the folder you mount inside. however, it will create new folder inside the folder with the uid of the user in the container. so uid mapping would be a good idea.

PR #223, opened 2021-05-03
resolution in WFS handler

ChristianBeilschmidt:


ChristianBeilschmidt:
[QueryResolution an WFS übergeben als Vendor-Parameter](https://trello.com/c/FhgOKL1M/152-queryresolution-an-wfs-%C3%BCbergeben-als-vendor-parameter)

services/src/handlers/wfs.rs
```diff
@@ -228,7 +228,9 @@ async fn get_feature<C: Context>(
             let time = TimeInstance::from(chrono::offset::Utc::now());
             TimeInterval::new_unchecked(time, time)
         }),
-        spatial_resolution: SpatialResolution::zero_point_one(),
+        spatial_resolution: request
+            .query_resolution
+            .unwrap_or_else(SpatialResolution::zero_point_one),
```
michaelmattig, line 233: is this is meaningful default value? Should id maybe be dependant of the projection? Maybe put a TODO here to check this

michaelmattig (changes requested):
A question

services/src/handlers/wfs.rs
ChristianBeilschmidt: f820d8f

PR #228, opened 2021-05-03
fix float/int confusion in ogr source

ChristianBeilschmidt:


ChristianBeilschmidt:
[Upload-Bug bei fields.gpkg](https://trello.com/c/v2v74SfX/160-upload-bug-bei-fieldsgpkg)

PR #229, opened 2021-05-04
clip bbox to valid area of projection

michaelmattig:


michaelmattig:
[Fix suggest_pixel_size to respect projection bounds](https://trello.com/c/wL53inOL/161-fix-suggestpixelsize-to-respect-projection-bounds)

datatypes/src/operations/reproject.rs
```diff
@@ -214,14 +226,60 @@ where
...
+
+impl<P> ReprojectClipped<P> for BoundingBox2D
+where
+    P: CoordinateProjection,
+{
+    type Out = BoundingBox2D;
+    fn reproject_clipped(&self, projector: &P) -> Result<BoundingBox2D> {
+        const POINTS_PER_LINE: i32 = 7;
+
+        let use_bbox = projector.target_srs().area_of_use()?;
```
jdroenner, line 249: i guess we need to transform the usebox to the source srs before using it? Its lat/lon right?

datatypes/src/operations/reproject.rs
```diff
@@ -214,14 +226,60 @@ where
...
+
         let proj_outline_coordinates: Vec<Coordinate2D> =
             project_coordinates_fail_tolerant(&outline_coordinates, projector)
                 .into_iter()
                 .flatten()
                 .collect();
 
-        // TODO: check min coords or use grid? e.g. ensure!(proj_cs.len() ). fail only if there are not enough coordinates after reprojection. This could also require us to use a grid instead of bounds. The good news is that the Grid can already provide this.
-        Ok(MultiPoint::new_unchecked(proj_outline_coordinates).spatial_bounds())
+        let out = MultiPoint::new(proj_outline_coordinates)?.spatial_bounds();
```
jdroenner, line 275: while this removes the todo it will also work with only one point

datatypes/src/spatial_reference.rs
```diff
@@ -69,6 +70,19 @@ impl SpatialReference {
             }
         }
     }
+
+    pub fn area_of_use(self) -> Result<BoundingBox2D> {
```
jdroenner, line 74: maybe add a note that this will be in lat/lon. And maybe add a method for a projected area of use?

jdroenner (changes requested):
please check what will happen if the source projection is not epsg:4326. The query rewrite will transform in both directions and i think that the proposed changes only work if source is epsg:4326

datatypes/src/operations/reproject.rs
michaelmattig: 7c3dcc44c62f4ead9564512f5d7b4f05479b4d43

datatypes/src/spatial_reference.rs
michaelmattig: 7c3dcc44c62f4ead9564512f5d7b4f05479b4d43

datatypes/src/operations/reproject.rs
michaelmattig: but I check in the next line that the bbox has a real size and is not a point

jdroenner: sorry did not see that -_-

datatypes/src/operations/reproject.rs
```diff
@@ -214,14 +226,71 @@ where
...
+        let target_area_of_use = projector.target_srs().area_of_use()?;
+        let area_of_use = source_area_of_use
+            .intersection(&target_area_of_use)
+            .ok_or(Error::BboxesDoNotIntersect)?;
+        let area_of_use = area_of_use.reproject(&area_of_use_projector)?;
+        let clipped_bbox = self
+            .intersection(&area_of_use)
+            .ok_or(Error::BboxesDoNotIntersect)?;
+
+        // project points on the bbox
```
jdroenner, line 264: you could just call project on the clipped bbox. It would also be nice if you could move the empty check to the non clipped one.

datatypes/src/error.rs
```diff
@@ -212,6 +211,12 @@ pub enum Error {
     NoMatchingVectorDataTypeForOgrGeometryType,
 
     NoMatchingFeatureDataTypeForOgrFieldType,
+
+    InvalidProjDefinition,
+    NoAreaOfUseDefined,
```
jdroenner, line 216: why not add the projection and display it?

datatypes/src/error.rs
```diff
@@ -212,6 +211,12 @@ pub enum Error {
...
+    BboxesDoNotIntersect,
+
+    CouldNotReprojectBbox,
```
jdroenner, line 219: why not add the box and the projection?

datatypes/src/error.rs
```diff
@@ -212,6 +211,12 @@ pub enum Error {
     NoMatchingVectorDataTypeForOgrGeometryType,
 
     NoMatchingFeatureDataTypeForOgrFieldType,
+
+    InvalidProjDefinition,
```
jdroenner, line 215: why not add the projection?

operators/src/processing/reprojection.rs
```diff
@@ -160,16 +161,24 @@ where
...
-    let p_bbox = query.bbox.reproject(&projector)?;
-    let p_spatial_resolution =
-        suggest_pixel_size_from_diag_cross(query.bbox, query.spatial_resolution, &projector)?;
+    let projector_source_target = CoordinateProjector::from_known_srs(source, target)?;
+    let projector_target_source = CoordinateProjector::from_known_srs(target, source)?;
+
+    let p_bbox = query.bbox.reproject_clipped(&projector_target_source)?;
+    let s_bbox = p_bbox.reproject(&projector_source_target)?;
+
+    let p_spatial_resolution = suggest_pixel_size_from_diag_cross(
```
jdroenner, line 177: you have both bounding boxes now. so we can avoid the projection in suggest_pixel_size_from_diag_cross... maybe add a non projecting version o

datatypes/src/operations/reproject.rs
michaelmattig: we discussed that there are different error handlings in the two methods. The empty check is not needed in the non-clipped one

operators/src/processing/reprojection.rs
michaelmattig: ca3ae8b132458a61c13a109d75abd0d2569b3883

datatypes/src/error.rs
michaelmattig: 3ae7a825079e9e6b720dd7a8badb99d6b2ac6e0f

datatypes/src/error.rs
michaelmattig: 3ae7a825079e9e6b720dd7a8badb99d6b2ac6e0f

datatypes/src/error.rs
michaelmattig: 3ae7a825079e9e6b720dd7a8badb99d6b2ac6e0f

PR #231, opened 2021-05-06
set spatial filter in ogr source

jdroenner:


jdroenner:
[set spatial filter in ogr source](https://trello.com/c/OAuDZFk0/167-set-spatial-filter-in-ogr-source)

operators/src/source/ogr_source.rs
```diff
@@ -423,6 +426,15 @@ where
         // TODO: add OGR time filter if forced
         let dataset = Self::open_gdal_dataset(&dataset_information)?;
         let mut layer = dataset.layer_by_name(&dataset_information.layer_name)?;
+        let use_ogr_spatial_filter = dataset_information.force_ogr_spatial_filter
+            || layer.has_capability(gdal::vector::LayerCaps::OLCFastSpatialFilter);
```
ChristianBeilschmidt, line 430: This way we use it automatically if OGR says it is possible.
So `force_ogr_spatial_filter` only has an effect if it is true and OGR says it cannot use it.
Does this make sense? I would suggest removing the parameter (in the operator spec) and just use it when OGR says it can use it.

operators/src/source/ogr_source.rs
```diff
@@ -423,6 +426,15 @@ where
...
+
+        if use_ogr_spatial_filter {
+            // rectangular geometry from West, South, East and North values.
+            let filter_geometry = query_rectangle.bbox.try_into()?;
+            // TODO: log uses spatial filter
+            layer.set_spatial_filter(&filter_geometry);
```
ChristianBeilschmidt, line 436: Why does this work? `filtere_geometry` is dropped after this if-block… so is `set_spatial_filter` copying the geometry?

ChristianBeilschmidt (changes requested):

operators/src/source/ogr_source.rs
jdroenner: yes thats the way i think it is usefull. we can override the driver bbut the default is that we filter it if the driver can not provide it (fast).

operators/src/source/ogr_source.rs
jdroenner: it works... maybe the compiler is smart enough. maybe it is a bug

ChristianBeilschmidt: use https://gdal.org/doxygen/ogr__api_8h.html#a5cba569e0779a02a95327f041d9f7a13?

ChristianBeilschmidt: Furthermore, we should add a TODO, that the OGR-filter may be inaccurately allowing more features that should be returned in a "strict" fashion, whatever this means. cf . the docu

PR #237, opened 2021-05-12
fix histogram for empty inputs & no data,

ChristianBeilschmidt:


ChristianBeilschmidt:
[Histogram-Operator on no data raster with range type data: Operator: DataTypeError: Plot exception: Histograms max value must be larger than its min value](https://trello.com/c/IB4pFpdL/155-histogram-operator-on-no-data-raster-with-range-type-data-operator-datatypeerror-plot-exception-histograms-max-value-must-be-lar)

operators/src/plot/histogram.rs
```diff
@@ -313,6 +323,19 @@ impl HistogramRasterQueryProcessor {
 
         Ok(chart)
     }
+
+    fn empty_histogram(
+        &self,
+    ) -> Result<<HistogramRasterQueryProcessor as PlotQueryProcessor>::OutputFormat> {
+        let histogram =
+            geoengine_datatypes::plots::Histogram::builder(1, 0., 0., self.measurement.clone())
```
jdroenner, line 331: Thats not working with the ui: 
"Operator: DataTypeError: Plot exception: Histograms max value must be larger than its min value"

operators/src/plot/histogram.rs
```diff
@@ -384,6 +407,19 @@ impl HistogramVectorQueryProcessor {
...
```
jdroenner, line 415: same here

operators/src/plot/histogram.rs
```diff
@@ -393,6 +429,12 @@ struct HistogramMetadata {
     pub max: f64,
 }
 
+impl HistogramMetadata {
+    fn has_invalid_parameters(&self) -> bool {
+        self.number_of_buckets == 0 || self.min > self.max
```
jdroenner, line 434: you need to check for min >= max i think

jdroenner (changes requested):
sadly this does not work for the ui

operators/src/plot/histogram.rs
ChristianBeilschmidt: 359ff6d

operators/src/plot/histogram.rs
ChristianBeilschmidt: 359ff6d

operators/src/plot/histogram.rs
ChristianBeilschmidt: 359ff6d

PR #238, opened 2021-05-12
operator for window aggregation of rasters

michaelmattig:


michaelmattig:
[Time-Aggregation (Min, Max, Mean) TemporalWindowAggregation](https://trello.com/c/1VqOp6IU/186-time-aggregation-min-max-mean-temporalwindowaggregation)

jdroenner (changes requested):
also please add the adapter to RasterStreamExt in mod.rs

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,549 @@
...
+                    step: self.window,
+                };
+
+                let s = RasterOverlapAdapter::<'a, P, _, _>::new(
+                    &self.source,
+                    query,
+                    self.tiling_specification,
+                    ctx,
+                    spec,
+                );
```
ChristianBeilschmidt, line 157: Could you add the overlap adapter as a method to our [`RasterStreamExt`](https://github.com/geo-engine/geoengine/blob/master/operators/src/adapters/mod.rs)? It wasn't available when this struct was written, but we should add it to improve our stream API. And it will fit nicely here when you can call it directly on `self.source`.

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,549 @@
...
+        }
+        .boxed();
+
+        let exe_ctx = MockExecutionContext {
+            tiling_specification: TilingSpecification::new((0., 0.).into(), [3, 2].into()),
+            ..Default::default()
+        };
+        let query_rect = QueryRectangle {
+            bbox: BoundingBox2D::new_unchecked((0., 0.).into(), (4., 3.).into()),
+            time_interval: TimeInterval::new_unchecked(0, 20),
```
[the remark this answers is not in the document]
michaelmattig: 267c68d08aa5f9b4c9eee5c96066f7b526f9a48f

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,549 @@
...
+                    tile_size_in_pixels: [3, 2].into(),
+                    global_geo_transform: Default::default(),
+                },
+                Grid2D::new([3, 2].into(), vec![6, 5, 4, 3, 2, 1], no_data_value).unwrap(),
+            )
+        );
+    }
+
+    #[tokio::test]
+    async fn test_max() {
```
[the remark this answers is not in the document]
michaelmattig: 267c68d08aa5f9b4c9eee5c96066f7b526f9a48f

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,549 @@
...
+
+                Ok(s.boxed())
+            }
+            AggregationType::Mean => Err(error::Error::NotYetImplemented), // TODO: implement
+        }
+    }
+}
+
+// TODO: better name
+pub trait CompareFunction {
```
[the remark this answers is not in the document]
michaelmattig: id doesn't only compare but compute the new accumulator value for the pixel. I renamed it bc077e96e2f6a624b0941d6f3dc7c0962d760fa9

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,549 @@
...
+    tiling_specification: TilingSpecification,
+    no_data_value: Option<P>,
+}
+
+impl<Q, P> TemporalRasterAggregationProcessor<Q, P>
+where
+    Q: RasterQueryProcessor<RasterType = P>,
+    P: Pixel,
+{
+    fn new(
```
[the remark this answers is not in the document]
michaelmattig: bc077e96e2f6a624b0941d6f3dc7c0962d760fa9

operators/src/processing/temporal_raster_aggregation.rs
michaelmattig: Please give me a hint how to do that

ChristianBeilschmidt: You could do it similar to [time_multi_fold](https://github.com/geo-engine/geoengine/blob/9cde48cb7c36249e6882653e4c1149cd71dfb293/operators/src/adapters/mod.rs#L36). You know how to call it from the implementation above, so it could work.

michaelmattig: TemporalRasterAggregationOperator needs a  query processor as input and not a stream so I don't see how I could implement this.

operators/src/engine/query_processor.rs
```diff
@@ -222,6 +222,63 @@ impl TypedRasterQueryProcessor {
     }
 }
 
+impl From<Box<dyn RasterQueryProcessor<RasterType = u8>>> for TypedRasterQueryProcessor {
```
jdroenner, line 225: where is this used?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,1074 @@
...
+use snafu::ensure;
+use typetag;
+
+use crate::engine::{ExecutionContext, Operator, RasterOperator, SingleRasterSource};
+
+#[derive(Debug, Clone, Copy, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TemporalRasterAggregationParameters {
+    aggregation_type: AggregationType,
+    window: TimeStep,
```
jdroenner, line 26: window_size?

jdroenner (changes requested):
only two small things

datatypes/src/primitives/time_step.rs
```diff
@@ -32,7 +32,7 @@ pub enum TimeGranularity {
 #[cfg_attr(feature = "postgres", derive(ToSql, FromSql))]
 pub struct TimeStep {
     pub granularity: TimeGranularity,
-    pub step: u32,
+    pub step: u32, // TODO: ensure on deserialization it is > 0
```
ChristianBeilschmidt, line 35: What about [NonZeroU32](https://doc.rust-lang.org/std/num/struct.NonZeroU32.html)?

datatypes/src/raster/tiling.rs
```diff
@@ -205,6 +205,36 @@ impl TileInformation {
...
+
+        let overlap_y = crate::util::ranges::value_in_range(
+            this_bbox.lower_left().y,
+            bbox.lower_left().y,
+            bbox.upper_right().y,
+        ) || crate::util::ranges::value_in_range(
+            bbox.lower_left().y,
+            this_bbox.lower_left().y,
+            this_bbox.upper_right().y,
+        );
```
ChristianBeilschmidt, line 234: Why not use ranges from the stdlib and [contains](https://doc.rust-lang.org/std/ops/struct.Range.html#method.contains) instead of a custom method?

operators/src/adapters/raster_time.rs
```diff
@@ -188,15 +191,15 @@ mod tests {
                 data: vec![
                     RasterTile2D {
                         time: TimeInterval::new_unchecked(0, 5),
-                        tile_position: [0, 0].into(),
+                        tile_position: [-1, 0].into(),
```
ChristianBeilschmidt, line 194: What was the rationale behind changing the method and the test? Was the test previously wrong?

operators/src/processing/expression.rs
```diff
@@ -542,7 +542,7 @@ mod tests {
         let raster_tile = RasterTile2D::new_with_tile_info(
             TimeInterval::default(),
             TileInformation {
-                global_tile_position: [0, 0].into(),
+                global_tile_position: [-1, 0].into(),
```
ChristianBeilschmidt, line 545: Why was it necessary to change this here and why do the tests still run through?

operators/src/processing/raster_vector_join/points_aggregated.rs
```diff
@@ -228,7 +234,7 @@ mod tests {
             "foo",
             AggregationMethod::First,
             QueryRectangle {
-                bbox: BoundingBox2D::new((0.0, 0.0).into(), (3.0, 2.0).into()).unwrap(),
+                bbox: BoundingBox2D::new((0.0, -3.0).into(), (2.0, 0.).into()).unwrap(),
```
ChristianBeilschmidt, line 237: Why was it necessary to change this?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,1074 @@
...
+                        query,
+                        self.tiling_specification,
+                        ctx,
+                        spec,
+                    );
+
+                    Ok(s.boxed())
+                }
+            }
+            AggregationType::Mean => Err(error::Error::NotYetImplemented), // TODO: implement
```
ChristianBeilschmidt, line 225: I guess this also holds for the median, so do we add both as NotYetImplemented or just omit them for now?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,1074 @@
...
+        query: crate::engine::QueryRectangle,
+        ctx: &'a dyn crate::engine::QueryContext,
+    ) -> Result<futures::stream::BoxStream<'a, Result<RasterTile2D<Self::RasterType>>>> {
+        match self.aggregation_type {
+            AggregationType::Min => {
+                if self.ignore_no_data {
+                    let spec = TemporalRasterAggregationSubQuery {
+                        fold_fn: no_data_ignoring_fold_future::<P, MinIgnoreNoDataAccFunction>,
+                        no_data_value: self.no_data_value,
+                        initial_value: P::max_value(),
```
ChristianBeilschmidt, line 156: I guess you should add tests with empty raster streams. Do we get a very large minimum there?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,1074 @@
...
+}
+pub trait NoDataIgnoringAccFunction {
+    /// produce new accumulator value from current state and new value, ignoring no data values
+    fn acc_ignore_no_data<T: Pixel>(no_data: Option<T>, acc: T, value: T) -> T;
+}
+
+struct MinAccFunction {}
+
+impl AccFunction for MinAccFunction {
+    fn acc<T: Pixel>(no_data: Option<T>, acc: T, value: T) -> T {
```
ChristianBeilschmidt, line 242: Shouldn't be the no_data values rather in the struct than in every call?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -0,0 +1,1074 @@
...
+    tile: RasterTile2D<T>,
+) -> impl Future<Output = Result<(RasterTile2D<T>, ())>>
+where
+    T: Pixel,
+    C: AccFunction,
+{
+    tokio::task::spawn_blocking(|| fold_fn::<T, C>(accu, tile)).then(async move |x| match x {
+        Ok(r) => Ok(r),
+        Err(e) => Err(e.into()),
+    })
```
ChristianBeilschmidt, line 377: Why is it necessary to spawn tasks? Couldn't you - as you once told me - just say that the method is async in order to get a future?

ChristianBeilschmidt (changes requested):
@jdroenner asked me to review this as well, so here we are.

operators/src/engine/query_processor.rs
michaelmattig: here: https://github.com/geo-engine/geoengine/blob/bc077e96e2f6a624b0941d6f3dc7c0962d760fa9/operators/src/processing/temporal_raster_aggregation.rs#L94

operators/src/processing/temporal_raster_aggregation.rs
michaelmattig: window is a TimeStep which is Granularity + step_size

datatypes/src/primitives/time_step.rs
michaelmattig: good idea but needs to be a separate task. The TimeStep is used in a way where the step can be 0. 
* There are tests that assure it works with 0
* the TimeStepIter multiples the step with a running index that starts at zero
* the num_step function uses steps with 0 value

there are many changes across the project necessary which are beyond the scope of this PR

datatypes/src/raster/tiling.rs
michaelmattig: the function already existed and I just used it. We can have an improvement task which replaces its uses with the stdlib Range

operators/src/adapters/raster_time.rs
michaelmattig: as discussed the Mockrastersources previously didn't filter spatially and now it does. It then became apparent that the test data are wrong (tile index doesn't correspond to coordinates) which went unnoticed because the mockrastersource always returned the same tile ignoring the query bbox.

operators/src/processing/expression.rs
michaelmattig: cf other comment

operators/src/processing/raster_vector_join/points_aggregated.rs
michaelmattig: cf other comment

operators/src/processing/temporal_raster_aggregation.rs
michaelmattig: 087b477bfeb7dff173a5258175d9c442682cbd8c

operators/src/processing/temporal_raster_aggregation.rs
michaelmattig: there are no empty raster stream, rather a stream with no data in a query rectangle would get a no data tile which would lead to no data output or, in the case of ignore no data, max or min values. but then it's the users fault for ignoring the no data.

operators/src/processing/temporal_raster_aggregation.rs
michaelmattig: it is a static method so we can't access data in the struct

operators/src/processing/temporal_raster_aggregation.rs
michaelmattig: 0d2561a7b8cc2e6b8b3690a7ce2b9d6cbfa8ae24

datatypes/src/primitives/time_step.rs
jdroenner: add a task / idea?

datatypes/src/raster/tiling.rs
jdroenner: add a task / idea

ChristianBeilschmidt: or just do it here and leave the function where it is for now

operators/src/processing/temporal_raster_aggregation.rs
jdroenner: i do not think that this is correct. We discussed this and  need to find a solution.

operators/src/processing/temporal_raster_aggregation.rs
ChristianBeilschmidt: Okay, and does it work with just one empty tile?

datatypes/src/primitives/time_step.rs
michaelmattig: https://trello.com/c/3a9SrZnQ

datatypes/src/raster/tiling.rs
michaelmattig: https://trello.com/c/T4rtdrVe

operators/src/processing/temporal_raster_aggregation.rs
michaelmattig: f8880760839c283a01ce233a024c663ed914dd0e

operators/src/processing/temporal_raster_aggregation.rs
michaelmattig: e1f6564563051a9654ea86b41c8379f18c66e178

ChristianBeilschmidt (approved):

PR #241, opened 2021-05-17
projection should start with empty tile

jdroenner:


jdroenner:
[projection shuld start with empty tile](https://trello.com/c/OAaqMc2G/192-projection-shuld-start-with-empty-tile)

PR #244, opened 2021-05-18
set stable ids for datasets

michaelmattig:


michaelmattig:
[Stabile Ids für Datasets](https://trello.com/c/62j1m3r9/136-stabile-ids-f%C3%BCr-datasets)

PR #249, opened 2021-05-20
stac_s2

jdroenner:
This adds a dataset provider which allows to use S2 tiles listed in a STAC result.


jdroenner:
[Gdal Dataset Loading Infos/Provider für S3-Buckets](https://trello.com/c/rY46NXzq/181-gdal-dataset-loading-infos-provider-f%C3%BCr-s3-buckets)

PR #253, opened 2021-05-25
add first, last temporal raster agg

jdroenner:


jdroenner:
[First / Laster Temporal Raster Aggr.](https://trello.com/c/pP7A8pXg/199-first-laster-temporal-raster-aggr)

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -141,85 +136,128 @@ where
 {
     type RasterType = P;
 
+    #[allow(clippy::too_many_lines)]
```
michaelmattig, line 139: can you maybe extract the creation of the adapter into a method? would make it more readable also

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -141,85 +136,128 @@ where
...
+                    spec,
+                );
+
+                Ok(s.boxed())
+            }
+            AggregationType::FirstValid => {
+                let spec = TemporalRasterAggregationSubQuery {
+                    fold_fn: no_data_ignoring_fold_future::<P, FirstValidAccFunction>,
+                    no_data_value: self.no_data_value,
+                    initial_value: self.no_data_value.unwrap_or_else(P::zero), // TODO: what to do if there is no nodata?
```
michaelmattig, line 230: I think it would be better to just return `Err` here.

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -141,85 +136,128 @@ where
...
-                        spec,
-                    );
-
-                    Ok(s.boxed())
-                }
+            AggregationType::LastValid => {
+                let spec = TemporalRasterAggregationSubQuery {
+                    fold_fn: no_data_ignoring_fold_future::<P, LastValidAccFunction>,
+                    no_data_value: self.no_data_value,
+                    initial_value: self.no_data_value.unwrap_or_else(P::zero), // TODO: what to do if there is no nodata?
```
michaelmattig, line 248: I think it would be better to just return Err here

michaelmattig (changes requested):
some comments

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -145,119 +158,71 @@ where
...
-                    self.tiling_specification,
-                    ctx,
-                    spec,
-                );
-
-                Ok(s.boxed())
+                let no_data_value =
+                    self.no_data_value
+                        .ok_or(error::Error::OperationRequiresNoDataValue {
+                            operation_name: "TemporalRasterAggregation::FirstValid".to_string(),
```
michaelmattig, line 212: wrong operation_name. Maybe it would make more sense to have a nested error here. Maybe a TemporalRasterAggreation error with a nested  MissingNoData error?

michaelmattig (changes requested):
- please fix checks
- is the too many lines function now short enough for clippy?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -13,26 +14,33 @@ use geoengine_datatypes::{
...
+pub enum Aggregation {
+    #[serde(rename_all = "camelCase")]
+    Min {
+        ignore_no_data: bool,
+    },
+    #[serde(rename_all = "camelCase")]
+    Max {
+        ignore_no_data: bool,
+    },
+    First,
```
michaelmattig, line 42: Chris said ignore_no_data makes sense for First and Last after all:
- First
  - ignore_no_data = true: takes the first valid value, if there is any
  - ignore_no_data = false: takes the first value, even if it is no_data

there are applications for both cases. What do you think?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -141,85 +160,113 @@ where
...
+                .into_raster_overlap_adapter(&self.source, query, ctx, self.tiling_specification)
+                .boxed()),
+            Aggregation::Max {
+                ignore_no_data: false,
+            } => Ok(self
+                .create_subquery(fold_future::<P, MaxAccFunction>, P::min_value())
+                .into_raster_overlap_adapter(&self.source, query, ctx, self.tiling_specification)
+                .boxed()),
+            Aggregation::First {
+                ignore_no_data: true,
```
michaelmattig, line 201: shouldn't it be the other way around? Ignoring no data mean just taking the pixels from the first tile no matter if they are no data (ignoring no data).

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -141,85 +160,113 @@ where
...
+                    .boxed())
+            }
+            Aggregation::First {
+                ignore_no_data: false,
+            } => {
+                let no_data_value = self
+                    .no_data_value
+                    .ok_or(error::Error::TemporalRasterAggregationFirstValidRequiresNoData)?;
+                Ok(self
+                    .create_subquery(first_tile_fold_future::<P>, no_data_value)
```
michaelmattig, line 226: can we avoid processing all tiles of the timeseries when all we want is the first one?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -391,6 +478,93 @@ where
...
+    T: Pixel,
+{
+    tokio::task::spawn_blocking(|| last_tile_fold_fn(accu, tile)).then(async move |x| match x {
+        Ok(r) => Ok(r),
+        Err(e) => Err(e.into()),
+    })
+}
+
+#[derive(Debug, Clone)]
+pub struct TemporalRasterAggregationTileAccu<T> {
```
michaelmattig, line 545: why do you need this? can't you just use a `Option<RasterTile2D>` as accumulator and start with None?

operators/src/processing/temporal_raster_aggregation.rs
```diff
@@ -1054,6 +1238,188 @@ mod tests {
...
+                    no_data_value: no_data_value.map(AsPrimitive::as_),
+                },
+            },
+        }
+        .boxed();
+
+        let agg = TemporalRasterAggregation {
+            params: TemporalRasterAggregationParameters {
+                aggregation: Aggregation::First {
+                    ignore_no_data: true,
```
michaelmattig, line 1261: please add tests with `ignore_no_data` false

michaelmattig (changes requested):

operators/src/processing/temporal_raster_aggregation.rs
jdroenner: i don't think that is possible here. you need to cancel the subquery for that.

operators/src/processing/temporal_raster_aggregation.rs
jdroenner: well i could implement the trait for option i guess.

operators/src/processing/temporal_raster_aggregation.rs
jdroenner: ok, you designed the ignore_no_data semantics

michaelmattig: I am still unsure. Take this example:
- NDVI data set starts at 1.1.2014
- set query time 1.12.2013: result is all no_data
- add temporal aggregation operator, window 2 month, type: First, ignore_no_data: false
- current result: we get the image from 1.1.2014

however, ignore_no_data ist only another way of saying: take only valid data. If it is false, then we also take no_data values into the aggregation. Thus, in this case we should take the first raster values, regardless if there are no_data. Right?

That's actually also what I wanted in my comment above, but it seems like it was right before.

PR #259, opened 2021-06-07
async query method in QueryProcessor

michaelmattig:


michaelmattig:
[Query-Methode async machen](https://trello.com/c/oZgwrF9D/210-query-methode-async-machen)

PR #264, opened 2021-06-10
Sentinel 2 S3 provider

michaelmattig:


michaelmattig:
[Externe Datenquelle für die S2 S3 Tiles bauen mit earth search.](https://trello.com/c/4xuNOFJn/201-externe-datenquelle-f%C3%BCr-die-s2-s3-tiles-bauen-mit-earth-search)

operators/src/error.rs
```diff
@@ -193,6 +193,12 @@ pub enum Error {
 
     TemporalRasterAggregationLastValidRequiresNoData,
     TemporalRasterAggregationFirstValidRequiresNoData,
+
+    LoadingInfo {
+        reason: String, // TODO: find cleaner way to make errors in services crate visible
```
jdroenner, line 198: maybe just a box error?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+    async fn initialize(
+        self: Box<Self>,
+    ) -> crate::error::Result<Box<dyn crate::datasets::listing::DatasetProvider>> {
+        Ok(Box::new(SentinelS2L2aCogsDataProvider::new(
+            self.id,
+            self.api_url,
+        )))
+    }
+
+    fn type_name(&self) -> String {
```
jdroenner, line 55: whats this? / why is this needed? why is it not static?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+
+#[derive(Debug, Clone)]
+pub struct SentinelDataset {
+    band: Band,
+    zone: Zone,
+    listing: DatasetListing,
+}
+
+pub struct SentinelS2L2aCogsDataProvider {
+    #[allow(dead_code)]
```
jdroenner, line 111: why is this dead?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+            id,
+            api_url,
+            datasets: Self::create_datasets(&id, &meta_data),
+            meta_data,
+        }
+    }
+
+    fn load_metadata() -> SentinelMetaData {
+        // TODO: fetch dataset metadata from config or remote
+        // TODO: is there a no data value?
```
jdroenner, line 132: yes and you also set it :)

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+        SentinelMetaData {
```
jdroenner, line 133: this will never scale there are like 12 bands and 80 zones.

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+                })
+            })
+            .collect()
+    }
+}
+
+#[async_trait]
+impl DatasetProvider for SentinelS2L2aCogsDataProvider {
+    async fn list(&self, _options: Validated<DatasetListOptions>) -> Result<Vec<DatasetListing>> {
+        // TODO: options
```
jdroenner, line 194: what kind of options?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+    ) -> Result<Vec<StacFeature>> {
+        let mut features = vec![];
+
+        let mut collection = self.load_collection(params, 1).await?;
+        features.append(&mut collection.features);
+
+        let num_pages =
+            (collection.context.matched as f64 / collection.context.limit as f64).ceil() as u32;
+
+        for page in 2..=num_pages {
```
jdroenner, line 344: what happens if num_pages is 1?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+            time.start()
+                .as_utc_date_time()
+                .ok_or(geoengine_operators::error::Error::DataType {
+                    source: geoengine_datatypes::error::Error::NoDateTimeValid {
+                        time_instance: time.start(),
+                    },
+                })?;
+
+        // shift start by 1 day to ensure getting the most recent data for start time
+        let t_start = t_start - Duration::days(1);
```
jdroenner, line 387: i guess we can also move it by -1 minutes. this will be enough to include a tile if it started before.

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+        debug!("number of features in current zone: {}", num_features);
+        for i in 0..num_features {
+            let feature = &features[i];
+
+            let start = feature.properties.datetime;
+            // feature is valid until next feature starts
+            let end = if i < num_features - 1 {
+                features[i + 1].properties.datetime
+            } else {
+                start + Duration::hours(12) // TODO: determine correct validity for last tile
```
jdroenner, line 241: maybe just add 1 minute. this should be fine..

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+
+    async fn result_descriptor(&self) -> geoengine_operators::util::Result<RasterResultDescriptor> {
+        Ok(RasterResultDescriptor {
+            data_type: self.band.data_type,
+            spatial_reference: SpatialReference::new(
+                SpatialReferenceAuthority::Epsg,
+                self.zone.epsg,
+            )
+            .into(),
+            measurement: Measurement::Unitless,
```
jdroenner, line 424: add a todo to add this later

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+            no_data_value: self.band.no_data_value,
+        })
+    }
+
+    fn box_clone(&self) -> Box<dyn MetaData<GdalLoadingInfo, RasterResultDescriptor>> {
```
jdroenner, line 429: can we default impl this?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+        Ok(Box::new(SentinelS2L2aCogsMetaData {
+            api_url: self.api_url.clone(),
+            zone: dataset.zone.clone(),
+            band: dataset.band.clone(),
+        }))
+    }
+}
+
+#[async_trait]
+impl MetaDataProvider<MockDatasetDataSourceLoadingInfo, VectorResultDescriptor>
```
jdroenner, line 457: do we really have to implement all kinds if they are not really implemented? There should be a task to solve this

jdroenner (commented):
i think this works for a first version. please explain some design choices

operators/src/error.rs
michaelmattig: the error occurs in the services crate so the operators Error cannot know the error and thus also a `Box<geoengine_services::Error>` won't work

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: it is the type of the provider, while the name ist the concrete instance. "WCS" vs. "DWD Geoserver".
It cannot be static because then it wouldn't be deserializable with type_tag

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: it's dead right now because I create all the MetaData on `new` but in practice it should be retrieved from STAC and in order to generate the external dataset id the provider needs to know its own provider id, which is part of the dataset id. I can remove it for now if you want

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -0,0 +1,640 @@
...
+    band: Band,
+    zone: Zone,
+    listing: DatasetListing,
+}
+
+pub struct SentinelS2L2aCogsDataProvider {
+    #[allow(dead_code)]
+    id: DatasetProviderId,
+    api_url: String,
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
michaelmattig: cf. above

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: yeah, we would need hierarchical datasets here I guess

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: the listing options. sorting, filtering etc.

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: then no further pages have to be downloaded. the first page is always downloaded

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: are you sure? is the satellite also recording while it's on the night side?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: are you sure? is the satellite also recording while it's on the night side?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: a case for trait specialization... maybe we can create a macro that builds NotImplented-stubs

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: I guess we could if we create another trait CloneableMetaData that has only that function. Shall I try it?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: 3bbd917d73e944b7697dfdd5f375cf69d7c52ffb

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: 3bbd917d73e944b7697dfdd5f375cf69d7c52ffb

operators/src/error.rs
jdroenner: maybe use Trait std::error::Error or Into Trait std::error::Error

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: add a todo

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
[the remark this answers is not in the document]
jdroenner: as above

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: or a different selection ui

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: no but then there is no tile at night. i think tats a better semantic then a tile which is for ever valid

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: as above

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: no then it is to much overhead. cant you simply implement it in the MetaData trait?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: i'm still not convinced that we need to require all the implementations everyehere. This is what happens if you do this. i guess there should be a better metadata provider instantiation / retrival mechanisim

operators/src/error.rs
michaelmattig: `std::error::Error` is not Send + Sync und thus our `Result` would no longer be Send + Sync. Also our own `Error` doesn't even implement `std::error::Error` or does it?

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: 3ed69c0b2d4e186ec765d553ddc10d46fc780e39

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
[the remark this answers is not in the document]
michaelmattig: 3ed69c0b2d4e186ec765d553ddc10d46fc780e39

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: we could have a special dialog for Sentinal datasets but I would prefer a generic solution

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: if I only shift by 1 minute there may be no tile at all because the latest tile was from the prevous day. ThatÄs what I menat.

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: if I only shift by one minute there may be no tile at all because the latest tile is from previous day

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: I don't know how. I used explicitly this workaround of having a box_clone method and implementing Clone for the Box because it's not possible IMHO to directly make the trait Clone

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: I can't see how that's possible, you would have to explain me how to

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: so add a task for it

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: yes but i think this is ok.  ypu need to aggregate anyway

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: yes but i think this is ok. ypu need to aggregate anyway

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: then we have to keep it like this

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: yes but that is a another task

operators/src/error.rs
jdroenner: you can do:

```
#[snafu(display("LoadingInfoError: {}", source))]
    LoadingInfo {
        source: Box<dyn std::error::Error + Send + Sync>, // TODO: find cleaner way to make errors in services crate visible
    },

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: we already have one https://trello.com/c/ebe6iQwz/115-bring-back-datasets-with-more-then-one-layer-band

operators/src/error.rs
michaelmattig: 4e19dc43043d20988a199028294ebe05e3ea7533

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: 89397a1bc02a646b760c2b2655eec984131b749b

services/src/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: 89397a1bc02a646b760c2b2655eec984131b749b

PR #265, opened 2021-06-10
also parse datetime and dates without timezone

jdroenner:


jdroenner:
[Fix Custom Time handling for OGR DAta](https://trello.com/c/VVksCjCC/217-fix-custom-time-handling-for-ogr-data)

operators/src/source/ogr_source.rs
```diff
@@ -533,8 +538,17 @@ where
             }),
             OgrSourceTimeFormat::Custom { custom_format } => Box::new(move |field: FieldValue| {
                 let date = field.into_string().ok_or(Error::OgrFieldValueIsNotString)?;
-                let date_time = DateTime::parse_from_str(&date, &custom_format)?;
-                Ok(date_time.timestamp_millis().try_into()?)
+                let date_time_result = DateTime::parse_from_str(&date, &custom_format)
```
michaelmattig, line 541: - can't you pull out the `timestamp_millis` because it's common among all three cases?
- can you please add a test for each case?

michaelmattig (changes requested):
.

operators/src/source/ogr_source.rs
jdroenner: i don't think so. its not the same type

PR #271, opened 2021-06-16
spatial partition

michaelmattig:


michaelmattig:
[Zwei BBoxes braucht das Land](https://trello.com/c/jMDDXCW2/189-zwei-bboxes-braucht-das-land)

datatypes/src/primitives/spatial_partition.rs
```diff
@@ -0,0 +1,221 @@
...
+        self.lower_right_coordinate.x - self.upper_left_coordinate.x
+    }
+
+    fn size_y(&self) -> f64 {
+        self.upper_left_coordinate.y - self.lower_right_coordinate.y
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ChristianBeilschmidt, line 209: I guess you should also test the intersection, contains, etc.

operators/src/engine/query.rs
```diff
@@ -1,13 +1,45 @@
...
     pub spatial_resolution: SpatialResolution,
 }
 
+/// A spatio-temporal rectangle for querying data with a spatial partition
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct RasterQueryRectangle {
+    pub partition: SpatialPartition,
+    pub time_interval: TimeInterval,
+    pub spatial_resolution: SpatialResolution,
+}
```
ChristianBeilschmidt, line 19: 1. I would'nt call it raster/vector, since we don't want to create a PlotQueryRectangle, I guess. So rather Partitioned and Bbox or something like this.
2. We could also say QueryRectangle<B: BoxShaped>, then we can have the rest in common.
3. We could then have only aliases for our types.

operators/src/engine/clonable_operator.rs
```diff
@@ -64,25 +63,17 @@ impl Clone for Box<dyn PlotOperator> {
...
-    type Descriptor: ResultDescriptor;
-    type Processor: QueryProcessor;
-
-    fn clone_boxed(&self) -> Box<dyn InitializedOperator<Self::Descriptor, Self::Processor>>;
-}
-
 /// Helper trait for making boxed `InitializedRasterOperator`s cloneable
 pub trait CloneableInitializedRasterOperator {
-    fn clone_boxed_raster(&self) -> Box<InitializedRasterOperator>;
+    fn clone_boxed_raster(&self) -> Box<dyn InitializedRasterOperator>;
```
ChristianBeilschmidt, line 68: Why wasn't this `dyn` before and what changed?

operators/src/engine/operator.rs
```diff
@@ -65,41 +64,80 @@ pub trait PlotOperator: CloneablePlotOperator + Send + Sync + std::fmt::Debug {
     }
 }
 
-pub type InitializedVectorOperator =
-    dyn InitializedOperator<VectorResultDescriptor, TypedVectorQueryProcessor>;
+pub trait InitializedRasterOperator: Send + Sync {
```
ChristianBeilschmidt, line 67: Why do we need a trait here instead of a concrete type?

operators/src/engine/operator.rs
```diff
@@ -65,41 +64,80 @@ pub trait PlotOperator: CloneablePlotOperator + Send + Sync + std::fmt::Debug {
...
     }
 }
 
-impl<R, Q> InitializedOperator<R, Q> for Box<dyn InitializedOperator<R, Q>>
-where
-    R: ResultDescriptor,
-    Q: QueryProcessor,
-{
-    fn result_descriptor(&self) -> &R {
+pub trait InitializedPlotOperator: Send + Sync {
```
ChristianBeilschmidt, line 99: Why do we need a trait here instead of a concrete type?

operators/src/engine/query_processor.rs
```diff
@@ -9,25 +11,14 @@ use geoengine_datatypes::plots::{PlotData, PlotOutputFormat};
...
-#[async_trait]
-pub trait QueryProcessor: Send + Sync {
-    type Output;
-    async fn query<'a>(
-        &'a self,
-        query: QueryRectangle,
-        ctx: &'a dyn QueryContext,
-    ) -> Result<BoxStream<'a, Result<Self::Output>>>;
-}
-
```
ChristianBeilschmidt, line 22: I guess there is no need to remove it, we can add another type, cf. https://github.com/geo-engine/geoengine/pull/269

operators/src/engine/query_processor.rs
```diff
@@ -39,29 +30,13 @@ pub trait RasterQueryProcessor: Sync + Send {
...
-    type RasterType = T;
-    async fn raster_query<'a>(
-        &'a self,
-        query: QueryRectangle,
-        ctx: &'a dyn QueryContext,
-    ) -> Result<BoxStream<'a, Result<RasterTile2D<Self::RasterType>>>> {
-        self.query(query, ctx).await
-    }
-}
-
```
ChristianBeilschmidt, line 57: I guess we can still default implement it and just pass in a different Qrect

operators/src/engine/query_processor.rs
```diff
@@ -111,44 +70,32 @@ pub trait PlotQueryProcessor: Sync + Send {
...
-impl<T> QueryProcessor for Box<dyn RasterQueryProcessor<RasterType = T>>
+impl<T> RasterQueryProcessor for Box<dyn RasterQueryProcessor<RasterType = T>>
 where
     T: Pixel,
 {
-    type Output = RasterTile2D<T>;
+    type RasterType = T;
 
-    async fn query<'a>(
+    async fn raster_query<'a>(
```
ChristianBeilschmidt, line 79: I guess we can still use query then.

operators/src/engine/query_processor.rs
```diff
@@ -111,44 +70,32 @@ pub trait PlotQueryProcessor: Sync + Send {
...
 #[async_trait]
-impl<V> QueryProcessor for Box<dyn VectorQueryProcessor<VectorType = V>>
+impl<V> VectorQueryProcessor for Box<dyn VectorQueryProcessor<VectorType = V>>
 where
     V: 'static,
 {
-    type Output = V;
-    async fn query<'a>(
+    type VectorType = V;
+    async fn vector_query<'a>(
```
ChristianBeilschmidt, line 94: same here

datatypes/src/primitives/mod.rs
```diff
@@ -29,9 +30,21 @@ pub use multi_line_string::{MultiLineString, MultiLineStringAccess, MultiLineStr
...
+pub use spatial_partition::{SpatialPartition, SpatialPartitioned};
 pub use spatial_resolution::SpatialResolution;
 pub use spatio_temporal_bounded::{SpatialBounded, TemporalBounded};
 
 pub use time_instance::TimeInstance;
 pub use time_interval::TimeInterval;
 pub use time_step::{TimeGranularity, TimeStep, TimeStepIter};
+
+/// Common trait for axis-parallel boxes
+pub trait BoxShaped: Copy {
```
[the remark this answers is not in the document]
michaelmattig: yes, BoundingBox2D is Copy I exchanged it with BoxShaped where appropriate and BoxShaped needs to be Copy in order for the code to continue to work

datatypes/src/primitives/mod.rs
```diff
@@ -29,9 +30,21 @@ pub use multi_line_string::{MultiLineString, MultiLineStringAccess, MultiLineStr
...
+    fn lower_left(&self) -> Coordinate2D;
+    fn upper_left(&self) -> Coordinate2D;
+    fn upper_right(&self) -> Coordinate2D;
+    fn lower_right(&self) -> Coordinate2D;
+
+    fn size_x(&self) -> f64;
+    fn size_y(&self) -> f64;
```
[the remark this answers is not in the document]
michaelmattig: that would be less efficient because the corner coordinates have to be built first

operators/src/engine/query.rs
michaelmattig: 1. I find it very convenient that it's clear by the name which QueryRectangle to use. We can have a PlotQueryRectangle that is a type alias for VectorQueryRectangle
2. would be possible but we wouldn't save a lot. I think it's better if they are separate types especially because it's possible to turn a vectorQueryRectangle into a RasterQueryRectangle but not vice versa.

operators/src/engine/clonable_operator.rs
michaelmattig: it was always dyn but hidden inside the type alias

operators/src/engine/operator.rs
michaelmattig: it was always a trait. It was a type alias for a trait

operators/src/engine/operator.rs
```diff
@@ -65,41 +64,80 @@ pub trait PlotOperator: CloneablePlotOperator + Send + Sync + std::fmt::Debug {
...
+    fn boxed(self) -> Box<dyn InitializedRasterOperator>
+    where
+        Self: Sized + 'static,
+    {
+        Box::new(self)
+    }
+}
 
-pub trait InitializedOperator<Descriptor, Processor>: Send + Sync {
+pub trait InitializedVectorOperator: Send + Sync {
```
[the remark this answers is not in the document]
michaelmattig: I didn't design it this way but just split up the type

operators/src/engine/operator.rs
michaelmattig: I didn't design it this way but just split up the type

operators/src/engine/query_processor.rs
michaelmattig: We don't use `query` anymore so I guess we don't need the common trait anymore

operators/src/engine/query_processor.rs
michaelmattig: why do we need it?

operators/src/engine/query_processor.rs
michaelmattig: why do we need it?

operators/src/engine/query_processor.rs
michaelmattig: why do we need it?

datatypes/src/primitives/bounding_box.rs
```diff
@@ -424,6 +342,90 @@ impl BoundingBox2D {
     }
 }
 
+impl BoxShaped for BoundingBox2D {
```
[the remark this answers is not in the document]
michaelmattig: ff87f5795cf2e8f04760764f8700c26e4dcf0e25

datatypes/src/primitives/spatial_partition.rs
michaelmattig: 319239ad65edb889ea3181ed6e30f0a9674a5dfd

datatypes/src/primitives/mod.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Looking into the future, doesn't this prevent a raster tile from being `BoxShaped`?
Maybe we need to bound on QrRect rather than here :thinking:

operators/src/engine/query.rs
ChristianBeilschmidt: I guess we have to discuss this.

operators/src/engine/query_processor.rs
ChristianBeilschmidt: We can use query, you just change it in this PR.

operators/src/engine/query_processor.rs
ChristianBeilschmidt: We still have a common interface for querying stuff.

operators/src/engine/query_processor.rs
ChristianBeilschmidt: We still have a common interface for querying stuff.

datatypes/src/primitives/mod.rs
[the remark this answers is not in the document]
michaelmattig: RasterTile shouldn't be BoxShaped. It has a TileInformation which is SpatialPartitioned and the SpatialPartition is BoxShaped. The trait is just a convenience for operations that operate on der border coordinates.

operators/src/engine/query_processor.rs
michaelmattig: but what is the benefit of using a common query method if both the input and the output types are different?

operators/src/engine/query_processor.rs
michaelmattig: but what is the benefit of using a common query method if both the input and the output types are different?

operators/src/engine/query_processor.rs
michaelmattig: but what is the benefit of using a common query method if both the input and the output types are different?

operators/src/engine/query_processor.rs
michaelmattig: but what is the benefit of using a common query method if both the input and the output types are different?

operators/src/engine/query_processor.rs
ChristianBeilschmidt: It is not different, we have the same method name, a context and something that is BoxShaped

PR #272, opened 2021-06-17
tag enums with type

michaelmattig:


michaelmattig:
[API enums anders taggen](https://trello.com/c/fdqe0BLV/106-api-enums-anders-taggen)

operators/src/mock/mock_raster_source.rs
```diff
@@ -189,7 +188,9 @@ mod tests {
                 "resultDescriptor": {
                     "dataType": "U8",
                     "spatialReference": "EPSG:4326",
-                    "measurement": "unitless",
+                    "measurement": {
+                        "type": "unitless"
+                    },
```
ChristianBeilschmidt, line 193: Is this really better in this case?
Maybe "unitless" or { type…} ?

operators/src/plot/histogram.rs
```diff
@@ -972,7 +972,8 @@ mod tests {
                     "type": "OgrSource",
                     "params": {
                         "dataset": {
-                            "internal": dataset_id
+                            "type": "internal",
+                            "dataset": dataset_id
```
ChristianBeilschmidt, line 976:
```suggestion
                            "id": dataset_id
```

operators/src/plot/histogram.rs
```diff
@@ -983,7 +984,9 @@ mod tests {
 
         let mut execution_context = MockExecutionContext::default();
         execution_context.add_meta_data(
-            DatasetId::Internal(dataset_id),
+            DatasetId::Internal {
+                dataset: dataset_id,
```
ChristianBeilschmidt, line 988:
```suggestion
                id: dataset_id,
```

operators/src/source/ogr_source.rs
```diff
@@ -1058,14 +1058,13 @@ mod tests {
...
-                        },
-                        "duration": 42
-                    }
+                    "type": "start",
+                    "startField": "start",
+                    "startFormat": {
+                        "format": "custom",
+                        "customFormat": "YYYY-MM-DD"
+                    },
+                    "duration": 42
```
ChristianBeilschmidt, line 1067: I'm not sure if this improves this setting since we could specify "start" twice now

services/src/projects/project.rs
```diff
@@ -210,7 +211,7 @@ impl Layer {
     pub fn layer_type(&self) -> LayerType {
         match self.symbology {
             Symbology::Raster(_) => LayerType::Raster,
-            Symbology::Vector(_) => LayerType::Vector,
+            _ => LayerType::Vector,
```
ChristianBeilschmidt, line 214: I would recommend reverting this change since it is more error-prone to have this as catch all

services/src/projects/project.rs
```diff
@@ -583,9 +582,10 @@ mod tests {
                         "legend": false,
                     },
                     "symbology": {
-                        "raster": {
-                            "opacity": 1.0,
-                            "colorizer": "rgba"
+                        "type": "raster",
```
ChristianBeilschmidt, line 585: So we differentiate between raster for raster and polygon/point/line for vector?
Maybe this is okay but what is your rational behind it?

services/src/projects/project.rs
```diff
@@ -659,49 +659,52 @@ mod tests {
 
     #[test]
     fn serialize_symbology() {
-        let symbology = Symbology::Vector(VectorSymbology::Point(PointSymbology {
-            radius: NumberParam::Static(1),
+        let symbology = Symbology::Point(PointSymbology {
+            radius: NumberParam::Static { value: 1 },
```
ChristianBeilschmidt, line 663: Why do we need a dict for this?

services/test-data/dataset_defs/ndvi (3587).json
```diff
@@ -1,51 +1,53 @@
...
     "sourceOperator": "GdalSource"
   },
   "metaData": {
-    "GdalMetaDataRegular": {
-      "resultDescriptor": {
-        "dataType": "U8",
-        "spatialReference": "EPSG:3857",
-        "measurement": "unitless",
-        "noDataValue": 0.0
+    "type": "gdalMetaDataRegular",
```
ChristianBeilschmidt, line 12: looks weird, but I guess it's okay(?)

ChristianBeilschmidt (changes requested):

operators/src/mock/mock_raster_source.rs
michaelmattig: it's at least consistent to always have a type

operators/src/plot/histogram.rs
michaelmattig: I think it's more consistent to call the dataset_id `dataset` in the internal and external case

operators/src/plot/histogram.rs
michaelmattig: I think it's more consistent to call the dataset_id `dataset` in the internal and external case

operators/src/source/ogr_source.rs
michaelmattig: please elaborate, I don't get the problem

services/src/projects/project.rs
michaelmattig: there is no `VectorSymbology` any more because it was inconvient to have `Symbology::Vector(VectorSymbology::Point`. I'd much rather split the `Layer` into a raster and a vector variant like in the frontend. but that's not part of this PR.

services/src/projects/project.rs
michaelmattig: it was inconvient to have symbology -> vectorSymbology -> pointSymbology.

services/src/projects/project.rs
michaelmattig: because I changed the serialization to tag with type and specified `value` as attributename for the payload.

services/test-data/dataset_defs/ndvi (3587).json
michaelmattig: it's at least consistent with all the other fields that are camelCase. We could make an exception for operators (which are still Capitalized I believe) and meta data. But where do we draw the line then?

operators/src/plot/histogram.rs
ChristianBeilschmidt: You have

```json
{
 "dataset": {
    "dataset":  "name"
  }
}
```

operators/src/source/ogr_source.rs
ChristianBeilschmidt: We had fields in the dict, namely start, and end.
And now you add basically just a list of dicts and specify it via *type*.
This doesn't pose an improvement I guess, but on the other hand a user can specify "start" twice.

ChristianBeilschmidt:
```
{
 "dataset": {
    "type": "internal",
    "dataset_id":  "111-4444-5555"
  }
}
{
 "dataset": {
    "type": "external",
    "provider_id": "111-111-111",
    "dataset_id":  "abc"
  }
}
```

ChristianBeilschmidt (approved):

PR #277, opened 2021-06-23
open gdal datasets in operators crate

michaelmattig:


michaelmattig:
[Gdal-Problematik in Tests](https://trello.com/c/QvlU0C1O/212-gdal-problematik-in-tests)

operators/src/util/gdal.rs
```diff
@@ -69,3 +75,11 @@ pub fn add_ndvi_dataset(ctx: &mut MockExecutionContext) -> DatasetId {
     ctx.add_meta_data(id.clone(), Box::new(create_ndvi_meta_data()));
     id
 }
+
+pub fn open_dataset(path: &Path) -> Result<Dataset> {
```
jdroenner, line 79: please add a comment that this is a workaround to avoid strange gdal errors

operators/src/util/gdal.rs
```diff
@@ -69,3 +75,11 @@ pub fn add_ndvi_dataset(ctx: &mut MockExecutionContext) -> DatasetId {
...
+    Dataset::open(&path).context(error::Gdal)
+}
+
+pub fn open_dataset_ex(path: &Path, dataset_options: DatasetOptions) -> Result<Dataset> {
```
jdroenner, line 83: please add a comment that this is a workaround to avoid strange gdal errors

jdroenner (changes requested):
i think we should also prefix the methods with gdal_ or call themn open_gdal_dataset...

PR #278, opened 2021-06-23
case insensitive wms bool

ChristianBeilschmidt:


ChristianBeilschmidt:
[case-insensitive WMS bools](https://trello.com/c/77namyB6/229-case-insensitive-wms-bools)

PR #279, opened 2021-06-23
partitioned_bbox_2

ChristianBeilschmidt:


ChristianBeilschmidt:
[Zwei BBoxes braucht das Land](https://trello.com/c/jMDDXCW2/189-zwei-bboxes-braucht-das-land)

michaelmattig:
> 
> 
> Hm, this is inverse to what exactly?
> 
> And if you have half-open intervals, why do you exclude min?

it's inverse because it's `min < value <= max` instead of `min <= value < max`. It's necessary because for a spatial partition in y-direction the y-value of the lower right coordinate is not part of the partition. Thus I have to check wether the value is > min and <= max.

operators/src/processing/raster_vector_join/points.rs
```diff
@@ -246,14 +247,15 @@ impl<P: Pixel> PointRasterJoiner<P> {
 }
 
 #[async_trait]
-impl VectorQueryProcessor for RasterPointJoinProcessor {
-    type VectorType = MultiPointCollection;
+impl QueryProcessor for RasterPointJoinProcessor {
```
michaelmattig, line 250: why is it better to use `QueryProcessor` here and define the Output and SpatialBounds manually when you can just use the VectorQueryProcessor which defines at least the latter one for you? The query-method is still available even if you implement VectorQueryProcessor here

michaelmattig: the same holds for the other processors you changed

michaelmattig (changes requested):
I think it would be better to implement the QueryProcessors as RasterQueryProcessor and VectorQueryProcessor, cf. my comment. Also please merge the master into the PR

operators/src/processing/raster_vector_join/points.rs
ChristianBeilschmidt: It isn't better, it is just still possible. You can use both when they fit best.

michaelmattig: then why change it if it isn't better? It's easier to understand the processor if you use the more precise type. Also please merge the master into this pr,

ChristianBeilschmidt: I tried it, you can have problems using query on non-boxed types that implement just VectorQueryProcessor or RasterQueryProcessor since we can only implement
```
impl QueryProcessor for T
where T: RasterQueryProcessor
```
once. If you implement QueryProcessor, it automatically implements VectorQueryProcessor or RasterQueryProcessor that works all the time. So it is just one additional line and it works in more cases.

michaelmattig (approved):

michaelmattig (approved):

PR #283, opened 2021-06-29
add optional symbology to dataset

michaelmattig:


michaelmattig:
[Dataset-Info hat Default-Symbology](https://trello.com/c/HjsYyo9I/27-dataset-info-hat-default-symbology)

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -167,6 +167,7 @@ impl SentinelS2L2aCogsDataProvider {
                             no_data_value: band.no_data_value,
                         }
                         .into(),
+                        symbology: None, // TODO
```
ChristianBeilschmidt, line 170: Can't you specify this 1-6000 or what this is, here?

services/src/datasets/storage.rs
```diff
@@ -77,6 +81,7 @@ impl From<ImportDataset> for Dataset {
             description: value.description,
             result_descriptor: value.result_descriptor,
             source_operator: value.source_operator,
+            symbology: None,
```
ChristianBeilschmidt, line 84: We should have at least one test for the dataset and one for the handler I guess.

ChristianBeilschmidt (changes requested):

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: the colorizer is different for each band, so I can't easily specify it here

services/src/datasets/storage.rs
michaelmattig: what do you want me to test?

ChristianBeilschmidt: We should test the case where we have specified a symbology and return it to the client.
Then, we know when something breaks in our return type.

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
ChristianBeilschmidt: We can use 0-10k and implement an if-else-block later.
Currently, you don't see anything which isn't a working solution.

https://gis.stackexchange.com/questions/233874/what-is-the-range-of-values-of-sentinel-2-level-2a-images

services/src/datasets/storage.rs
michaelmattig: do you mean checking the serialization of a Dataset? I can do it, but we already have tests for Symbology serialization

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: caca41556981504247dd1879887e075b91eedc47

PR #284, opened 2021-06-30
impr-tile-loading

jdroenner:


jdroenner:
[tune tile loading (fokus on s3 tiles)](https://trello.com/c/qu8vBqIp/243-tune-tile-loading-fokus-on-s3-tiles)

operators/src/source/gdal_source.rs
```diff
@@ -531,6 +551,14 @@ impl RasterOperator for GdalSource {
 
         debug!("Initializing GdalSource for {:?}.", &self.params.dataset);
 
+        // TODO: move somewhere else?
+        // gdal::config::set_config_option("GDAL_NUM_THREADS", "2")?;
```
michaelmattig, line 555: remove commented out settings

operators/src/source/gdal_source.rs
```diff
@@ -531,6 +551,14 @@ impl RasterOperator for GdalSource {
...
+        gdal::config::set_config_option("GDAL_DISABLE_READDIR_ON_OPEN", "TRUE")?;
```
michaelmattig, line 556: so these are global options which affect all subsquent gdal operations? Is this what we want? Then we should set them on start of the server. Otherwise we need to encapsulate this stuff somehow. Maybe a global lock on gdal and resetting the config after opening? Do they only take effect during dataset_open?

michaelmattig (changes requested):
.

operators/src/source/gdal_source.rs
jdroenner: https://gdal.org/user/configoptions.html

it is also possible to use them on a local thread. should we add something to loadinginfo to provide them?

operators/src/source/gdal_source.rs
jdroenner: e714a5a

operators/src/source/gdal_source.rs
michaelmattig: I don't think we always want  GDAL_DISABLE_READDIR_ON_OPEN because that also means missing additional meta data like overviews that lie next to the specified file(?). Or can we specify the additional files explicitly? Then we could always disable the readdir I think. Otherwise we'd need to extend the loading info and only set it for the current thread. But I guess we'd need to reset the config afterwards because our threads are reused by tokio?

michaelmattig:
there is a test failing btw

operators/src/source/gdal_source.rs
jdroenner: i removed it. we should think about a concept to enhance the loading info.

jdroenner:
i fixed the test

PR #290, opened 2021-07-13
inner aggregation

michaelmattig:


michaelmattig:
[Innere Aggregation bei Raster-Vektor-Join](https://trello.com/c/iqYOGYuJ/92-innere-aggregation-bei-raster-vektor-join)

operators/src/processing/raster_vector_join/aggregator.rs
```diff
@@ -93,6 +101,14 @@ impl TypedAggregator {
             TypedAggregator::MeanNumber(aggregator) => aggregator.is_satisfied(),
         }
     }
+
```
jdroenner, line 104: why not impl Aggregator for TypedAggregator?

operators/src/processing/raster_vector_join/aggregator.rs
```diff
@@ -175,6 +191,37 @@ where
     fn is_satisfied(&self) -> bool {
         self.number_of_pristine_values == 0 || self.number_of_non_null_values == 0
     }
+
+    fn add_feature_data(&mut self, data: FeatureData, weight: u64) -> Result<()> {
```
jdroenner, line 195: i think this could be a default implementation in the trait.

operators/src/processing/raster_vector_join/aggregator.rs
```diff
@@ -278,6 +325,37 @@ impl Aggregator for MeanValueAggregator {
     fn is_satisfied(&self) -> bool {
         self.number_of_non_null_values == 0
     }
+
+    fn add_feature_data(&mut self, data: FeatureData, weight: u64) -> Result<()> {
```
jdroenner, line 329: same as above

operators/src/processing/raster_vector_join/mod.rs
```diff
@@ -17,30 +17,53 @@ use async_trait::async_trait;
...
+    /// Specifies which method is used for aggregating values for a feature
+    pub feature_aggregation: FeatureAggregationMethod,
+
+    /// Specifies which method is used for aggregating values over time
+    pub temporal_aggregation: TemporalAggregationMethod,
+}
+
+/// How to aggreagate the values for the geometries inside a feature
+/// e.g. add all the raster values corresponding to the individual
+/// points inside a `MultiPoint` feature. Choosing None leads to an
```
jdroenner, line 51: there is no none

jdroenner (changes requested):
looks fine so far. lets discuss it

operators/src/processing/raster_vector_join/aggregator.rs
michaelmattig: I tried it, but then the TypedAggregator would need to be generic and this requires a lot of changes. I don't think it's worth the change but maybe you can try to do it

operators/src/processing/raster_vector_join/aggregator.rs
michaelmattig: see above

operators/src/processing/raster_vector_join/aggregator.rs
michaelmattig: see above

operators/src/processing/raster_vector_join/mod.rs
michaelmattig: b3ed11c817df2a474264b18f148eedec8b11e407

PR #291, opened 2021-07-15
make WCS more flexible s.t. owslib may access it

michaelmattig:


michaelmattig:
[Auf Raster mit WCS zugreifen](https://trello.com/c/P34mfEge/232-auf-raster-mit-wcs-zugreifen)

services/src/handlers/wcs.rs
```diff
@@ -92,7 +93,12 @@ async fn get_capabilities<C: Context>(
                 <ows:Title>Web Coverage Service</ows:Title>
                 <ows:ServiceType>WCS</ows:ServiceType>
                 <ows:ServiceTypeVersion>1.1.1</ows:ServiceTypeVersion>
+                <ows:Fees>NONE</ows:Fees>
+                <ows:AccessConstraints>NONE</ows:AccessConstraints>
             </ows:ServiceIdentification>
+            <ows:ServiceProvider>
+                <ows:ProviderName>Provider Name</ows:ProviderName>
```
jdroenner, line 100: Geo Engine?

services/src/handlers/wcs.rs
```diff
@@ -311,18 +318,19 @@ async fn get_coverage<C: Context>(
...
-        }
-    };
+    let spatial_resolution: SpatialResolution =
+        if let Some(spatial_resolution) = request.spatial_resolution() {
+            spatial_resolution?
+        } else {
+            // TODO: proper default resolution
+            SpatialResolution {
+                x: request_partition.size_x() / 256.,
+                y: request_partition.size_y() / 256.,
```
jdroenner, line 328: config params?

services/src/handlers/wcs.rs
```diff
@@ -466,7 +474,12 @@ mod tests {
                 <ows:Title>Web Coverage Service</ows:Title>
                 <ows:ServiceType>WCS</ows:ServiceType>
                 <ows:ServiceTypeVersion>1.1.1</ows:ServiceTypeVersion>
+                <ows:Fees>NONE</ows:Fees>
+                <ows:AccessConstraints>NONE</ows:AccessConstraints>
             </ows:ServiceIdentification>
+            <ows:ServiceProvider>
+                <ows:ProviderName>Provider Name</ows:ProviderName>
```
jdroenner, line 481: Geo Engine?

services/src/ogc/util.rs
```diff
@@ -347,27 +358,11 @@ mod tests {
...
-                    (-162., 81.).into(),
-                    (162., -81.).into()
-                ),
-                spatial_reference: SpatialReference::new(SpatialReferenceAuthority::Epsg, 3857),
-            }
-        )
-    }
-
-    #[test]
-    fn it_parses_wcs_bbox_epsg_4326() {
```
jdroenner, line 360: why remove a test?

services/src/handlers/wcs.rs
michaelmattig: This is the operator of the WCS endpoint. We wouln't want to have Geo Engine there if somebody else hosts it.

services/src/handlers/wcs.rs
michaelmattig: I am not sure. I think WCS would respond with the native resolution if none is specified.

services/src/handlers/wcs.rs
michaelmattig: This is the operator of the WCS endpoint. We wouln't want to have Geo Engine there if somebody else hosts it.

services/src/ogc/util.rs
michaelmattig: because the axis ordering is no longer done during parsing now, as the spatial reference is optional now. It has to be done in a separate stept and is tested there

services/src/handlers/wcs.rs
jdroenner: but there is no native resolution i the engine

michaelmattig: isn't it still the plan to support that one day? I can make a configuration but not sure if that makes any sense because it is constant for all zoom levels and extents. If we don't plan to make native resolutions available some day I would suggest to throw an error if no resolution is specified.

PR #295, opened 2021-07-20
first version of natur40 provider

michaelmattig:


michaelmattig:
[Wölli-RasterDB-Provider](https://trello.com/c/rsVzz0bB/239-w%C3%B6lli-rasterdb-provider)

operators/src/source/gdal_source.rs
```diff
@@ -385,7 +388,18 @@ where
...
+            GdalDataset::open_ex(
+                &dataset_params.file_path,
+                DatasetOptions {
+                    open_options: Some(options.as_slice()),
+                    ..DatasetOptions::default()
+                },
+            )
+        } else {
+            GdalDataset::open(&dataset_params.file_path)
+        };
```
ChristianBeilschmidt, line 402: We could use only one open_ex call and pass `None` or an empty vector. This would mean less paths :shrug:.

operators/src/util/gdal.rs
```diff
@@ -88,3 +93,58 @@ pub fn gdal_open_dataset(path: &Path) -> Result<Dataset> {
...
+pub fn gdal_parameters_from_dataset(
+    dataset: &Dataset,
+    band: isize,
+    path: &Path,
+) -> Result<GdalDatasetParameters> {
+    let rasterband = &dataset.rasterband(band)?;
+
+    Ok(GdalDatasetParameters {
+        file_path: PathBuf::from(path),
+        rasterband_channel: usize::try_from(band).unwrap_or(0), // TODO: is the band in the metadata Dataset the same as in the actual data file?
```
ChristianBeilschmidt, line 141: What do you mean with metadata dataset?

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+        Ok(Box::new(Natur40DataProvider {
+            id: self.id,
+            base_url: self.base_url,
+            user: self.user,
+            password: self.password,
+        }))
+    }
+
+    fn type_name(&self) -> String {
+        "Natur4.0".to_owned()
```
ChristianBeilschmidt, line 59:
```suggestion
        "Nature 4.0".to_owned()
```

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+    rasterdbs: Vec<RasterDb>,
+    #[serde(deserialize_with = "string_or_string_array")]
+    tags: Vec<String>,
+    session: String,
+}
+
+#[async_trait]
+impl DatasetProvider for Natur40DataProvider {
+    async fn list(&self, _options: Validated<DatasetListOptions>) -> Result<Vec<DatasetListing>> {
+        // TODO: query the other dbs as well
```
ChristianBeilschmidt, line 111: Which other dbs?

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+            .map(|db| self.load_dataset(db.url(&self.base_url)));
+        let datasets: Vec<Result<gdal::Dataset>> = join_all(datasets).await;
+
+        for (db, dataset) in raster_dbs.rasterdbs.iter().zip(&datasets) {
+            if let Ok(dataset) = dataset {
+                for band_index in 1..dataset.raster_count() {
+                    if let Ok(result_descriptor) =
+                        raster_descriptor_from_dataset(&dataset, band_index, None)
+                    {
+                        // TODO: get label from rasterband
```
ChristianBeilschmidt, line 128: Sohaib is trying to get this info?

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+                            id: DatasetId::External(ExternalDatasetId {
+                                provider_id: self.id,
+                                dataset_id: format!("{}:{}", db.name.clone(), band_index),
+                            }),
+                            name: db.title.clone(),
+                            description: format!("Band: {}", band_index),
+                            tags: db.tags.clone(),
+                            source_operator: "GdalSource".to_owned(),
+                            result_descriptor: TypedResultDescriptor::Raster(result_descriptor),
+                            symbology: None, // TODO: build symbology
```
ChristianBeilschmidt, line 140: Does Stephan provide infos about this?
If so, we should add at least a task to the story.

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+
+        Ok(listing
+            .into_iter()
+            .filter_map(|d: Result<DatasetListing>| if let Ok(d) = d { Some(d) } else { None })
+            .collect())
+    }
+
+    async fn load(
+        &self,
+        // _session: S,
```
ChristianBeilschmidt, line 162:
```suggestion
```

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+                    ..DatasetOptions::default()
+                },
+            )?;
+            Ok(dataset)
+        })
+        .await
+        .context(error::TokioJoin)?
+    }
+
+    async fn load_raster_dbs(&self) -> Result<RasterDbs> {
```
ChristianBeilschmidt, line 193: Are there multiple?

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+
+        let db_url = RasterDb::url_from_name(&self.base_url, db_name);
+        let dataset = self.load_dataset(db_url.clone()).await.map_err(|e| {
+            geoengine_operators::error::Error::LoadingInfo {
+                source: Box::new(e),
+            }
+        })?;
+
+        Ok(Box::new(GdalMetaDataStatic {
+            time: None,
```
ChristianBeilschmidt, line 248: Every dataset is static from this type of db?

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+        Box<
+            dyn MetaData<
+                MockDatasetDataSourceLoadingInfo,
+                VectorResultDescriptor,
+                VectorQueryRectangle,
+            >,
+        >,
+        geoengine_operators::error::Error,
+    > {
+        todo!()
```
ChristianBeilschmidt, line 273: This panics, maybe return an error instead?

```suggestion
        Err(geoengine_operators::error::Error::NotYetImplemented)
```

services/src/handlers/datasets.rs
```diff
@@ -93,6 +93,7 @@ async fn list_external_datasets<C: Context>(
         .await?
         .list(options)
         .await?;
+    // TODO: it appears errors here lead to the internal datasets being listed because the route also matches /datasets,
```
ChristianBeilschmidt, line 96: I guess this _must_ be fixed on the Actix change?

services/src/datasets/external/natur40.rs
```diff
@@ -0,0 +1,290 @@
...
+    source::{GdalLoadingInfo, OgrSourceDataset},
+};
+use log::info;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use snafu::ResultExt;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Natur40DataProviderDefinition {
```
ChristianBeilschmidt, line 38: I would recommend renaming the file and the providers to *Nature40*, cf. https://www.uni-marburg.de/en/fb19/natur40/about-us/nature-4-0-in-a-nutshell

operators/src/util/gdal.rs
michaelmattig: the function gets a dataset, band index and a path, and builds the GdalDatasetParameters  from the info in the dataset and uses the given path as file_path. But the path could point to a different file than the Dataset comes from. So you could use the fourth band of dataset a to build the parameters for opening dataset b which only has one band.

services/src/datasets/external/natur40.rs
michaelmattig: currently we don't have spaces in type names. they aren't supposed to be used as text anyway (even though currently they are) but to distinguish different types of providers

services/src/datasets/external/natur40.rs
michaelmattig: I only query the rasterdb but there are others in the project that we have to integrate at some point

services/src/datasets/external/natur40.rs
michaelmattig: no, we only get the infos about the dataset that I mapped in the `RasterDbs` type

services/src/datasets/external/natur40.rs
michaelmattig: yes every dataset has its own db

services/src/datasets/external/natur40.rs
michaelmattig: I think yes

services/src/handlers/datasets.rs
michaelmattig: in actix this shouldn't happen

operators/src/source/gdal_source.rs
michaelmattig: 93573df8a3d72df781b957bc4249b90c9b63a5d3

services/src/datasets/external/natur40.rs
michaelmattig: 64f671c676b48a0741b2d54adb495a5cbe2abd8e

services/src/datasets/external/natur40.rs
michaelmattig: 1944f13ce4342297fba7bfcb43637b0ffe59e3a4

services/src/datasets/external/natur40.rs
michaelmattig: f65088f41f89f69c260284e8f82ee3c3b851bf5f

services/src/datasets/external/natur40.rs
michaelmattig: f65088f41f89f69c260284e8f82ee3c3b851bf5f

services/src/datasets/external/natur40.rs
ChristianBeilschmidt: Ah, and you would integrate them into this provider?

services/src/datasets/external/natur40.rs
ChristianBeilschmidt: Are there many labels and is this called often?
Since delete(0) is an expensive op for a vector, you can decide if it is worth it changing it to dismiss it in the loop or just leave it as it is.

operators/src/util/gdal.rs
ChristianBeilschmidt: Is it worth it checking this or do we just get an error (no panic) then?

services/src/datasets/external/natur40.rs
ChristianBeilschmidt: Maybe we (=Sohaib) can discuss this with Stephan if there is default colorization. This would mean either creating a task or just leaving the comment for later.

services/src/datasets/external/natur40.rs
ChristianBeilschmidt: Maybe collect this (and the colorization) in a list of open topics to discuss for the Nature 4.0 guys?

ChristianBeilschmidt (approved):

operators/src/util/gdal.rs
michaelmattig: maybe we could introduce another parameter for the band. Then the caller can decide which band to take for creating the Loading Info from and which one to use to load the data?

services/src/datasets/external/natur40.rs
michaelmattig: I would imagine them all in this provider and the possibility to filter the external datasets by type etc. I think it would be worse having to look up project data in multiple places

services/src/datasets/external/natur40.rs
michaelmattig: there is one label for each band in the dataset and the datasets have only 1 or 2 bands. For each dataset we perform a WCS request so the delete() should be insignificant

services/src/datasets/external/natur40.rs
michaelmattig: https://trello.com/c/IlvLaYLo

services/src/datasets/external/natur40.rs
michaelmattig: I added a checklist to the story card

operators/src/util/gdal.rs
michaelmattig: e49700a325e17c2a5dec1dfee35beadc4f0297ca

services/src/error.rs
```diff
@@ -180,9 +185,17 @@ pub enum Error {
     WcsBoundingboxCrsMustEqualGridBaseCrs,
     WcsInvalidGridOffsets,
 
+    ExpectedExternalDatasetId,
+    InvalidExternalDatasetId {
+        provider: DatasetProviderId,
+    },
+
+    Natur40UnknownRasterDbname,
```
ChristianBeilschmidt, line 193: Feature gate?

services/Cargo.toml
```diff
@@ -53,6 +56,7 @@ log = "0.4"
 clap = "3.0.0-beta.2"
 tempfile = "3.1"
 xml-rs = "0.8.3"
+httptest = "0.15.2"
```
ChristianBeilschmidt, line 59: maybe sort?

ChristianBeilschmidt (changes requested):

services/src/error.rs
michaelmattig: 007e9af6875a625fcd0e8604650e29aae6dd4d7b

services/Cargo.toml
michaelmattig: 007e9af6875a625fcd0e8604650e29aae6dd4d7b

PR #299, opened 2021-07-24
add debug for stac and gdal_source

jdroenner:


jdroenner:
[fix ad debug stac](https://trello.com/c/ERgcyZey/276-fix-ad-debug-stac)

datatypes/src/raster/grid_traits.rs
```diff
@@ -146,9 +146,12 @@ pub trait NoDataValue {
 
     fn no_data_value(&self) -> Option<Self::NoDataType>;
 
+    #[allow(clippy::eq_op)]
     fn is_no_data(&self, value: Self::NoDataType) -> bool {
-        self.no_data_value()
-            .map_or(false, |no_data_value| value == no_data_value)
+        self.no_data_value().map_or(false, |no_data_value| {
+            // value is equal no-data value or both are NAN.
+            value == no_data_value || no_data_value != no_data_value && value != value
```
ChristianBeilschmidt, line 153: 1. Haven't we run into this problem already?

2. So if value != value but NO DATA is not NAN, then it is not NODATA?

3. Moreover, I guess it would help to put braces to this statement.

4. I guess we also should add a test to this function for checking that it works. It is pretty central.

operators/src/source/gdal_source.rs
```diff
@@ -530,14 +530,17 @@ where
...
-
         debug!(
             "Querying GdalSourceProcessor<{:?}> with: {:?}.",
             P::TYPE,
             &query
         );
 
+        let meta_data = self.meta_data.loading_info(query).await?;
+
+        debug!("GdalLoadingInfo: {:?}.", &meta_data);
```
ChristianBeilschmidt, line 541: Doesn't this mean we get the loading info anyway but throw it away usually?

ChristianBeilschmidt (changes requested):

datatypes/src/raster/grid_traits.rs
jdroenner: 93be0ec

ChristianBeilschmidt (approved):

PR #303, opened 2021-07-28
get dataset ids from operators and provenance infos from dataset providers

michaelmattig:


michaelmattig:
[Zitate (im System und Tabelle)](https://trello.com/c/1U87RGjN/269-zitate-im-system-und-tabelle)

PR #320, opened 2021-08-13
Statistics for operators

jdroenner:
This is an idea how to get processing stats from the workflow graph.
The idea is to inject / wrap each operator with a statistics provider.
Instead of logging calls, the statistics operator / query_processor / stream_adapter would send messages to a channel provided by the (pro) context OR a global channel.


jdroenner:
[Operator Statistik](https://trello.com/c/fo0HU2Oy/334-operator-statistik)

PR #326, opened 2021-08-19
default_time

michaelmattig:


michaelmattig:
[Time Support für WCS](https://trello.com/c/D8PHHecn/262-time-support-f%C3%BCr-wcs)

Settings-default.toml
```diff
@@ -51,6 +51,12 @@ enable_buffering = false
 # Use this option if another folder should be used.
 #log_directory = "/var/log/"
 
+[ogc.default_time]
+# type "Value" with start/end as ISO strings or "Now" for using the current time, use wcs.default_time etc. for override
+type = "Value"
```
ChristianBeilschmidt, line 56: Are Value and Now case sensitive? And if so, could we change that? Or at least also support value and now?

ChristianBeilschmidt (changes requested):

Settings-default.toml
michaelmattig: I use serde type tag which per default uses the name of the enum variant as it is. I don't know if we want to apply the same guidelines like in the API. But then we would also need to use camelCase for `default_time`. I think in order to make it case insensitive we need a custom deserializer but I am not sure about that.

I would leave it like this but I can also change it to lower case if you want.

ChristianBeilschmidt: Maybe just allow the lowercase variant as well and then we are fine. I guess chances are large that people write "value" but less likely that they write "vALue".

michaelmattig: #[serde(alias = "now")]

ChristianBeilschmidt (approved):

PR #329, opened 2021-08-24
Add pangaea as data provider

koerberm:


koerberm:
[PANGAEA als externe Datenquelle](https://trello.com/c/PDs4yAKz/271-pangaea-als-externe-datenquelle)

services/test-data/pangaea/pangaea_geo_none.tsv
```diff
@@ -0,0 +1,118 @@
+/* DATA DESCRIPTION:
```
ChristianBeilschmidt, line 1: For all inserted datasets: Did you check that we are allowed to use the data, e.g. CC0 or something similar?

operators/src/source/gdal_source.rs
```diff
@@ -560,6 +570,10 @@ where
 
         debug!("GdalLoadingInfo: {:?}.", &meta_data);
 
+        if let Some(hook) = self.meta_data.pre_load_hook() {
+            hook.execute().await?
```
ChristianBeilschmidt, line 574:
```suggestion
            hook.execute().await?;
```

services/src/datasets/external/pangaea/meta.rs
```diff
@@ -0,0 +1,624 @@
...
+                .iter()
+                .filter(|&c| matches!(c, PangeaParam::String { .. }))
+                .map(PangeaParam::full_name)
+                .collect(),
+        }
+    }
+
+    fn get_ogr_source_ds(&self, memfile_name: &str) -> OgrSourceDataset {
+        let path: PathBuf = PathBuf::from(memfile_name);
+        let layer = path.file_stem().unwrap().to_str().unwrap().to_string();
```
ChristianBeilschmidt, line 129: I guess it is better to not unwrap but throw an error.
If you are 100% sure that it won't panic, you can use expect.

Otherwise, our threads will die at some point.

Please also look at the other unwraps in this file (tests are fine, but the ones in regular functions)

services/src/datasets/external/pangaea/meta.rs
```diff
@@ -0,0 +1,624 @@
...
+        gdal::vsi::create_mem_file(memfile_name.as_str(), data)?;
+        Ok(MemfileHandle { memfile_name })
+    }
+}
+
+impl Drop for MemfileHandle {
+    fn drop(&mut self) {
+        gdal::vsi::unlink_mem_file(self.memfile_name.as_str()).unwrap();
+    }
+}
```
ChristianBeilschmidt, line 310: How do you guarantee that this isn't dropped before you access the data?
You get the metadata and then create some loading info and discard this(?).

I guess it has to be put into the stream in order to keep this alive.
But then it is no pre-load-hook but something different :thinking:

ChristianBeilschmidt (changes requested):
I guess the main thing is to explain the pre-load-hook to us :smile:

services/test-data/pangaea/pangaea_geo_none.tsv
koerberm: All files are: https://creativecommons.org/licenses/by/4.0/
I guess that is OK?

services/Cargo.toml
```diff
@@ -56,6 +56,7 @@ typetag = "0.1"
 uuid = { version = "0.8", features = ["serde", "v4", "v5"] }
 warp = "0.3"
 zip = "0.5"
+url = { version = "2.2", features = ["serde"] }
```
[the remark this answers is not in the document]
koerberm: 4e0d9bf6e6b31d3d62256d1fc2baa6f56ce0f1b2

services/src/datasets/external/pangaea/meta.rs
koerberm: 4fe6e64bb653a4451d5a92e4ef1d66e9b22952e9

services/src/datasets/external/pangaea/meta.rs
koerberm: The Memfile is stored within the hook, and the hook is part of the MetaData. I assumed the MetaData to outlive the stream. If this is not the case, we could make the hook return some result and store it along with the stream (or even store the hook itself).

operators/src/source/ogr_source.rs
```diff
@@ -496,42 +525,64 @@ where
 
         let mut dataset_options = DatasetOptions::default();
 
+        let headers = format!(
+            "HEADERS={}",
+            columns
+                .csv_headers
+                .as_ref()
+                .map_or("AUTO", CsvHeaders::option_value)
+        );
```
ChristianBeilschmidt, line 534: You could put the `HEADERS=` into your to-string method (option_value).
Then you can just say `columns.csv_headers.unwrap_or(CsvHeaders::Auto).to_string()`.

I'm just looking for means to keep these huge methods as concise as possible :sweat_smile: .

operators/src/source/ogr_source.rs
```diff
@@ -178,9 +180,11 @@ impl Default for OgrSourceTimeFormat {
 ///  - text: an array of column names containing alpha-numeric values
 ///  - rename: a. optional map of column names from data source to the name in the resulting collection
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
 pub struct OgrSourceColumnSpec {
     pub x: String,
     pub y: Option<String>,
+    pub csv_headers: Option<CsvHeaders>,
```
ChristianBeilschmidt, line 187: Maybe we should CSV-specific stuff into some _ format-specific parameters_ > csv field?
We would have for each file type one struct then, starting with CSV.

operators/src/source/ogr_source.rs
```diff
@@ -496,42 +525,64 @@ where
...
             &format!("GEOM_POSSIBLE_NAMES={}", columns.x),
+            headers.as_str(),
             "AUTODETECT_TYPE=YES",
         ];
         dataset_options.open_options = Some(open_opts);
-
         Ok(Dataset::open_ex(&dataset_info.file_name, dataset_options)?)
     }
 
+    // TODO: reliably detect CSV files or allow defining them as such in params
```
ChristianBeilschmidt, line 566: I guess `CSV:` is the only way to reliably load CSV files with OGR so I guess we could remove this TODO.

operators/src/source/ogr_source.rs
```diff
@@ -863,15 +914,29 @@ where
...
+        ) {
+            Ok(g) => g,
+            Err(Error::Gdal {
+                source: GdalError::InvalidFieldIndex { .. },
+            }) if dataset_information.default_geometry.is_some() => dataset_information
+                .default_geometry
+                .as_ref()
+                .expect("Impossible")
+                .clone()
+                .try_into()?,
```
ChristianBeilschmidt, line 937: You could try to get the right geometry from the method before and insert an Option<G> here.

Maybe this will get the method more concise. Do you have any idea if there would be a good way to get rid of the *if_some() then unwrap()* pattern? If not, we can keep it as is.

operators/src/source/ogr_source.rs
```diff
@@ -1190,9 +1256,13 @@ mod tests {
                     step: 42,
                 }),
             },
+            default_geometry: Some(TypedGeometry::MultiPoint(
```
ChristianBeilschmidt, line 1259: I guess we need one test here for using the default geometry, right?

services/Cargo.toml
```diff
@@ -22,6 +22,7 @@ pro = ["postgres", "geoengine-operators/pro", "geoengine-datatypes/pro"]
 async-trait = "0.1"
 base64 = "0.13"
 bb8-postgres = { version = "0.7", features = ["with-uuid-0_8", "with-chrono-0_4", "with-serde_json-1"], optional = true }
+bstr = "0.2"
```
ChristianBeilschmidt, line 25: Why do you have this dependency?
It seems you don't use it.

ChristianBeilschmidt (changes requested):

operators/src/source/ogr_source.rs
koerberm: 0fcb8508b434b9d310b9344204d08121dc741701

operators/src/source/ogr_source.rs
koerberm: 0fcb8508b434b9d310b9344204d08121dc741701

operators/src/source/ogr_source.rs
koerberm: 32d66d3b8da3971ea60747ad73c0c63b653e5794

operators/src/source/ogr_source.rs
koerberm: 023113f1a5d7d0dd935dc67b36f3d1ddd322ce5b

operators/src/source/ogr_source.rs
koerberm: dae2a60b1eb1c7c9fd50d3676daaececcfda0b79

.github/workflows/ci.yml
```diff
@@ -36,6 +36,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+      - name: Add UbuntuGIS ppa
+        run: sudo add-apt-repository ppa:ubuntugis/ppa
```
ChristianBeilschmidt, line 40: Please add the info about this requirement into our README.md, under Deployment or Dependencies.

ChristianBeilschmidt (changes requested):

.github/workflows/ci.yml
koerberm: 22fc0234da600237aecb0e4deedfb5bf3d476cd2

ChristianBeilschmidt (approved):

PR #339, opened 2021-09-09
make postgres context test idempotent

ChristianBeilschmidt:


ChristianBeilschmidt:
[Postgres Tests: Erstelle ad-hoc-Schema (random) für jeden Test und lösche es nach Ende des Tests.](https://trello.com/c/EBZR3fEL/311-postgres-tests-erstelle-ad-hoc-schema-random-f%C3%BCr-jeden-test-und-l%C3%B6sche-es-nach-ende-des-tests)

services/src/pro/contexts/postgres.rs
```diff
@@ -362,62 +362,125 @@ mod tests {
...
+
+        tear_down_db(pg_config, &schema).await;
+
+        // then throw errors afterwards
+        if let Err(err) = executed_fn {
+            std::panic::resume_unwind(err);
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
michaelmattig, line 451: is this really needed? the documentation says multi_threaded is the default flavor.

Settings-test.toml
```diff
@@ -1,5 +1,10 @@
 [postgres]
-config_string = "postgresql://geoengine:geoengine@localhost:5432"
+host = "localhost"
+port = 5432
+database = "geoengine"
+schema = "pg_temp" # we need the right to create new schemata for tests
```
michaelmattig, line 5: where is this schema used? In the test you create a geoengine_test_1234 schema.

services/src/pro/contexts/postgres.rs
ChristianBeilschmidt: yes, otherwise block_on doesn't work

Settings-test.toml
ChristianBeilschmidt: I wanted to specify a schema that doesn't provide any harm as default. You don't want to override your public schema on false usage.

michaelmattig (approved):

PR #343, opened 2021-09-14
Time-strings

ChristianBeilschmidt:


ChristianBeilschmidt:
[More flexible date in FilePAth](https://trello.com/c/N1b7hAyx/135-more-flexible-date-in-filepath)

operators/src/source/gdal_source.rs
```diff
@@ -96,13 +98,26 @@ pub enum GdalLoadingInfoPartIterator {
...
+#[serde(rename_all = "camelCase")]
+pub struct GdalSourceTimePlaceholder {
+    pub time_format: String,
+    pub which: WhichTime,
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub enum WhichTime {
+    TimeStart,
```
michaelmattig, line 117: just call it `start` and `end`

operators/src/source/gdal_source.rs
```diff
@@ -96,13 +98,26 @@ pub enum GdalLoadingInfoPartIterator {
     Dynamic {
         time_step_iter: TimeStepIter,
         params: GdalDatasetParameters,
-        placeholder: String,
-        time_format: String,
+        time_placeholders: HashMap<String, GdalSourceTimePlaceholder>,
```
michaelmattig, line 101: maybe name it `file_time_injection`? idk but it's more than a placeholder.

Also: do we have to check that there is at least one placeholder defined?

michaelmattig (changes requested):

operators/src/source/gdal_source.rs
ChristianBeilschmidt: The name is more confusing to the user as *placeholder* I guess.

We don't need to check that one is defined. If there is none, there is just no replacement, right?

operators/src/source/gdal_source.rs
ChristianBeilschmidt: 354160f

operators/src/source/gdal_source.rs
```diff
@@ -337,8 +351,7 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
 pub struct GdalMetaDataRegular {
     pub result_descriptor: RasterResultDescriptor,
     pub params: GdalDatasetParameters,
-    pub placeholder: String,
-    pub time_format: String,
+    pub time_placeholders: HashMap<String, GdalSourceTimePlaceholder>,
```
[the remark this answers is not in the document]
ChristianBeilschmidt: 354160f

operators/src/source/gdal_source.rs
```diff
@@ -96,13 +98,26 @@ pub enum GdalLoadingInfoPartIterator {
...
         step: TimeStep,
         max_t2: TimeInstance,
     },
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GdalSourceTimePlaceholder {
+    pub time_format: String,
+    pub which: WhichTime,
```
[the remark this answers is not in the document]
ChristianBeilschmidt: 354160f

operators/src/source/gdal_source.rs
```diff
@@ -382,25 +394,93 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
 }
 
 impl GdalDatasetParameters {
+    /// Placeholders are surrounded by `%` and are replaced by formatted time value.
```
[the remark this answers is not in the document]
ChristianBeilschmidt: I don't know, lots of template engines work this way, e.g. in TypeScript you use `${variable}`.
I hope we won't get many replacements since multiple replace calls on top of each other aren't a good solution.

06785c7

operators/src/source/gdal_source.rs
michaelmattig: > The name is more confusing to the user as placeholder I guess.

ok

> We don't need to check that one is defined. If there is none, there is just no replacement, right?

if no placeholder is defined, the time is irrelevant and the user should use static meta data instead.

ChristianBeilschmidt: So we should fail on deserialization?
But what if the user specifies one but doesn't use it in the file_path?

michaelmattig: if we want to enforce this we should either fail on deserialization or on operator initialization.

pro:
 - fail early, otherwise it might be hard to figure out why the data is not like desired

contra:
- inconvienent for the user. maybe they have only one file currently and don't know the pattern but want to define the dataset as timer series already

ChristianBeilschmidt: 3345c04

michaelmattig (approved):

PR #344, opened 2021-09-14
make it work with proj 8

jdroenner:


datatypes/src/primitives/spatial_partition.rs
```diff
@@ -83,9 +83,9 @@ impl SpatialPartition2D {
     /// Checks if a coordinate is located inside spatial partition
     pub fn contains_coordinate(&self, coordinate: &Coordinate2D) -> bool {
         coordinate.x >= self.upper_left_coordinate.x
-            && coordinate.y >= self.upper_left_coordinate.y
+            && coordinate.y <= self.upper_left_coordinate.y
             && coordinate.x < self.lower_right_coordinate.x
-            && coordinate.y < self.lower_right_coordinate.y
+            && coordinate.y > self.lower_right_coordinate.y
```
ChristianBeilschmidt, line 88: Why does the logic change when changing into a new `proj` version?
Was this a bug? If so, we should write a test for that.

operators/src/adapters/raster_subquery_adapter.rs
```diff
@@ -611,6 +611,17 @@ where
...
+        let proj = CoordinateProjector::from_known_srs(self.out_srs, self.in_srs)?;
+        let valid_bounds_proj =
+            CoordinateProjector::from_known_srs(SpatialReference::epsg_4326(), self.out_srs)?;
+
+        let valid_bounds = self
+            .in_srs
+            .area_of_use::<SpatialPartition2D>()?
+            .reproject(&valid_bounds_proj)?;
+
+        dbg!(&valid_bounds);
```
ChristianBeilschmidt, line 623:
```suggestion
```

operators/src/adapters/raster_subquery_adapter.rs
```diff
@@ -611,6 +611,17 @@ where
...
             EmptyGrid::new(tile_info.tile_size_in_pixels, self.no_data_and_fill_value);
 
+        let proj = CoordinateProjector::from_known_srs(self.out_srs, self.in_srs)?;
+        let valid_bounds_proj =
+            CoordinateProjector::from_known_srs(SpatialReference::epsg_4326(), self.out_srs)?;
+
+        let valid_bounds = self
+            .in_srs
+            .area_of_use::<SpatialPartition2D>()?
+            .reproject(&valid_bounds_proj)?;
```
ChristianBeilschmidt, line 621: I'm not 100% sure why 4326 is used every time.
I guess this could be a separate function and have a documentation of what this is.

operators/src/adapters/raster_subquery_adapter.rs
```diff
@@ -619,9 +630,9 @@ where
                     .tile_geo_transform()
                     .grid_idx_to_upper_left_coordinate_2d(i)
             })
+            .filter(|c| dbg!(valid_bounds.contains_coordinate(c)))
```
ChristianBeilschmidt, line 633:
```suggestion
            .filter(|c| valid_bounds.contains_coordinate(c))
```

Maybe also add a line why this is done. I guess it has to do with the set of points that determine the other qrect.

ChristianBeilschmidt (changes requested):

operators/src/adapters/raster_subquery_adapter.rs
michaelmattig: the is a `area_of_use_projected` method which you can use instead

operators/src/adapters/raster_subquery_adapter.rs
michaelmattig: as we have a raster here which is aligned to the axes, can't we more efficiently filter the coordinates by intersecting the whole raster tile with the area of use and only check the pixels in the intersection?

PR #345, opened 2021-09-15
Zoom_far

michaelmattig:


michaelmattig:
[WMS gibt Fehler aus beim reinzoomen](https://trello.com/c/WThUeXDV/157-wms-gibt-fehler-aus-beim-reinzoomen)

services/src/handlers/wms.rs
```diff
@@ -612,4 +612,52 @@ mod tests {
...
+        let res = warp::test::request()
+            .method("GET")
+            .path(&format!(
+                "/wms?{}",
+                serde_urlencoded::to_string(params).unwrap()
+            ))
+            .reply(&wms_handler(ctx))
+            .await;
+
+        assert_eq!(res.status(), 200);
```
ChristianBeilschmidt, line 661: Don't we have to compare against specific pixels or an image result?

Vice-versa: If we would change it back from f64 to f32, would this test fail?

ChristianBeilschmidt (changes requested):

services/src/handlers/wms.rs
michaelmattig: actually it doesn't fail with f32. My approach was to zoom in in the UI until an error occured. This was not the InvalidGridBounds error in the card but a blit error because of no intersection. I fixed that one. I tried the request from the card and it worked. So it seems to have accidentally been fixed before.

I can compare against pixels by "visual validation", if that has any value at that zoom level... Or I can just remove the test and we close the card because of "cannot reproduce"

ChristianBeilschmidt (approved):

PR #351, opened 2021-09-20
add Nature40 time metadata

2younis:


services/src/datasets/external/nature40.rs
```diff
@@ -129,34 +131,54 @@ impl DatasetProvider for Nature40DataProvider {
...
-                        );
+                let mut time_indexes = 1;
+                let mut time_layer = false;
+
+                if !time_positions.is_empty() {
+                    time_indexes = time_positions.len();
+                    time_layer = true;
+                }
+
+                for time_index in 0..time_indexes {
```
michaelmattig, line 144: - different time positions should not be separate datasets
  - we want to use one dataset which changes over time, i.e. querying the dataset with a different time should yield a different result

services/src/datasets/external/nature40.rs
```diff
@@ -254,6 +276,31 @@ impl Nature40DataProvider {
         }
     }
 
+    async fn get_time_positions(
```
michaelmattig, line 279: - don't read the wcs description file twice but parse labels and positions simulatenously

services/src/datasets/external/nature40.rs
```diff
@@ -332,8 +416,17 @@ impl MetaDataProvider<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectan
...
+
+        if let Some(time_index) = time_index_option {
+            let time_position = &time_positions[time_index];
+            let date_time = NaiveDateTime::parse_from_str(time_position, "%Y-%m-%dT%H:%M")?;
+            time = Some(TimeInterval::new_instant(TimeInstance::from(date_time))?);
+        }
+
         Ok(Box::new(GdalMetaDataStatic {
-            time: None,
+            time,
```
michaelmattig, line 429: - MetaDataProvider shouldn't return a *static* dataset
  - it should be a new type of GdalMetaData that stores the individual time positions
  - in the `loading_info` method it should use the `time_interval` of the `query` parameter to produce the correct loading info
    - imho this should mean gathering all time positions of the WCS dataset that are contained in the query `time_interval` and returning them as `GdalLoadingInfo`
	- the time position has to be injected into the WCS query in the `file_path` of the loading info
  - there are different ways to implement such a provider
    - a) like `GdalMetaDataStatic`, but instead of having a single config of the `GdalDatasetParameters` we would have a separate config for each time step
	- b) like `GdalMetaRegular`, but instead of specifying the time step between to rasters we would specify the valid time positions as an array

michaelmattig (changes requested):
- please check your code with `cargo clippy --all-features --all-targets` for code lints
- please use rustfmt to format your code
  - either with `cargo fmt` or better: use fomat on save in your IDE
- I enabled the CI for your PR so you can see if everything works

services/src/datasets/external/nature40.rs
2younis: DatasetProvider for Nature40 restored to original.

services/src/datasets/external/nature40.rs
2younis: A new GdalMetaData type for fixed time positions has been created.

services/src/datasets/external/nature40.rs
```diff
@@ -284,6 +289,31 @@ impl Nature40DataProvider {
         }
         Ok(txt)
     }
+
+    fn parse_time_positions(dataset: &gdal::Dataset) -> Result<Vec<String>> {
+        let mut reader = Reader::from_file(&dataset.description()?)?;
```
michaelmattig, line 294: can you please try to avoid reading the dataset description file twice? You should be able to load it once and use it for getting the labels *and* the time steps

michaelmattig (changes requested):

operators/src/source/gdal_source.rs
```diff
@@ -370,6 +370,84 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+    pub params: GdalDatasetParameters,
+    pub result_descriptor: RasterResultDescriptor,
+}
+
+#[async_trait]
+impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
+    for GdalMetadataFixedTimes
+{
+    async fn loading_info(&self, query: RasterQueryRectangle) -> Result<GdalLoadingInfo> {
+        let valid_duration = TimeInterval::new(
```
michaelmattig, line 388: you assume that the time intervals are sorted. We have to ensure that they are. Ideally we would check this on creation of the struct (in code or while deserializing). But maybe check it here for now and add a todo to check it earlier and only once

operators/src/source/gdal_source.rs
```diff
@@ -370,6 +370,84 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
     }
 }
 
+#[derive(PartialEq, Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct GdalMetadataFixedTimes {
+    pub time_steps: Vec<TimeInterval>,
+    pub minimum_step: TimeStep,
```
michaelmattig, line 377: what do you need the minimum step for?

operators/src/source/gdal_source.rs
```diff
@@ -370,6 +370,84 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+            self.time_steps[0].start(),
+            self.time_steps[self.time_steps.len() - 1].start(),
+        )?;
+
+        let parts = match valid_duration.intersect(&query.time_interval) {
+            Some(time_interval) => {
+                let mut loading_info_parts: Vec<GdalLoadingInfoPart> = Vec::new();
+
+                for (index, time_position) in self.time_steps.iter().enumerate() {
+                    if time_interval.contains(time_position) {
```
michaelmattig, line 398: I think it should be `intersects` here. Please add a test case where an element of `time_steps` overlaps the query time interval, e.g. starting before the `query.time_interval.start` and ending after the start.

operators/src/source/gdal_source.rs
```diff
@@ -370,6 +370,84 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+        let valid_duration = TimeInterval::new(
+            self.time_steps[0].start(),
+            self.time_steps[self.time_steps.len() - 1].start(),
+        )?;
+
+        let parts = match valid_duration.intersect(&query.time_interval) {
+            Some(time_interval) => {
+                let mut loading_info_parts: Vec<GdalLoadingInfoPart> = Vec::new();
+
+                for (index, time_position) in self.time_steps.iter().enumerate() {
```
michaelmattig, line 397: the `time_position` is now an interval, so maybe call it `loading_info_validity`? Maybe you have a better idea for a name

operators/src/source/gdal_source.rs
```diff
@@ -370,6 +370,84 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+                        loading_info_parts.push(
+                            self.params
+                                .replace_time_placeholders(&self.time_placeholders, *time_position)
+                                .map(|loading_info_part_params| GdalLoadingInfoPart {
+                                    time: *time_position,
+                                    params: loading_info_part_params,
+                                })?,
+                        );
+
+                        if index < self.time_steps.len() - 1
```
michaelmattig, line 408: There can also be gaps before the first element in `time_steps` and after the last one. Basically the whole `query.time_interval` has to be covered by the sum off all the `parts`.

operators/src/source/gdal_source.rs
```diff
@@ -370,6 +370,84 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+                            );
+                        }
+                    }
+                }
+
+                loading_info_parts
+            }
+            .into_iter(),
+
+            None => vec![].into_iter(),
```
michaelmattig, line 432: the result shouldn't be empty. It should be one single loading info that's valid for the entire `query.time_interval` and should produce nodata tiles. Just use the `replace_time_placeholders` with the `query.time_interval` here. When the gdal source tries to load the data it will find nothing and produce no data.

operators/src/source/gdal_source.rs
```diff
@@ -1302,6 +1380,99 @@ mod tests {
...
+                    time_interval: TimeInterval::new_unchecked(4, 12),
+                    spatial_resolution: SpatialResolution::one(),
+                })
+                .await
+                .unwrap()
+                .info
+                .map(|p| p.unwrap().params.file_path.to_str().unwrap().to_owned())
+                .collect::<Vec<_>>(),
+            &[
+                "/foo/bar_step_005000000.tiff",
```
michaelmattig, line 1451: Why are the gaps not filled here?

services/src/datasets/external/nature40.rs
```diff
@@ -332,17 +351,56 @@ impl MetaDataProvider<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectan
...
+                .map(TimeInstance::from)
+                .map(|time_istance| TimeInterval::new_instant(time_istance).unwrap())
+                .collect::<Vec<_>>();
+
+            Ok(Box::new(GdalMetadataFixedTimes {
+                time_steps,
+                params,
+                result_descriptor,
+                time_placeholders: hashmap! {
+                    "%TIME%".to_string() => GdalSourceTimePlaceholder {
```
michaelmattig, line 393: you have to actually use the `%TIME%` placeholder somewhere. I guess in the `time` parameter of the WCS url in the `GetCoverage` request

michaelmattig (changes requested):
Please have look at my comments and also merge the master into your branches, because 

> This branch is 14 commits ahead, 100 commits behind geo-engine:master.

operators/src/source/gdal_source.rs
2younis: I couldn't come up with a better name for now, perhaps we can change it later.

operators/src/source/gdal_source.rs
```diff
@@ -372,6 +374,155 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+#[serde(rename_all = "camelCase")]
+pub struct GdalMetadataFixedTimes {
+    pub time_steps: Vec<TimeInterval>,
+    pub params: GdalDatasetParameters,
+    pub result_descriptor: RasterResultDescriptor,
+    pub time_placeholders: HashMap<String, GdalSourceTimePlaceholder>,
+}
+
+impl GdalMetadataFixedTimes {
+    #[allow(dead_code)]
```
michaelmattig, line 387: remove

operators/src/source/gdal_source.rs
```diff
@@ -372,6 +374,155 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
+    for GdalMetadataFixedTimes
+{
+    async fn loading_info(&self, query: RasterQueryRectangle) -> Result<GdalLoadingInfo> {
+        let minimum_step = TimeStep {
+            granularity: TimeGranularity::Millis,
+            step: 1,
+        };
+
+        let sorted_time_steps = if self
```
michaelmattig, line 416: you already sort them in the `new` method why sort again?

operators/src/source/gdal_source.rs
```diff
@@ -372,6 +374,155 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+        };
+
+        let mut time_steps_no_gaps = sorted_time_steps.clone();
+
+        for (index, time) in sorted_time_steps.iter().enumerate() {
+            if index < sorted_time_steps.len() - 1 {
+                if time.is_instant() {
+                    time_steps_no_gaps.push(TimeInterval::new(
+                        (time.end() + minimum_step)?,
+                        sorted_time_steps[index + 1].start(),
```
michaelmattig, line 435: could this be out of bounds?

operators/src/source/gdal_source.rs
```diff
@@ -372,6 +374,155 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+            .all(|w| w[0].end() <= w[1].start())
+        {
+            self.time_steps.clone()
+        } else {
+            let mut sorted_time_steps = self.time_steps.clone();
+            sorted_time_steps.sort_by(|a, b| a.end().partial_cmp(&b.start()).unwrap());
+            sorted_time_steps
+        };
+
+        let mut time_steps_no_gaps = sorted_time_steps.clone();
```
michaelmattig, line 428: you should fill the gaps in the `new` method instead of every time `loading_info` is called.

also: how do you distinguish between gaps and data? In a gap, we should directly create NO-DATA-tiles instead of requesting the server with the a time where we know there is no data.

operators/src/source/gdal_source.rs
```diff
@@ -372,6 +374,155 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+                } else {
+                    time_steps_no_gaps.push(TimeInterval::new(
+                        time.end(),
+                        sorted_time_steps[index + 1].start(),
+                    )?);
+                }
+            }
+        }
+
+        time_steps_no_gaps.sort_by(|a, b| a.start().partial_cmp(&b.start()).unwrap());
```
michaelmattig, line 446: is it cheaper to add the gaps to the end of the vec and then sort, or to insert them in the right place right away?

operators/src/source/gdal_source.rs
```diff
@@ -372,6 +374,155 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+                                params: loading_info_part_params,
+                            })?,
+                    );
+                }
+
+                loading_info_parts
+            }
+            .into_iter(),
+
+            None => vec![self
```
michaelmattig, line 500: if there is no intersection, the source should just create no-data tiles instead of trying to load data.

operators/src/source/gdal_source.rs
```diff
@@ -372,6 +374,155 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+                        self.params
+                            .replace_time_placeholders(&self.time_placeholders, gap)
+                            .map(|loading_info_part_params| GdalLoadingInfoPart {
+                                time: gap,
+                                params: loading_info_part_params,
+                            })?,
+                    );
+                }
+                for time_position in &time_steps_no_gaps {
+                    if let Some(step_interval) = time_interval.intersect(time_position) {
```
michaelmattig, line 471: do you really have to check for intersction every time?

michaelmattig (changes requested):
I think we are still missing a crucial piece here. In the gaps between valid TimeIntervals, the GdalSource should produce no data tiles without trying to load data. My suggestions would be to make the produced `GdalLoadingInfoParts` contiguous (no gaps), but indicate whether there actually is data or not (basically making the `GdalDatasetParameters` an Option) and adjust the `tile_stream` method accordingly so that it can either produce no-data directly, or load it using Gdal. It would be great if you could take a shot at that and present it in one of the next dev-meetings. I am sure Johannes will have input on that.

Other than that also have a look at my other comments.

operators/src/source/gdal_source.rs
2younis: I couldn't figure out how to fix the warning so I changed it to `#[allow(unused)]`

operators/src/source/gdal_source.rs
2younis: Fixed, moved all relevant code to the `new` method

operators/src/source/gdal_source.rs
2younis: I think the if condition `if index < sorted_time_steps.len() - 1` should prevent this

operators/src/source/gdal_source.rs
2younis: I added a new field in `GdalMetadataFixedTimes` to store the time steps with no gaps and made all fields private, so now everything can be done in the `new` method.

operators/src/source/gdal_source.rs
2younis: This is to create the intersect `TimeInterval`, which is then compared with the original `time_steps` to check if it should create valid or no-data tile.

datatypes/src/primitives/time_interval.rs
```diff
@@ -212,6 +215,8 @@ impl TimeInterval {
         other == self
             || value_in_range(self.start, other.start, other.end)
             || value_in_range(other.start, self.start, self.end)
+            || { value_in_range_inclusive(self.start, other.start, other.end) && self.is_instant() }
```
michaelmattig, line 218: - why do you use curly braces? please use regular braces here
- I don't think it's correct to use `inclusive here`. The end of a time interval is exclusive
- can you please add tests that would have failed before your change and do work now

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -54,6 +53,134 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
         Box::new(self.clone())
     }
 }
+#[derive(PartialEq, Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct GdalMetadataFixedTimes {
+    time_steps: Vec<TimeInterval>,
+    time_steps_no_gaps: Vec<TimeInterval>,
```
michaelmattig, line 60: why do you store both, time steps with and without gaps? Especially since you derive Deserialize it's inconvenient having to specify the time steps twice e.g. in a JSON file. I think it would be better to just have a time_steps field and require the timesteps to have no gaps and be sorted. Add a todo to check this during serialization

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -54,6 +53,134 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+    }
+}
+
+#[async_trait]
+impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
+    for GdalMetadataFixedTimes
+{
+    async fn loading_info(&self, query: RasterQueryRectangle) -> Result<GdalLoadingInfo> {
+        let valid_duration = TimeInterval::new(
+            self.time_steps_no_gaps[0].start(),
```
michaelmattig, line 130: this panics if the time_steps_no_gaps is empty. Please return an error instead.

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -54,6 +53,134 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+        })
+    }
+
+    fn create_temporal_slice(
+        &self,
+        step_interval: TimeInterval,
+        empty: bool,
+    ) -> Result<GdalLoadingInfoTemporalSlice> {
+        self.params
+            .replace_time_placeholders(&self.time_placeholders, step_interval)
```
michaelmattig, line 107: you only have to replace the timeholders if empty is false, so please check for that first and replace afterwards

datatypes/src/primitives/time_interval.rs
2younis: The inclusive check is only for `TimeInstance`, since `value_in_range` returns `false` when the other `TimeInstance` is at the end of the `TimeInterval`.

operators/src/source/gdal_source.rs
2younis: This method is only used externally for instantiating the struct and not used within `loading_info.rs`, thus it is unused.

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -54,6 +53,129 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+    }
+}
+
+#[async_trait]
+impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
+    for GdalMetadataFixedTimes
+{
+    async fn loading_info(&self, query: RasterQueryRectangle) -> Result<GdalLoadingInfo> {
+        if self.time_steps.is_empty() {
+            return Err(Error::EmptyTimeSteps);
```
michaelmattig, line 111: return an empty time slice here instead

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -54,6 +53,129 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+                            loading_info_parts
+                                .push(self.create_temporal_slice(gap_step_interval, true)?);
+                        }
+                    };
+
+                    if time_position.start() > time_interval.end() {
+                        break;
+                    }
+
+                    if let Some(step_interval) = time_interval.intersect_inclusive(time_position) {
```
michaelmattig, line 147: why does it have to be `inclusive` here? If it is really necessary, then please move this method into `impl GdalMetadataFixedTimes` we don't want this used anywhere else in the code because it violates our time semantics

2younis: It is necessary to have `inclusive` here to include the last `time_step` when it is a `TimeInstance` and at the end of the `time_interval`. For example if the `time_interval` is [10, 20) and the last `time_step` within `time_interval` is [20, 20), then it is necessary to  have `inclusive` to include the last `time_step`.

michaelmattig: the `end` of a time interval doesn't belong to it. So the instance [20, 20) in your example shouldn't be included right?. It's like if I am querying January as [1.1.2000, 1.2.2000) I wouldn't want an Instant on the first of February returned.

2younis: Okay, I will change it back to `intersect` without `inclusive`. I was under the assumption that all TimeInstances in the query interval have to be returned.

PR #355, opened 2021-09-23
make postgres backend work again; use in memory db for datasets

michaelmattig:


michaelmattig:
[Persistenter Workflow-Storage](https://trello.com/c/jelQaSn5/400-persistenter-workflow-storage)

services/src/pro/contexts/postgres.rs
```diff
@@ -38,6 +46,8 @@ where
     user_db: Db<PostgresUserDb<Tls>>,
     project_db: Db<PostgresProjectDb<Tls>>,
     workflow_registry: Db<PostgresWorkflowRegistry<Tls>>,
+    dataset_db: Db<ProHashMapDatasetDb>, // TODO: implement postgres version
```
ChristianBeilschmidt, line 49: This means you cannot restore uploaded datasets?
I'm not sure if we don't need this :thinking:

services/src/pro/contexts/postgres.rs
```diff
@@ -58,6 +68,34 @@ where
...
+        add_datasets_from_directory(&mut dataset_db, dataset_defs_path).await;
+        add_providers_from_directory(&mut dataset_db, provider_defs_path.clone()).await;
+        add_providers_from_directory(&mut dataset_db, provider_defs_path.join("pro")).await;
+
+        Ok(Self {
+            user_db: Arc::new(RwLock::new(PostgresUserDb::new(pool.clone()))),
+            project_db: Arc::new(RwLock::new(PostgresProjectDb::new(pool.clone()))),
+            workflow_registry: Arc::new(RwLock::new(PostgresWorkflowRegistry::new(pool.clone()))),
+            dataset_db: Arc::new(RwLock::new(dataset_db)),
+            thread_pool: Default::default(),
```
ChristianBeilschmidt, line 98: Could we write a test that creates a context, stores something, drops the context, creates a new context, and retrieves the workflow?
Then, could see if it works.

ChristianBeilschmidt (changes requested):

services/src/pro/contexts/postgres.rs
michaelmattig: yes, the task read minimum viable to store workflows and that works. I guesstimate it's 20x more work to get persist the datasets also

services/src/pro/contexts/postgres.rs
michaelmattig: We already test that we can write to and read from the database. I am not sure why we want to test whether the database is able to persistently stored data.

ChristianBeilschmidt: Well, we distribute things into different handlers user data, workflows, etc.
This way we see what we can retrieve and what not. For instance, currently, it should fail to upload a dataset, shut down the Postgres context and restore it.

services/src/pro/contexts/postgres.rs
ChristianBeilschmidt: Ok, so then we should add this as a new task to the sprint log.

michaelmattig: https://trello.com/c/cTGGU4bs

services/src/pro/contexts/postgres.rs
michaelmattig: f19d3a9f13266c3023348d6628595a668f06b2a2

services/src/datasets/upload.rs
```diff
@@ -39,11 +41,12 @@ impl Upload {
...
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[cfg_attr(feature = "postgres", derive(ToSql, FromSql))]
 pub struct FileUpload {
     pub id: FileId,
     pub name: String,
-    pub byte_size: usize,
+    pub byte_size: i64,
```
ChristianBeilschmidt, line 49: Hm, is this a good strategy to change this type because of PostgreSQL?

services/src/pro/datasets/postgres.rs
```diff
@@ -5,114 +5,599 @@ use crate::datasets::storage::{
...
+        let meta_data: Result<GdalMetaDataRegular, _> = serde_json::from_value(row.get(0));
+
+        if let Ok(meta_data) = meta_data {
+            return Ok(Box::new(meta_data));
+        }
+
+        // ... or static.
+        let meta_data: GdalMetaDataStatic = serde_json::from_value(row.get(0))?;
+        dbg!(&meta_data);
+        Ok(Box::new(meta_data))
```
ChristianBeilschmidt, line 409: Isn't this an enum that we could deserialize?

Moreover, please remove the `dbg!`

ChristianBeilschmidt (changes requested):

services/src/datasets/upload.rs
michaelmattig: not the best bust most pragmatic. I could also implement the To/FromSQL manually or duplicate the type but with usize/isize, derive the traits and convert both into each other.

services/src/pro/datasets/postgres.rs
michaelmattig: 9b7d107295722dc5fc39d86cedaae2c509438a0a

services/src/datasets/upload.rs
ChristianBeilschmidt: I would say only an unsigned value makes sense, so either u64 or size.
You have more experience if implementing `To/FromSQL` is easier then using the DAO object.

michaelmattig: f76c432022b0dfab1387d83f20c5060c306187a2

I changed it to u64 but I couldn't get the >= 0 constraint to work on postgres side

ChristianBeilschmidt (approved):

PR #364, opened 2021-10-07
write COG instead of plain GeoTiff by default

ChristianBeilschmidt:


operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -205,6 +218,16 @@ fn gdal_writer<T: Pixel + GdalType>(
...
     }
 
+    if as_cog {
+        // Override file with COG driver.
+
+        // Since COGs are written "overviews first, data second", we cannot generate a GeoTiff (where we first have to
+        // write the data in order to generate overviews) that fulfills this property. So, we have to do it as a
+        // separate step.
+
+        dataset.create_copy(&Driver::get("COG")?, output_file_path, &options)?;
```
jdroenner, line 228: i don't think it is a good idea to write in the same file. what if the original file is not entirely in memory? then the file you read is overwriten while you still need data.

services/src/handlers/workflows.rs
```diff
@@ -236,6 +236,14 @@ struct RasterDatasetFromWorkflow {
     name: String,
     description: Option<String>,
     query: RasterQueryRectangle,
+    #[serde(default = "default_as_cog")]
+    as_cog: bool,
+}
+
+/// By default, we set [`RasterDatasetFromWorkflow::as_cog`] to true to produce cloud-optmized `GeoTiff`s.
+#[inline]
```
jdroenner, line 244: why inline this? the compiler will  optimize it anyway

operators/src/util/raster_stream_to_geotiff.rs
ChristianBeilschmidt: 2ad0dc1

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -112,6 +112,10 @@ fn gdal_writer<T: Pixel + GdalType>(
     as_cog: bool,
 ) -> Result<()> {
     const INTERMEDIATE_FILE_SUFFIX: &str = "GEO-ENGINE-TMP";
+    const COG_BLOCK_SIZE: &str = "512";
+    const COMPRESSION_FORMAT: &str = "LZW";
+    const COMPRESSION_LEVEL: &str = "9"; // maximum compression
+    const COMPRESSION_NUM_THREADS: &str = "ALL_CPUS"; // full power for compression
```
jdroenner, line 118: should this really be a default? it will use all threads. if we handle 100 compression tasks this could cause problems

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -132,22 +136,34 @@ fn gdal_writer<T: Pixel + GdalType>(
...
             key: "COMPRESS",
-            value: "LZW",
+            value: COMPRESSION_FORMAT,
         },
         RasterCreationOption {
             key: "TILED",
             value: "YES",
         },
+        RasterCreationOption {
+            key: "ZLEVEL",
```
jdroenner, line 146: should be "LEVEL" https://gdal.org/drivers/raster/cog.html?highlight=level

jdroenner (changes requested):

operators/src/util/raster_stream_to_geotiff.rs
ChristianBeilschmidt: It is level for COG but ZLEVEL for GeoTiff

operators/src/util/raster_stream_to_geotiff.rs
ChristianBeilschmidt: We could either go back to one thread or we use a config setting for this.
What is your favorite?

jdroenner: config setting and a safe default like 8 i guess...

ChristianBeilschmidt: 9a953a1 

I added the config value.

Moreover, I refactored the functions that were too long either in their body or their signature to get rid of the Clippy warnings.

services/src/handlers/wcs.rs
```diff
@@ -349,8 +351,14 @@ async fn get_coverage<C: Context>(
...
             query_rect,
             query_ctx,
-            no_data_value,
-            request_spatial_ref,
+            GdalGeoTiffDatasetMetadata {
+                no_data_value,
+                spatial_reference: request_spatial_ref,
+            },
+            GdalGeoTiffOptions {
+                compression_num_threads: get_config_element::<crate::util::config::Gdal>()?.compression_num_threads,
```
jdroenner, line 359: i think we must be careful with the config elements. We changed most of the config lookup to run only once at server start. There we set the context variables and they are "nerver" read again. In this case there might be no harm in reading it in the handler BUT every call will read the config file and i think that is not a nice solution.

services/src/handlers/workflows.rs
```diff
@@ -358,8 +369,14 @@ async fn dataset_from_workflow_handler<C: Context>(
...
             query_rect,
             query_ctx,
-            no_data_value,
-            request_spatial_ref,
+            GdalGeoTiffDatasetMetadata {
+                no_data_value,
+                spatial_reference: request_spatial_ref,
+            },
+            GdalGeoTiffOptions {
+                compression_num_threads: get_config_element::<crate::util::config::Gdal>()?.compression_num_threads,
```
jdroenner, line 377: same here

services/src/handlers/workflows.rs
```diff
@@ -1003,8 +1020,16 @@ mod tests {
...
             query_rect,
             query_ctx,
-            Some(0.),
-            SpatialReference::epsg_4326(),
+            GdalGeoTiffDatasetMetadata {
+                no_data_value: Some(0.),
+                spatial_reference: SpatialReference::epsg_4326(),
+            },
+            GdalGeoTiffOptions {
+                compression_num_threads: get_config_element::<crate::util::config::Gdal>()
```
jdroenner, line 1028: and here

services/src/handlers/wcs.rs
ChristianBeilschmidt: Do you have an example for that?

jdroenner: we load the config in the server startup and store the values in the context: https://github.com/geo-engine/geoengine/blob/8a8e873ed2824eb608c52fdc1af8ac9ec913b1eb/services/src/pro/server.rs#L115

ChristianBeilschmidt: siehe Diskussion in Slack

services/src/handlers/workflows.rs
ChristianBeilschmidt: siehe Diskussion in Slack

services/src/handlers/workflows.rs
ChristianBeilschmidt: siehe Diskussion in Slack

jdroenner (approved):

jdroenner:
bors merge

PR #365, opened 2021-10-07
serde modules for floats with NaNs as strings

ChristianBeilschmidt:


ChristianBeilschmidt:
[NAN in Dataset-File als NO_DATA angeben lassen](https://trello.com/c/KZPVNvRA/411-nan-in-dataset-file-als-nodata-angeben-lassen)

operators/src/processing/expression.rs
```diff
@@ -38,39 +37,11 @@ use std::marker::PhantomData;
 pub struct ExpressionParams {
     pub expression: String,
     pub output_type: RasterDataType,
-    #[serde(deserialize_with = "parse_no_data")]
-    #[serde(serialize_with = "write_no_data")]
+    #[serde(with = "float_with_nan")]
```
michaelmattig, line 40: what was wrong with the old approach? Did it ever work?

operators/src/util/input/float_with_nan_serde.rs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            if v.to_lowercase() == "nan" {
+                Ok(f64::NAN)
+            } else {
+                Err(de::Error::invalid_value(de::Unexpected::Str(v), &self))
```
michaelmattig, line 47: what if the string is parsable as float?

operators/src/processing/expression.rs
ChristianBeilschmidt: The old approach used the JSON parser internally. But Serde isn't all JSON. But other than that it should work the same.

michaelmattig (approved):

PR #369, opened 2021-10-11
Box Plots

koerberm:
Implemented box plots for vector and raster data. For many data points (currently > 10000) the operator uses the P^2 algorithm to estimate quantiles. For details see:
R. Jain and I. Chlamtac, The P^2 algorithm for dynamic calculation of quantiles and
histograms without storing observations, Communications of the ACM,
Volume 28 (October), Number 10, 1985, p. 1076-1085.
(https://www.cse.wustl.edu/~jain/papers/ftp/psqr.pdf)

### DIFF
--- a/operators/src/util/statistics.rs
+++ b/operators/src/util/statistics.rs
@@ -1,0 +1,444 @@
+use num_traits::AsPrimitive;
+use snafu::Snafu;
+use std::marker::PhantomData;
+
+/// Enum for any errors encountered while creating statistics.
+#[derive(Debug, Snafu)]
+pub enum StatisticsError {
+    #[snafu(display("Error initializing statistics. Reason: {}", reason))]
+    Initialization { reason: String },
+    #[snafu(display("Cannot compute statistics on empty sample."))]
+    Empty,
+}
+
+/// Single quantile estimation with the P^2 algorithm
+///
+/// The P^2 algorithm estimates a quantile dynamically without storing samples. Instead of
+/// storing the whole sample cumulative distribution, only five points (markers) are stored. The heights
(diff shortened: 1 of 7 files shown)


operators/src/util/statistics.rs
```diff
@@ -0,0 +1,723 @@
...
+                }
+            }
+        }
+
+        // We require at least 5 valid initial samples
+        if sample_count < 5 {
+            return Err(StatisticsError::Initialization {
+                reason: "Insufficient valid samples.".to_owned(),
+            });
+        }
```
ChristianBeilschmidt, line 104: Maybe check this early on and fail early.

operators/src/util/statistics.rs
```diff
@@ -0,0 +1,723 @@
...
+            quantile,
+            positions,
+            markers,
+            desired,
+            increment,
+            sample_count: sample_count as u64,
+            _phantom: PhantomData {},
+        };
+
+        // Add remaining values
```
ChristianBeilschmidt, line 118: Are the remaining ones the initial ones?

operators/src/util/statistics.rs
```diff
@@ -0,0 +1,723 @@
...
+                break;
+            }
+        }
+
+        // We require at least 5 valid initial samples
+        if sample_count <= bucket_count {
+            return Err(StatisticsError::Initialization {
+                reason: "Insufficient valid samples.".to_owned(),
+            });
+        }
```
ChristianBeilschmidt, line 319: Also here, just fail early. You could also use the ensure!-macro from snafu for that.

ChristianBeilschmidt (changes requested):

PR #389, opened 2021-10-25
Permissions

michaelmattig:


services/src/datasets/in_memory.rs
```diff
@@ -231,6 +229,30 @@ impl DatasetProvider for HashMapDatasetDb {
...
+                .datasets
+                .iter()
+                .find(|d| d.id == *dataset)
+                .map(|d| ProvenanceOutput {
+                    dataset: d.id.clone(),
+                    provenance: d.provenance.clone(),
+                })
+                .ok_or(error::Error::UnknownDatasetId),
+            DatasetId::External(id) => {
+                self.dataset_provider(&SimpleSession::mock(), id.provider_id) // TODO: get correct session into dataset provider
```
ChristianBeilschmidt, line 249: Why don't you use `_session`?

services/src/datasets/listing.rs
```diff
@@ -72,24 +71,50 @@ pub enum OrderBy {
...
+    + MetaDataProvider<MockDatasetDataSourceLoadingInfo, VectorResultDescriptor, VectorQueryRectangle>
+    + MetaDataProvider<OgrSourceDataset, VectorResultDescriptor, VectorQueryRectangle>
+    + MetaDataProvider<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
+{
+    // TODO: authorization, filter, paging
+    async fn list(&self, options: Validated<DatasetListOptions>) -> Result<Vec<DatasetListing>>;
+
+    // TODO: authorization
+    // TODOis this method useful?
+    async fn load(&self, dataset: &DatasetId) -> Result<Dataset>;
```
ChristianBeilschmidt, line 104: Why isn't this method useful? If you wouldn't use it, you wouldn't have defined it?

services/src/datasets/listing.rs
```diff
@@ -72,24 +71,50 @@ pub enum OrderBy {
...
-        // session: &S, // TODO: authorization
-        dataset: &DatasetId,
-    ) -> Result<Dataset>;
+    async fn load(&self, session: &S, dataset: &DatasetId) -> Result<Dataset>;
+
+    async fn provenance(&self, session: &S, dataset: &DatasetId) -> Result<ProvenanceOutput>;
+}
+
+#[async_trait]
+pub trait ExternalDatasetProvider: Send
```
ChristianBeilschmidt, line 93: Maybe we could add some notes about that we would need to specify authorization as a parameter independent of the concrete context.

services/src/handlers/datasets.rs
```diff
@@ -88,7 +88,7 @@ async fn list_external_datasets_handler<C: Context>(
         .await
         .dataset_provider(&session, provider.into_inner())
         .await?
-        .list(options)
+        .list(options) // TODO: authorization
```
ChristianBeilschmidt, line 91: If you grab your dataset provider by providing your session, isn't this part of the inner list to use it rather than here?

services/src/pro/contexts/postgres.rs
```diff
@@ -297,9 +310,57 @@ where
...
+                        );
+
+                        CREATE VIEW user_permitted_datasets AS
+                            SELECT 
+                                u.id as user_id,
+                                p.dataset_id,
+                                p.permission
+                            FROM 
+                                users u JOIN user_roles r ON (u.id = r.user_id)
+                                    JOIN dataset_permissions p ON (r.role_id = p.role_id);
```
ChristianBeilschmidt, line 357: if the user id is in roles, couldn't you save joins?

services/src/pro/contexts/postgres.rs
```diff
@@ -297,9 +310,57 @@ where
...
+                            role_id UUID REFERENCES roles(id) ON DELETE CASCADE NOT NULL,
+                            PRIMARY KEY (user_id, role_id)
+                        );
+
+                        -- system user role
+                        INSERT INTO user_roles 
+                            (user_id, role_id)
+                        VALUES 
+                            ('d5328854-6190-4af9-ad69-4e74b0961ac9', 
+                            'd5328854-6190-4af9-ad69-4e74b0961ac9');
```
ChristianBeilschmidt, line 339: couldn't we define this as constants in the code rather than in text?

services/src/pro/datasets/add_from_directory.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        db.add_dataset_permission(
+            system_session,
+            DatasetPermission {
+                role: Role::anonymous_role_id(),
+                dataset: dataset_id,
+                permission: Permission::Read,
+            },
+        )
+        .await?;
```
ChristianBeilschmidt, line 60: So there is a generic user role id? This sounds strange. Couldn't we make a public role that only requires a single assignment? Maybe you have to explain your rationale behind this. Or maybe make some doc-comments.

services/src/pro/datasets/in_memory.rs
```diff
@@ -181,24 +201,29 @@ impl DatasetStore<UserSession> for ProHashMapDatasetDb {
...
         let options = options.user_input;
 
+        let iter = self
+            .dataset_permissions
+            .iter()
+            .filter(|p| session.roles.contains(&p.role))
+            .map(|p| {
+                self.datasets
+                    .get(&p.dataset)
+                    .expect("a dataset has at least one permission")
```
ChristianBeilschmidt, line 219: It is rather: You wouldn't have got here if the filter wouldn't have matched and this means: There is a permission :smile: .
But you don't check that the dataset has a permissions but rather that the dataset exists for that permission.

services/src/pro/datasets/storage.rs
```diff
@@ -1,31 +1,71 @@
...
+}
+
+impl Role {
+    pub fn system_role_id() -> RoleId {
+        RoleId::from_str("d5328854-6190-4af9-ad69-4e74b0961ac9").expect("valid")
+    }
+
+    pub fn user_role_id() -> RoleId {
+        RoleId::from_str("4e8081b6-8aa6-4275-af0c-2fa2da557d28").expect("valid")
+    }
```
ChristianBeilschmidt, line 35: What is this? User's should have their own id, right?

ChristianBeilschmidt (changes requested):

services/src/datasets/listing.rs
michaelmattig: it is never used and not even implemented for the existing providers. For internal datasets we use it though and before this PR they had the same trait. That's why it still exists.

services/src/handlers/datasets.rs
michaelmattig: you mena hat the DatasetDB could handle the permissions? I think the DatasetDB should just check whether the user is allowed to access the external provider at all, otherwise fail. And if they are the handler should get the credentials for the given user and provider and put it into the list method.

services/src/pro/datasets/storage.rs
michaelmattig: This is the equivalent to the user group in VAT. A user has their own id, but all the registered users get assigned this user_role additionaly for them to see all the datasets that shall be visible to all users.

services/src/pro/contexts/postgres.rs
michaelmattig: The view shows for each user the permitted datasets. So we need to make sure the role is actually a user. Thus the join

services/src/pro/contexts/postgres.rs
michaelmattig: yes, I will change it when/if we agree on the idea or users and roles

services/src/datasets/in_memory.rs
michaelmattig: adab347397faa698db32ae904b15882856f6dc56

services/src/datasets/listing.rs
michaelmattig: adab347397faa698db32ae904b15882856f6dc56

services/src/pro/datasets/in_memory.rs
michaelmattig: adab347397faa698db32ae904b15882856f6dc56

services/src/datasets/listing.rs
ChristianBeilschmidt: I guess if we don't use it and do not even implement it, then we should remove it.

services/src/handlers/datasets.rs
ChristianBeilschmidt: Ah okay, I though calling a dataset provider with a session "primes" it with this session and then you can just call methods on it.

services/src/pro/contexts/postgres.rs
ChristianBeilschmidt: But in `user_roles`, each id should be a user id.

services/src/pro/datasets/in_memory.rs
ChristianBeilschmidt: Should we really panic here?

I guess it is one thing to say:
```
len() > 0
get(0).expect("should be there")
```

The other thing is
```
assert!(a[0].is_some());
b[0].expect("must be there, we have this assumption")
```

I guess the one thing cannot happen and the other thing means, we fucked up in consistency.
But would be panic then?

services/src/datasets/listing.rs
michaelmattig: acac67181c8096c46aa941c5b43072a8f5ee32c9

services/src/pro/contexts/postgres.rs
michaelmattig: you are right 66724e08fe64b5bec6476afa3a89cb79ed52ea70

services/src/pro/contexts/postgres.rs
michaelmattig: 66724e08fe64b5bec6476afa3a89cb79ed52ea70

services/src/pro/datasets/in_memory.rs
michaelmattig: instead of `expect` I could `.filter_map(|p| self.datasets.get(&p.dataset));` but then it would be hard to debug why a dataset is not listed. Shall I log a warning then?

michaelmattig: 67fc660061e6cc71840b87653ed066137a5f9415

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors merge

PR #392, opened 2021-10-29
Scatter plot

koerberm:


operators/src/plot/scatter_plot.rs
```diff
@@ -0,0 +1,1009 @@
...
+        Ok(TypedPlotQueryProcessor::JsonVega(processor.boxed()))
+    }
+}
+
+/// A query processor that calculates the scatter plot about its vector input.
+pub struct ScatterPlotQueryProcessor {
+    input: TypedVectorQueryProcessor,
+    column_x: String,
+    column_y: String,
+}
```
ChristianBeilschmidt, line 117: Would it be a large overhead to allow two rasters as input?

Michael wrote some temporal alignment adapter and when we query the raster with the same bbox, they are spatially aligned.

Maybe we should also talk to Johannes if this operation is useful.

In any case, maybe it could be also a different task.

ChristianBeilschmidt (changes requested):

operators/src/plot/scatter_plot.rs
koerberm: We made this a new task: https://trello.com/c/viUYH237/466-scatter-plot-f%C3%BCr-2-raster

operators/src/plot/scatter_plot.rs
```diff
@@ -0,0 +1,1009 @@
...
+/// The maximum number of elements for a scatter plot
+const SCATTER_PLOT_THRESHOLD: usize = 500;
+
+/// The number of elements to process at once (i.e., without switching from scatter-plot to histogram)
+const BATCH_SIZE: usize = 1000;
+
+/// The maximum number of elements before we turn the collector into a histogram.
+/// At this point, the bounds of the histogram are fixed (i.e., further values exceeding
+/// the min/max seen so far are ignored)
+const COLLECTOR_TO_HISTOGRAM_THRESHOLD: usize = 10_000;
```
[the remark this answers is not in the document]
koerberm: 1. a860f5c24007686abc43051e49dd258291c4b218
2. I added a new task: https://trello.com/c/KFh9p6La/467-histogram-bounds-calculation

ChristianBeilschmidt (approved):

PR #393, opened 2021-11-01
Switch to Rayon thread pool

ChristianBeilschmidt:


ChristianBeilschmidt:
[Thread-Pool austauschen durch Rayon](https://trello.com/c/m0CuNQbq/453-thread-pool-austauschen-durch-rayon)

operators/src/util/rayon.rs
```diff
@@ -0,0 +1,52 @@
...
+                "Do not spawn rayon global pool on purpose",
+            ))
+        })
+        .build_global()
+        .is_err());
+}
+
+/// Create a rayon thread pool with the given number of threads.
+/// Use `num_threads = 0` for auto number of threads.
+pub fn create_rayon_thread_pool(num_threads: usize) -> Arc<ThreadPool> {
```
michaelmattig, line 28: this method is used in 9 places across the code but may only be called exactly once in the entire lifetime of a Geo Engine instance. Can we encapsulate it better, so there are as few places that use it as possible? E.g. don't create the thread pool in the MockExecution context but in the context context?

please also indicate in the comment that this method will fail on subsequent calls.

michaelmattig (changes requested):

operators/src/util/rayon.rs
ChristianBeilschmidt: But how/why should I prevent it from being called in the contexts. You could call it with auto-Threads, specify it, etc. Should I create a wrapper type around it that implements all rayon methods?

Moreover, why does this fail on subsequent calls? The tests create contexts in parallel and this works.

michaelmattig (approved):

PR #396, opened 2021-11-02
add crs to plot requests and result descriptors

michaelmattig:


michaelmattig:
[Plot: CRS angeben lassen und auto-projizieren](https://trello.com/c/qe5s6zn1/456-plot-crs-angeben-lassen-und-auto-projizieren)

services/src/handlers/plots.rs
```diff
@@ -135,10 +142,26 @@ async fn get_plot_handler<C: Context>(
         .await
         .context(error::Operator)?;
 
+    // handle request and workflow crs matching
```
jdroenner, line 145: is this identical to wfs and wms ? if yes, could we make a util method?

services/src/handlers/plots.rs
```diff
@@ -135,10 +142,26 @@ async fn get_plot_handler<C: Context>(
...
+    let workflow_spatial_ref: Option<SpatialReference> =
+        initialized.result_descriptor().spatial_reference().into();
+    let workflow_spatial_ref = workflow_spatial_ref.ok_or(error::Error::InvalidSpatialReference)?;
+
+    let request_spatial_ref: SpatialReference = params.crs.unwrap_or(workflow_spatial_ref);
```
jdroenner, line 150: i guess this should be a default value instead  of using the result of the workflow.  it might happen that the user does not know what the resulting srs is.  And if  the user really wants this behavior the user can request the workflows srs.

jdroenner (changes requested):

services/src/handlers/plots.rs
michaelmattig: it is not identical. In WMS/WFS we append a reprojection operator (which reprojects the bbox). Here, we reproject the bbox ourselves (because there is no reprojection operator for plots).

services/src/handlers/plots.rs
michaelmattig: so I should make the SpatialReference in the params mandatory instead of Option?

jdroenner: i think the  params.crs.unwrap_or(workflow_spatial_ref); should default to a specific SRS. we have the same TODO in wms :(

    // TODO: use a default spatial reference if it is not set?
    let request_spatial_ref: SpatialReference =
        request.crs.ok_or(error::Error::InvalidSpatialReference)?;

michaelmattig: If no CRS is given we can either
- throw an error
- use a default CRS

both is fine with me. If you want it changed you have to tell me which default crs I should use.

jdroenner: then throw an error like wms does it and we schould  find a common solution

michaelmattig: 124d898158afea7f20987d03fb47fe09432cc538

jdroenner (approved):

jdroenner (approved):

PR #416, opened 2021-12-08
Async_msg

jdroenner:


jdroenner:
[use spawn_blocking for work in msg operators](https://trello.com/c/BqN455Hi/513-use-spawnblocking-for-work-in-msg-operators)

operators/src/processing/meteosat/mod.rs
```diff
@@ -5,28 +5,28 @@ mod reflectance;
 mod satellite;
 mod temperature;
 
-fn slope_key() -> RasterPropertiesKey {
+fn new_slope_key() -> RasterPropertiesKey {
```
michaelmattig, line 8: can those be `const` functions, or just constants? They are always the same

operators/src/processing/meteosat/radiance.rs
```diff
@@ -191,47 +232,19 @@ where
...
-                        tile.global_geo_transform,
-                        out.into(),
-                        tile.properties,
-                    ))
-                }
-                Err(e) => Err(e),
+        let rs = src
+            .map(move |tile| {
+                tile.and_then(|t| {
+                    let offset = t.properties.number_property::<f32>(&self.offset_key)?;
```
michaelmattig, line 238: could these be determined outside of the `map`?

operators/src/processing/meteosat/temperature.rs
```diff
@@ -197,22 +199,60 @@ where
             })
         }
     }
+}
+
+fn create_lookup_table(channel: &Channel, offset: f64, slope: f64) -> Vec<f32> {
+    // let channel = self.channel(tile, satellite)?;
```
michaelmattig, line 205: remove lines

operators/src/processing/meteosat/temperature.rs
```diff
@@ -228,47 +268,28 @@ where
...
-                            out.data[idx] = *lut
-                                .get(lut_idx as usize)
-                                .ok_or(Error::UnsupportedRasterValue)?;
-                        }
+        let rs = src
+            .map(move |tile| {
+                tile.and_then(|t| {
+                    let satellite = self.satellite(&t)?;
+                    let channel = self.channel(&t, satellite)?;
+                    let offset = t.properties.number_property::<f64>(&self.offset_key)?;
```
michaelmattig, line 276: could these be determined outside of the `map`?

michaelmattig (changes requested):
Would have been a good opportunity to also parallelize these things, but we can do it in another task.

operators/src/processing/meteosat/temperature.rs
jdroenner: no you need the tile

operators/src/processing/meteosat/radiance.rs
jdroenner: no you need the tile for that

operators/src/processing/meteosat/mod.rs
jdroenner: i guess we will have them as options the user can set in a next version. So i think we should leave it like this. They can't be const since Strings are on the heap

operators/src/processing/meteosat/temperature.rs
jdroenner: did that

jdroenner:
i reworked them to use the pool for parallelism

michaelmattig (approved):

michaelmattig:
bors merge

PR #418, opened 2021-12-10
Nfdi

koerberm:
A first implementation of a connection to the NFDI core storage; It currently supports single-file vector and raster data;


koerberm:
[NFDI-Core-Storage-Data-Provider](https://trello.com/c/c3GGDaIo/446-nfdi-core-storage-data-provider)

Settings-default.toml
```diff
@@ -33,7 +33,7 @@ path = "upload"
 [logging]
 # Minimum log level. Can be one of error, warn, info, debug, trace
 # or a more detailed spec. See https://docs.rs/flexi_logger/0.17.1/flexi_logger/struct.LogSpecification.html.
-log_spec = "info"
+log_spec = "debug"
```
ChristianBeilschmidt, line 36: We should not change the default here.

Settings-test.toml
```diff
@@ -9,6 +9,11 @@ password = "geoengine"
 [operators.gdal_source]
 raster_data_root_path = "../test_data/raster" # relative to sub crate directory for tests
 
+[logging]
+# Minimum log level. Can be one of error, warn, info, debug, trace
+# or a more detailed spec. See https://docs.rs/flexi_logger/0.17.1/flexi_logger/struct.LogSpecification.html.
+log_spec = "debug"
```
ChristianBeilschmidt, line 15: Do we need `debug` logging here or is this just for own debug purposes?

services/Cargo.toml
```diff
@@ -15,6 +15,7 @@ xml = ["quick-xml"]
 postgres = ["postgres-types", "bb8-postgres"]
 odm = ["pro"]
 gfbio = ["postgres", "geoengine-datatypes/postgres"]
+nfdi = []
```
ChristianBeilschmidt, line 18: I guess we should merge gfbio and nfdi as a flag. I suggest both is nfdi now.

services/Cargo.toml
```diff
@@ -53,6 +54,7 @@ pwhash = "1.0"
 quick-xml = { version = "0.22", optional = true }
 rayon = "1.5"
 reqwest = { version = "0.11", features = ["json", "multipart", "stream"] }
+scienceobjectsdb_rust_api = { git = "https://github.com/ScienceObjectsDB/rust-api", branch="main" }
```
ChristianBeilschmidt, line 57: Can't we move this dependency to only be used if nfdi is enabled?
It seems to be an unstable dependency by now.

services/src/datasets/external/nfdi/mod.rs
```diff
@@ -0,0 +1,1031 @@
...
+
+        OgrSourceDataset {
+            file_name: PathBuf::from(link),
+            layer_name,
+            data_type,
+            time,
+            default_geometry: None,
+            columns: Some(column_spec),
+            force_ogr_time_filter: false,
+            force_ogr_spatial_filter: false,
```
ChristianBeilschmidt, line 327: Do filters work or do we have to force this?

services/src/datasets/external/nfdi/mod.rs
```diff
@@ -0,0 +1,1031 @@
...
+                    _phantom: Default::default(),
+                    object_stub: self.object_stub.clone(),
+                };
+                Ok(Box::new(res))
+            }
+            DataType::SingleVectorFile(_) => Err(geoengine_operators::error::Error::InvalidType {
+                found: md.data_type.to_string(),
+                expected: "SingleRasterFile".to_string(),
+            }),
+        }
```
ChristianBeilschmidt, line 469: For single files, the 15-minute-key should be fine. We have to find another approach for multiple files and large (chunked) ogr sources.

services/src/datasets/external/nfdi/mod.rs
```diff
@@ -0,0 +1,1031 @@
...
+    //     let result: Vec<MultiPointCollection> = proc
+    //         .query(qr, &ctx)
+    //         .await
+    //         .unwrap()
+    //         .map(Result::unwrap)
+    //         .collect()
+    //         .await;
+    //     let sum: usize = result.iter().map(MultiPointCollection::len).sum();
+    //     println!("{}", sum);
+    // }
```
ChristianBeilschmidt, line 1030: I guess these are up to implement

ChristianBeilschmidt (changes requested):

Settings-default.toml
koerberm: 8fb81dbed4ec76c5ed14d5584083f7938818d7fe

Settings-test.toml
koerberm: Was for local testing. Removed it. 8fb81dbed4ec76c5ed14d5584083f7938818d7fe

services/Cargo.toml
koerberm: c5fa0fb1b47bcbd79d7efced816f5f360cbc0d39

services/Cargo.toml
koerberm: d760b3962b50d6acba8502aa4527c7b00ba83b5d

services/src/datasets/external/nfdi/mod.rs
koerberm: Yes, I just tested it.

services/src/datasets/external/nfdi/mod.rs
koerberm: I think we need to discuss this with Marius

services/src/datasets/external/nfdi/mod.rs
koerberm: Should I remove them? We could basically reuse them after we mocked the core-storage API

ChristianBeilschmidt (dismissed):

ChristianBeilschmidt (approved):

PR #425, opened 2021-12-20
Build-rs

ChristianBeilschmidt:
I did the following things:

1. Removed `anyhow` from our build-dependencies and just unwrapped the result in the `build.rs` file.
2. Clippy wanted to add #[must_use] labels to functions that take `&self` and return `Self`. This makes sense since not using would mean you did something wrong.
3. The arrow example was not working anymore :shrug:. I changed the pointer cast. Fortunately, it seems that our in-code conversions still work.
4. I updated `tokio` and could remove some Clippy lint exclusions (`// TODO: remove when https://github.com/tokio-rs/tokio/issues/4245 is fixed`).
5. I updated Actix-web and they removed `AnyBody`. This means I had to wrap our error handler differently. You now have to put it into an EitherBody-enum and this takes a BoxBody. So I had to call `boxed()` here as well. Please check that the HTTP responses still work, for me, it looks fine.

### DIFF
--- a/datatypes/src/operations/image/colorizer.rs
+++ b/datatypes/src/operations/image/colorizer.rs
@@ -529,6 +529,7 @@ impl RgbaColor {
     /// On debug, if factor is not in [0, 1]
     ///
     #[allow(unstable_name_collisions)]
+    #[must_use]
     pub fn factor_add(self, other: Self, factor: f64) -> Self {
         debug_assert!((0.0..=1.0).contains(&factor));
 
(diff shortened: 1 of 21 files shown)


ChristianBeilschmidt:
[Build-rs vereinfachen + 1 Clippy Lint](https://trello.com/c/YtN3CcFh/519-build-rs-vereinfachen-1-clippy-lint)

jdroenner:
bors merge

PR #435, opened 2022-01-06
faster and multithreaded projection implementation

jdroenner:


jdroenner:
[Multithreaded Projection](https://trello.com/c/IJiakKEh/532-multithreaded-projection)

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
```diff
@@ -144,6 +116,98 @@ where
...
+                        let y_idx = lin_idx / axis_size_x + chunk_start_y;
+                        let grid_idx = GridIdx2D::from([y_idx as isize, x_idx as isize]);
+                        tile_geo_transform.grid_idx_to_upper_left_coordinate_2d(grid_idx)
+                    })
+                    .collect::<Vec<Coordinate2D>>();
+
+                let chunk_bounds =
+                    SpatialPartition2D::new_unchecked(out_coords[0], out_coords[chunk_len - 1]);
+
+                let proj = CoordinateProjector::from_known_srs(out_srs, in_srs)?;
```
ChristianBeilschmidt, line 175: Is it necessary to call this per chunk? Couldn't this be initialized beforehand?

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
```diff
@@ -144,6 +116,98 @@ where
...
+    out_srs: SpatialReference,
+    in_srs: SpatialReference,
+    valid_out_area: &SpatialPartition2D,
+) -> Result<Grid2D<Option<Coordinate2D>>> {
+    const OVERPROVISIONING: usize = 1; // split the work into multiple chunks per thread.
+    const MAX_PARALELLISM: usize = 8; // max task / work split
+
+    let start = std::time::Instant::now();
+
+    let res = pool.install(|| {
```
ChristianBeilschmidt, line 131: Do you have this wrapped into a `spawn_blocking` at some point where this is called?

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
```diff
@@ -144,6 +116,98 @@ where
...
 }
 
+fn projected_coordinate_grid_parallel(
+    pool: &ThreadPool,
+    tile_info: TileInformation,
+    out_srs: SpatialReference,
+    in_srs: SpatialReference,
+    valid_out_area: &SpatialPartition2D,
+) -> Result<Grid2D<Option<Coordinate2D>>> {
+    const OVERPROVISIONING: usize = 1; // split the work into multiple chunks per thread.
```
ChristianBeilschmidt, line 126: Why would you want to do this?

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
```diff
@@ -144,6 +116,98 @@ where
...
+    const MAX_PARALELLISM: usize = 8; // max task / work split
```
ChristianBeilschmidt, line 127: Is it better to have a max parallelism or a minimum chunk size?
Maybe the latter one is more robust when the chunk size would change.

ChristianBeilschmidt (changes requested):

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: it is usefull when there is a large ddifference of work done in chunks. Currently we have as many chunks as we  have threads. So the runtime ultimately depends on the slowest task.

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: no and we know that this is an issue. There is a case missing in the subquery adapter. Here we need to await the accu creation:  https://github.com/geo-engine/geoengine/blob/624e2a7461f243271cae391b3c7b19c6929210f7/operators/src/adapters/raster_subquery/raster_subquery_adapter.rs#L312


however. this is not in the scope of this PR

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: i do not know but will test it. i thought proj objects are not threadsafe / send /sync.

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: changed that

jdroenner: 7508587

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: i removed it

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: i changed this in 22e6994

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
```diff
@@ -116,15 +104,50 @@ where
...
 fn projected_coordinate_grid_parallel(
     pool: &ThreadPool,
     tile_info: TileInformation,
     out_srs: SpatialReference,
     in_srs: SpatialReference,
     valid_out_area: &SpatialPartition2D,
 ) -> Result<Grid2D<Option<Coordinate2D>>> {
-    const OVERPROVISIONING: usize = 1; // split the work into multiple chunks per thread.
-    const MAX_PARALELLISM: usize = 8; // max task / work split
+    const MIN_ROWS_IN_PAR_CHUNK: usize = 64; // this must never be smaller than 1
```
ChristianBeilschmidt, line 150: If you use min rows instead of pixels, doesn't this mean that if the tile size increases, you say that one thread needs to process more pixels in order to be effective? And if you lower the tile size, you say it needs less to process pixels in order to be effective.

I guess the number of pixels is also highly platform-dependent. It depends on the CPU clock, on SIMD register sizes, etc.

Do we find a robust way for this?

How does rayon decide on how much to schedule? Maybe we can do something similar.

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
```diff
@@ -237,8 +260,7 @@ pub fn fold_by_coordinate_lookup_impl<T>(
 where
     T: Pixel,
 {
-    const OVERPROVISIONING: usize = 1; // split the work into multiple chunks per thread.
-    const MAX_PARALELLISM: usize = 16; // max task / work split
+    const MIN_ROWS_IN_PAR_CHUNK: usize = 32; // this must never be smaller than 1
```
ChristianBeilschmidt, line 263: Here, the size differs from above. But why? Is this empirically determined?

ChristianBeilschmidt (changes requested):
I guess the logic itself looks fine. We just need to discuss our magic constants once more before merging.

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: I use rows since that iis a very easy way to determine the pixel location. if we do it otherwise we have to assert that in each setup we use a number of pixels that is a multiple off a row lenght. I do not think we want to take that route. At least i don't want to implement that logic.
Yes, this also means that with larger tiles we have more pixels per thread but keep in mind that both axis increase when you change the tile size. We should never go below 512pix tiles.  In fact i think we should go up to 1024.

We will never find a perfect solution for all CPUs, CPU  Speed, CPU Arch,, RAM, HDDs, Network, .... We can determine what works and use that. I don't want to have any caps but i  guess the overhead of rayon is a think we have to deal with.

As i understand it rayon simply spawns as man threads as there are CPUs. If you use par_iter it will create one task per item. It simply does not decide how much work to schedule.

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: yes thats the 8/16 split from the graph i generated.

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
jdroenner: fb4f5b4

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #436, opened 2022-01-06
New-expression

ChristianBeilschmidt:


ChristianBeilschmidt:
[Expression mit PEST und Rust-AST](https://trello.com/c/DrizuwDV/530-expression-mit-pest-und-rust-ast)

operators/src/adapters/raster_conversion.rs
```diff
@@ -0,0 +1,45 @@
...
+    type SpatialBounds = SpatialPartition2D;
+
+    async fn query<'b>(
+        &'b self,
+        query: RasterQueryRectangle,
+        ctx: &'b dyn QueryContext,
+    ) -> Result<BoxStream<'b, Result<Self::Output>>> {
+        let stream = self.query_processor.query(query, ctx).await?;
+
+        let converted_steam = stream.map_ok(RasterTile2D::convert::<POut>);
```
michaelmattig, line 41: shouldn't we multithread here?

operators/src/processing/new_expression/mod.rs
```diff
@@ -0,0 +1,749 @@
...
+            b: b.transpose()?,
+            c: c.transpose()?,
+        })
+    }
+}
+
+/// Create a parameter name from an index.
+/// Starts with `A`.
+fn index_to_parameter(index: usize) -> String {
+    // TODO: handle indices > 26
```
michaelmattig, line 131: why? the source only supports three inputs

operators/src/processing/new_expression/parser.rs
```diff
@@ -0,0 +1,653 @@
...
+    fn build_ast(&self, pairs: Pairs<'_, Rule>) -> Result<AstNode> {
+        EXPRESSION_CLIMBER.climb(
+            pairs,
+            |pair| self.resolve_expression_rule(pair),
+            |left, op, right| self.resolve_infix_operations(left, &op, right),
+        )
+    }
+
+    fn resolve_expression_rule(&self, pair: Pair<Rule>) -> Result<AstNode> {
+        // dbg!(&pair);
```
michaelmattig, line 105: remove

operators/src/processing/new_expression/mod.rs
```diff
@@ -0,0 +1,749 @@
...
+        let query_processors: Vec<TypedRasterQueryProcessor> = self
+            .sources
+            .iter()
+            .map(InitializedRasterOperator::query_processor)
+            .collect::<Result<_>>()?;
+
+        Ok(match query_processors.len() {
+            1 => {
+                let [a] = <[_; 1]>::try_from(query_processors).expect("len previously checked");
+                call_on_generic_raster_processor!(a, p_a => {
```
michaelmattig, line 268: why is this not turned into_f64?

operators/src/processing/new_expression/mod.rs
```diff
@@ -0,0 +1,749 @@
...
+                    ExpressionQueryProcessor::new(
+                        expression,
+                        query_processors,
+                        output_no_data_value.as_(),
+                        self.map_no_data,
+                    )
+                    .boxed()
+                )
+            }
+            _ => return Err(crate::error::Error::InvalidNumberOfExpressionInputs), // TODO: handle more than two inputs
```
michaelmattig, line 324: more than three I guess

operators/src/processing/new_expression/query_processor.rs
```diff
@@ -0,0 +1,401 @@
...
+use super::compiled::LinkedExpression;
+
+pub struct ExpressionQueryProcessor<TO, Sources>
+where
+    TO: Pixel,
+{
+    pub sources: Sources,
+    pub phantom_data: PhantomData<TO>,
+    pub program: Arc<LinkedExpression>,
+    pub no_data_value: TO,
```
michaelmattig, line 29: if everything is turned into f64, do we need this generic?

operators/src/processing/new_expression/query_processor.rs
```diff
@@ -0,0 +1,401 @@
...
+                    thread_pool.install(move || {
+                        Tuple::compute_expression(rasters, &program, map_no_data, out_no_data)
+                    })
+                })
+                .await??;
+
+                let out =
+                    Grid2D::<TO>::new(output_grid_shape, data, Some(self.no_data_value))?.into();
+
+                Ok(RasterTile2D::new(
```
michaelmattig, line 95: where do you transform the pixels into the desired output data type?

operators/src/processing/new_expression/query_processor.rs
```diff
@@ -0,0 +1,401 @@
...
+            program.function_3::<f64, bool, f64>()?
+        };
+
+        // cannot be empty at this point
+        let tile = raster.into_materialized_tile();
+
+        let data = tile
+            .grid_array
+            .data
+            .par_iter()
```
michaelmattig, line 190: I think we need to chunk here, otherwise the overhead of the task is too high

operators/src/processing/new_expression/query_processor.rs
```diff
@@ -0,0 +1,401 @@
...
+            // we have to "trust" that the function has the signature we expect
+            program.function_5::<f64, bool, f64, bool, f64>()?
+        };
+
+        // TODO: allow iterating over empty rasters
+        let tile_0 = rasters.0.into_materialized_tile();
+        let tile_1 = rasters.1.into_materialized_tile();
+
+        let data = (&tile_0.grid_array.data, &tile_1.grid_array.data)
+            .into_par_iter()
```
michaelmattig, line 272: I think we need to chunk here

operators/src/processing/new_expression/query_processor.rs
```diff
@@ -0,0 +1,401 @@
...
+        let tile_0 = rasters.0.into_materialized_tile();
+        let tile_1 = rasters.1.into_materialized_tile();
+        let tile_2 = rasters.2.into_materialized_tile();
+
+        let data = (
+            &tile_0.grid_array.data,
+            &tile_1.grid_array.data,
+            &tile_2.grid_array.data,
+        )
+            .into_par_iter()
```
michaelmattig, line 376: I think we need to chunk here

operators/src/adapters/raster_conversion.rs
ChristianBeilschmidt: Should we remove the conversion method(s) then from `RasterTile2D` or should we insert means to thread? I guess just duplicating it with some threading does not make sense.
Moreover, I'm not sure if a conversion for all elements in a vector isn't SIMD optimized from a compiler :shrug:.

operators/src/processing/new_expression/query_processor.rs
ChristianBeilschmidt: We output a Raster of type `TO`, so I guess this makes sense.

operators/src/processing/new_expression/codegen.rs
```diff
@@ -0,0 +1,353 @@
...
+
+type Result<T, E = ExpressionError> = std::result::Result<T, E>;
+
+// TODO: prefix for variables and functions
+
+/// An expression as an abstract syntax tree.
+/// Allows genering Rust code.
+#[derive(Debug, Clone)]
+pub struct ExpressionAst {
+    name: Identifier,
```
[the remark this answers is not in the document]
ChristianBeilschmidt: It is the function name that is generated in the Rust module.

operators/src/processing/new_expression/compiled.rs
```diff
@@ -0,0 +1,126 @@
...
+/// Compiles and links an expression as a program and offers means to call it
+pub struct LinkedExpression {
+    library_folder: ManuallyDrop<TempDir>,
+    library: ManuallyDrop<Library>,
+    function_name: String,
+}
+
+impl LinkedExpression {
+    pub fn new(ast: &ExpressionAst) -> Result<Self> {
+        let library_folder = tempfile::tempdir().map_err(|error| {
```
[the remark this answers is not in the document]
ChristianBeilschmidt: We need two files, one for rustc and one for its output, which is used by the linker.

We could create separate task that tries to fiddle around with https://crates.io/crates/memfile.

operators/src/processing/new_expression/mod.rs
ChristianBeilschmidt: 084ff66

operators/src/processing/new_expression/mod.rs
```diff
@@ -0,0 +1,749 @@
...
+        // TODO: handle more then 2 inputs, i.e. 1-8
+        ensure!(
+            (1..=3).contains(&self.sources.number_of_sources()),
+            crate::error::InvalidNumberOfRasterInputs {
+                expected: 1..4,
+                found: self.sources.number_of_sources()
+            }
+        );
+
+        // we refer to raster's by A, B, C, …
```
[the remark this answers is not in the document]
ChristianBeilschmidt: Layer names are Schall und Rauch for Geo Engine. In our workflow, they don't have names. Thus, we cannot refer to them by any names.


0132b78

operators/src/processing/new_expression/mod.rs
```diff
@@ -0,0 +1,749 @@
...
+                .output_type
+                .is_valid(self.params.output_no_data_value),
+            crate::error::InvalidNoDataValueValueForOutputDataType
+        );
+
+        let sources = self.sources.initialize(context).await?;
+
+        let spatial_reference = sources.a.result_descriptor().spatial_reference;
+
+        for other_spatial_refenence in sources
```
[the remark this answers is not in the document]
ChristianBeilschmidt: 0132b78

operators/src/processing/new_expression/mod.rs
```diff
@@ -0,0 +1,749 @@
...
+                expected: 1..4,
+                found: self.sources.number_of_sources()
+            }
+        );
+
+        // we refer to raster's by A, B, C, …
+        let parameters = (0..self.sources.number_of_sources())
+            .flat_map(|i| {
+                let parameter = index_to_parameter(i);
+                let boolean_parameter = format!("is_{}_nodata", &parameter);
```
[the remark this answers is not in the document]
ChristianBeilschmidt: We cannot just build in NODATA checks into the language because we need the nodata value be fed from the outside. Otherwise, every pixel is an Option<f64>. This could work, but we really have to think about doing this.

We could name the variable `A_is_nodata` and when we see `A IS NODATA` somewhere, we just map the boolean expression to our boolean variable by adding underscores and checking if it actually exists.

I'm not sure what this syntax with the Slash is, but the other part of leaving out curly braces could be quite difficult for the parser due to a lack of differentiating symbols.

operators/src/processing/new_expression/mod.rs
```diff
@@ -0,0 +1,749 @@
...
+
+        sources.into_iter()
+    }
+}
+
+impl InitializedRasterOperator for InitializedExpression {
+    fn query_processor(&self) -> Result<TypedRasterQueryProcessor> {
+        let output_type = self.result_descriptor().data_type;
+        // TODO: allow processing expression without NO DATA
+        let output_no_data_value = self.result_descriptor().no_data_value.unwrap_or_default();
```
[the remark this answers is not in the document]
ChristianBeilschmidt: I copied this from the current expression. So your take would be to disallow rasters without NODATA by not allowing initializing the `InitializedRasterOperator`?

operators/src/processing/new_expression/mod.rs
ChristianBeilschmidt: Since we don't have such an explosion of symbols here, I thought we could spare the extra conversion.

operators/src/processing/new_expression/mod.rs
ChristianBeilschmidt: 5325488

operators/src/processing/new_expression/codegen.rs
```diff
@@ -0,0 +1,353 @@
...
+
+        for fn_name in &self.imports {
+            let prefixed_fn_name = format_ident!("import_{}", fn_name.as_ref());
+
+            tokens.extend(quote! {
+                #[inline]
+            });
+
+            let fn_tokens = match fn_name.as_ref() {
+                "max" => quote! {
```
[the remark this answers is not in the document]
ChristianBeilschmidt: My take would be to create a second task that deals with the functions:
We first collect a list of what we need and then implement it and perform good checks that these functions exist.

operators/src/processing/new_expression/mod.rs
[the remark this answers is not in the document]
jdroenner: An alternative approach would be to do  it like A.value  and A.is_no_data

operators/src/processing/new_expression/parser.rs
ChristianBeilschmidt: e287f4b

operators/src/processing/new_expression/query_processor.rs
jdroenner: i guess we have to benchmark this

operators/src/adapters/raster_conversion.rs
jdroenner: I guess we should create a task where we move the methods to  datatypes /  operations  we can implement them using rayon and replace the "old" single-threaded ones where possible.

operators/src/processing/new_expression/query_processor.rs
ChristianBeilschmidt: It is not too high, at least it outperforms OpenCL. And also the experiments prior to the implementation showed that this is not a major problem.

However, in a separate task, one could measure this and try to improve the performance.

operators/src/processing/new_expression/query_processor.rs
ChristianBeilschmidt: It is not too high, at least it outperforms OpenCL int the benchmark. And also the experiments prior to the implementation showed that this is not a major problem.

However, in a separate task, one could measure this and try to improve the performance.

operators/src/processing/new_expression/query_processor.rs
ChristianBeilschmidt: It is not too high, at least it outperforms OpenCL int the benchmark. And also the experiments prior to the implementation showed that this is not a major problem.

However, in a separate task, one could measure this and try to improve the performance.

operators/src/processing/new_expression/query_processor.rs
ChristianBeilschmidt: `data` has to be of type `TO` because of the `Grid2D::<TO>` bound here.

For instance, in line 199, you can spot `TO::from_(result)`.

operators/src/processing/new_expression/mod.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: This would mean a struct for each Pixel?

operators/src/adapters/raster_conversion.rs
ChristianBeilschmidt: :+1:

operators/src/processing/new_expression/mod.rs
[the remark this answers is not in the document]
jdroenner: yes but it has  the benefit that you can pass A.x/lon A.y/lat A.no_data_value A.grid_x A.grid_y ...

operators/src/processing/new_expression/mod.rs
michaelmattig: I don't see this addressed in the commit

operators/src/processing/new_expression/mod.rs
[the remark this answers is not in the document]
michaelmattig: > We could name the variable A_is_nodata and when we see A IS NODATA somewhere, we just map the boolean expression to our boolean variable by adding underscores and checking if it actually exists.

I'd be very much in favor of something like this. It would also make the function signature shown in the UI smaller because the `IS NODATA` is universal while the `is_X_nodata` have to be shown to the user for every input. I wouldn't expose the `is_x_nodata` then and only use it internally.

operators/src/processing/new_expression/mod.rs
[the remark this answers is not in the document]
michaelmattig: either enforce a NoData value by throwing an error during initialization or make it an Option in the Processore and throw an error when the expression tries to output no data

operators/src/processing/new_expression/compiled.rs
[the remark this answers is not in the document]
michaelmattig: We should only do that if creating the files on disk is actually a noticeable overhead. Can you measure that easily? If it's negligible we should just leave it and optimize it later if it ever becomes a bottleneck

operators/src/processing/new_expression/codegen.rs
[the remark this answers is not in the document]
michaelmattig: maybe you can add a doc comment for that because it wasn't obvious to me. I thought the Ast ist something different than the generated function.

operators/src/processing/new_expression/mod.rs
michaelmattig: can't that lead to different results if I have an expression on one raster and it operates on the original type and I add a second input and now it operates on float or do I misunderstand it?

operators/src/processing/new_expression/mod.rs
ChristianBeilschmidt: a74a07c

operators/src/processing/new_expression/compiled.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: I don't know except trying both versions and comparing them in benchmarks.

operators/src/processing/new_expression/codegen.rs
```diff
@@ -0,0 +1,353 @@
...
+pub struct Branch {
+    pub condition: BooleanExpression,
+    pub body: AstNode,
+}
+
+#[derive(Debug, Clone)]
+pub enum BooleanExpression {
+    Variable(Identifier),
+    Constant(bool),
+    Comparison {
```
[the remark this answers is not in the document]
ChristianBeilschmidt: Isn't `BooleanExpression` your predicate? It can contain comparisons, a constant (bool), nested comparisons, …
just currently not yet any function call.

operators/src/processing/new_expression/codegen.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: 53a6fbd

operators/src/processing/new_expression/mod.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: d5bfca6

operators/src/processing/new_expression/mod.rs
ChristianBeilschmidt: 46856e5

operators/src/processing/new_expression/mod.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: 901d41f

ChristianBeilschmidt:
I've added the option get up to eight inputs: 
da88075

michaelmattig (approved):

michaelmattig:
bors merge

PR #441, opened 2022-01-12
debugging cog overviews

michaelmattig:


michaelmattig:
[COG-Low-Level-Overviews sind leer](https://trello.com/c/cl4iuKWq/531-cog-low-level-overviews-sind-leer)

PR #504, opened 2022-03-15
Temporally align 2 raster sources

michaelmattig:


michaelmattig:
[Zeitreihen-Kombinator](https://trello.com/c/YnhZ58cx/654-zeitreihen-kombinator)

operators/src/adapters/raster_time.rs
```diff
@@ -106,15 +113,44 @@ where
...
+                        return Poll::Ready(Some(Err(e)));
+                    }
+                }
+            }
+        }
+
+        let next = if let Some(stream) = stream.as_mut().as_pin_mut() {
+            ready!(stream.poll_next(cx))
+        } else {
+            panic!("stream is None")
```
ChristianBeilschmidt, line 152: You missed a panic here

operators/src/adapters/raster_time.rs
```diff
@@ -128,14 +185,15 @@ where
...
+                    // advance current query rectangle
+                    query_rect.time_interval = TimeInterval::new_unchecked(
                         min(tile_a.time.end(), tile_b.time.end()),
                         query_rect.time_interval.end(),
                     );
-                    *time_end = None;
 
-                    stream.set(source_a(next_qrect).zip(source_b(next_qrect)));
+                    // reset the stream, the new stream will be created in the next iteration
+                    stream.set(None);
```
ChristianBeilschmidt, line 196: Can we make it in such a way, that if the two streams are already temporally aligned, we do not query each time step again and again?

operators/src/adapters/raster_time.rs
```diff
@@ -16,45 +16,59 @@ use std::task::{Context, Poll};
...
+    F2::Output: Future<Output = Result<F2::Stream>>,
 {
+    // the first source (wrapped QueryProcessor)
     source_a: F1,
+    // the second source (wrapped QueryProcessor)
     source_b: F2,
+    #[pin]
+    // place for storing the output of the `query` method of the two sources
+    // and "awaiting" it
+    query_a_b_fut: Option<Join<F1::Output, F2::Output>>,
```
ChristianBeilschmidt, line 37: I think that is good and viable to extend to use a tuple of sources and a (already) tuple out joined/zipped tiles.

operators/src/adapters/raster_time.rs
```diff
@@ -87,34 +101,77 @@ where
...
+
+        // get the next pair of tiles from the stream. As the stream has to exist, this is safe.
+        let next = if let Some(stream) = stream.as_mut().as_pin_mut() {
+            ready!(stream.poll_next(cx))
+        } else {
+            panic!("stream is None")
+        };
 
         match next {
             Some((Ok(tile_a), Ok(tile_b))) => {
```
ChristianBeilschmidt, line 177: If we calculated this (lines below) upfront in the constructor, couldn't we make it easier since we do not have the option?

operators/src/adapters/raster_time.rs
```diff
@@ -87,34 +101,77 @@ where
...
 
-        let next = ready!(stream.as_mut().poll_next(cx));
+        // The `stream` is a `Zip` of the two input streams.
+        if stream.is_none() {
+            // the `stream` is not initialized and we need to query both sources and "await" their results
+            if query_a_b_fut.is_none() {
+                // the queries are not initialized and we need to query both sources and "await" their results
+                let fut1 = source_a.query(*query_rect);
+                let fut2 = source_b.query(*query_rect);
+                let join = future::join(fut1, fut2);
```
ChristianBeilschmidt, line 145: Why do you join an do not just use the zip adapter?

ChristianBeilschmidt (changes requested):

operators/src/adapters/raster_time.rs
michaelmattig: this ist kind of a MustNotHappenException. If we get here and the stream is None the logic in the code above is wrong. I think we should panic then. I could also maybe us an uwrap on the stream here.

operators/src/adapters/raster_time.rs
michaelmattig: yes, but we can't compute it before, because we need the first tile to compute it

operators/src/adapters/raster_time.rs
michaelmattig: this case would only arise if 
 - both time series are regular time series
 - both have the same start/reference and the same periodicity
 - there are nor gaps in the stream

for a given query processor output we don't know anything about the periodicity of its output. So unfortunately we can't avoid this. We would have to make additional metadata information about a queryprocessor output available.

operators/src/adapters/raster_time.rs
michaelmattig: was that a compliment or a suggestion? Please elaborate :D

operators/src/adapters/raster_time.rs
ChristianBeilschmidt: Well, it arises for our most prominent use case, the NDVI calculation from sentinel bands.

Why can't we just proceed if we query all the sources and everything matches?
I bet this really slows down everything.
Maybe we have to measure the overhead here.

operators/src/adapters/raster_time.rs
ChristianBeilschmidt: Yeah, well, this is a good first step. Then, we should use tuples for the sources and implement it for 2 to 12 sources or something like this.

operators/src/adapters/raster_time.rs
ChristianBeilschmidt: Shouldn't we then use states, e.g. an enum instead of a struct?

operators/src/adapters/raster_time.rs
ChristianBeilschmidt: Hm, yeah, well, looks like an anti-pattern :smile: 
Maybe we should look into states

michaelmattig: 332c216472eeb34b73e2bd3809877f580b97295d

operators/src/adapters/raster_time.rs
michaelmattig: I introduced states but still we need to have a tile form the source to get calculate the number of tiles :(

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #511, opened 2022-03-22
remove external_address value from default settings

michaelmattig:


services/src/handlers/wms.rs
```diff
@@ -441,10 +433,11 @@ mod tests {
     async fn test_get_capabilities() {
         let res = get_capabilities_test_helper(Method::GET).await;
 
-        assert_eq!(res.status(), 200);
+        // assert_eq!(res.status(), 200);
```
ChristianBeilschmidt, line 436: Why did you remove this?

services/src/handlers/wms.rs
```diff
@@ -441,10 +433,11 @@ mod tests {
...
 
         // TODO: validate against schema
         let body = actix_web::test::read_body(res).await;
+        dbg!(&body);
```
ChristianBeilschmidt, line 440:
```suggestion
```

services/src/handlers/wms.rs
```diff
@@ -454,26 +447,11 @@ mod tests {
...
-            "http://localhost/",
-            "http://localhost/wms/df756642-c5a3-4d72-8ad7-629d312ae993",
-        );
-        assert_wms_base(
-            "https://localhost/api/",
-            "https://localhost/api/wms/df756642-c5a3-4d72-8ad7-629d312ae993",
-        );
-        assert_wms_base(
-            "https://www.foobar.de/path/to/wms/",
-            "https://www.foobar.de/path/to/wms/wms/df756642-c5a3-4d72-8ad7-629d312ae993",
```
ChristianBeilschmidt, line 476: Why did you remove this?
Didn't it check that both root dir as URL and subfolder works?
I remember adding this for a bugfix.

services/src/handlers/wms.rs
```diff
@@ -214,9 +202,13 @@ where
         .body(response))
 }
 
-fn wms_url(external_address: &Url, workflow: WorkflowId) -> Result<Url> {
-    external_address
-        .join("wms/")?
+fn wms_url(workflow: WorkflowId) -> Result<Url> {
```
ChristianBeilschmidt, line 205: I guess we had the external address as a parameter here to test this method. It once was buggy.

ChristianBeilschmidt (changes requested):

services/src/handlers/wms.rs
michaelmattig: for seeing the output. I forgot to revert it back. b7ebd05ac4a1b0120193696140d6b238ee129fa0

services/src/handlers/wms.rs
michaelmattig: b7ebd05ac4a1b0120193696140d6b238ee129fa0

services/src/handlers/wms.rs
michaelmattig: it checked that for different base urls the correct wms endpoint URL was created. As the method now takes the base URL directly from the config (like wcs and wfs do, so it's uniform now) we can no longer pass different base URL into the function. I think the sub tests are so trivial that it's ok to remove them. It's not like we implemented the joining of two URLs ourselves but we use the method from the URL crate. So why test what they already tested?

services/src/handlers/wms.rs
michaelmattig: the only thing that can be buggy using the wrong URL as a base and this wasn't actually covered by the tests at all. They merely checked whether the joining of the URL base with a path works and that's not even implemented by us but by the URL crate

ChristianBeilschmidt: cf. above

michaelmattig: the logic of joining the url did not change, it's just that now the binding address is used if there is no external_address given. I could extract the method for joining the base with /wms/{id} but as I said that's trivial and already tested in the url crate

> Doesn't this remove everything from the root and then joins it?

no it does not. From the docs:

```   
  /// let base = Url::parse("https://example.net/a/b.html")?;
  /// let url = base.join("c.png")?;
  /// assert_eq!(url.as_str(), "https://example.net/a/c.png");
```

services/src/handlers/wms.rs
michaelmattig: 7057f65bbb07e00a2bfceaf6c54d0e0e02f3cb3a

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #512, opened 2022-03-22
fix add_null in FirstValueAggregator

michaelmattig:


operators/src/processing/raster_vector_join/aggregator.rs
```diff
@@ -155,7 +155,7 @@ where
     }
 
     fn add_null(&mut self, feature_idx: usize) {
-        if !self.null[feature_idx] {
+        if self.pristine[feature_idx] {
             self.null[feature_idx] = true;
             self.number_of_non_null_values -= 1;
```
ChristianBeilschmidt, line 160: Shouldn't this… now… have some effect on the number of pristine values?

operators/src/processing/raster_vector_join/aggregator.rs
```diff
@@ -453,4 +453,17 @@ mod tests {
 
         assert_eq!(aggregator.nulls(), &[true, true]);
     }
+
+    #[test]
+    fn value_then_null() {
+        let mut aggregator = FirstValueIntAggregator::new(1).into_typed();
+
+        aggregator.add_value(0, 1337, 1);
+        aggregator.add_null(0);
```
ChristianBeilschmidt, line 462: I guess the tests need to call more than potentially the number of pristine/null values.
I think this code is still buggy.

And I think the mean variant is too. It all doesn't look that sound :thinking:

ChristianBeilschmidt (changes requested):

operators/src/processing/raster_vector_join/aggregator.rs
michaelmattig: if I understand the aggregator correctly it takes the first value, regardless of whether it is null or not. So the logic for add_value and add_null would be: if the value is pristine, use it as the aggregation result and mark the feature as not pristine. If it is not pristine: ignore the value. I think the aggregator does exactly that

operators/src/processing/raster_vector_join/aggregator.rs
michaelmattig: I fixed the bug where a null value that came after a valid value crashed the server. I fixed that and this tests checks for exactly that.

If there are other bugs, we still have to fin them :D

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #522, opened 2022-04-06
Mock_source_filling

jdroenner:


operators/src/mock/mock_raster_source.rs
```diff
@@ -18,14 +21,28 @@ where
...
-    fn new(data: Vec<RasterTile2D<T>>) -> Self {
-        Self { data }
+    fn new(
+        data: Vec<RasterTile2D<T>>,
+        no_data_value: Option<T>,
+        tiling_specification: TilingSpecification,
+    ) -> Self {
+        // check that all tiles have the size of the used tiling specification
+        data.iter()
+            .for_each(|f| assert_eq!(f.grid_shape(), tiling_specification.grid_shape()));
```
ChristianBeilschmidt, line 39: I guess it is only used in tests, but in fact, you can easily use it in any workflow and then easily let our system panic. So it should be a `Result`.

operators/src/mock/mock_raster_source.rs
```diff
@@ -18,14 +21,28 @@ where
     T: Pixel,
 {
     pub data: Vec<RasterTile2D<T>>,
+    pub no_data_value: Option<T>,
+    pub tiling_specification: TilingSpecification,
```
ChristianBeilschmidt, line 25: Why don't we just use this in (from) the initialize process and fail then?

operators/src/plot/statistics.rs
```diff
@@ -263,7 +271,7 @@ mod tests {
             sources: vec![raster_source].into(),
         };
 
-        let execution_context = MockExecutionContext::test_default();
+        let execution_context = MockExecutionContext::new_with_tiling_spec(tiling_specification);
```
ChristianBeilschmidt, line 274: It's correct to do this. If we have to replace every test default, the default is… not a default anymore.

ChristianBeilschmidt (changes requested):

operators/src/mock/mock_raster_source.rs
jdroenner: we get the tiling_spec from the operator when it is initialized

operators/src/plot/statistics.rs
jdroenner: yep but there are also tests that use the current default. i see no good solution here.

services/src/handlers/plots.rs
```diff
@@ -348,7 +360,7 @@ mod tests {
             .unwrap();
 
         let params = &[
-            ("bbox", "-180,-90,180,90"),
+            ("bbox", "0,-0.3,0.2,0"),
```
ChristianBeilschmidt, line 363: Couldn't we have a spec in such a way, that -180 to 180 reflects in the same 6 pixels than 0 to 0.2?

ChristianBeilschmidt (changes requested):

services/src/handlers/plots.rs
jdroenner: no a) our reference is 0.0 so there must be an equal count of pixels on each axis for -180, 180. b) the query is designed with 3 pixels on the y axis and 2 on the x axis. The resolution to query would be y: 60, x: 90. c) this is a query that diverges a lot from the original one. I tried to keep them as designed by the authors.

PR #523, opened 2022-04-10
refactor to custom datetime

ChristianBeilschmidt:


ChristianBeilschmidt:
[Chrono durch Time austauschen](https://trello.com/c/6J8tFLHT/448-chrono-durch-time-austauschen)

datatypes/src/primitives/datetime.rs
```diff
@@ -0,0 +1,668 @@
...
+use snafu::Snafu;
+use std::cmp::Ordering;
+use std::ops::{Add, Sub};
+use std::str::FromStr;
+
+/// An object that composes the date and a timestamp with time zone.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct DateTime {
+    datetime: NaiveDateTime,
+    offset: UtcOffset,
```
michaelmattig, line 14: internally we always work with UTC datetimes/TimeInstance. The only point where we need to handle timezones is during parsing/serializing. I think it would make sense to have separate types. One DateTime type for the whole TimeStep stuff that is always UTC and one DateTime type with timezones for in/output to/from the user

datatypes/src/primitives/datetime.rs
```diff
@@ -0,0 +1,668 @@
...
+    pub fn parse_from_rfc3339(input: &str) -> Result<Self, DateTimeError> {
+        let date_time = chrono::DateTime::<chrono::FixedOffset>::parse_from_rfc3339(input)
+            .map_err(|e| DateTimeError::DateParse {
+                source: Box::new(e),
+            })?;
+
+        Ok(date_time.into())
+    }
+
+    // TODO: is this inprecise?
```
michaelmattig, line 201: what does the precision refer to? format or temporal resolution?

datatypes/src/primitives/datetime.rs
```diff
@@ -0,0 +1,668 @@
...
+        chrono_date_time.to_rfc3339()
+    }
+
+    /// Now in UTC.
+    pub fn now() -> Self {
+        chrono::offset::Utc::now().into()
+    }
+
+    pub fn day_of_year(&self) -> u16 {
+        // TODO: calculate offset?
```
michaelmattig, line 230: what offset?

datatypes/src/primitives/datetime.rs
```diff
@@ -0,0 +1,668 @@
...
+    pub fn with_year(&self, year: i32) -> Option<Self> {
+        Some(Self {
+            datetime: self.datetime.with_year(year)?,
+            offset: self.offset,
+        })
+    }
+}
+
+impl UtcOffset {
+    fn _to_fixed_offset(self) -> chrono::FixedOffset {
```
michaelmattig, line 274: why do the names start with an underscore?

datatypes/src/primitives/datetime.rs
```diff
@@ -0,0 +1,668 @@
...
+    Ymd,
+    Unix,
+    Custom {
+        fmt: String,
+        has_tz: bool,
+        has_time: bool,
+    },
+}
+
+impl DateTimeParseFormat {
```
michaelmattig, line 388: We need to adjust the datasets definitions and extend the documentation if we introduce a new custom parseformat

datatypes/src/primitives/time_instance.rs
```diff
@@ -60,7 +54,13 @@ impl TimeInstance {
     }
 
     pub fn now() -> Self {
-        Self::from(chrono::offset::Utc::now())
+        Self::from(DateTime::now())
+    }
+
+    /// Converts a `TimeInstance` to a `DateTime`.
```
michaelmattig, line 60: shouldn't this return a Result instead of an Option? you convert it into a result in TimeStep anyway

datatypes/src/error.rs
```diff
@@ -11,6 +11,20 @@ use crate::{
     raster::RasterDataType,
 };
 
+pub trait ErrorSource: std::error::Error + Send + Sync + 'static + AsErrorSource {
+    // fn boxed(self) -> Box<Self>;
+}
+
+// pub trait SizedErrorSource: ErrorSource + Sized {}
```
michaelmattig, line 18: remove

datatypes/src/error.rs
```diff
@@ -11,6 +11,20 @@ use crate::{
...
+
+impl<T> ErrorSource for T where T: std::error::Error + Send + Sync + 'static {}
+
+// impl<T> SizedErrorSource for T where T: ErrorSource + Sized {}
```
michaelmattig, line 22: remove

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: We could use it for computations like: Give me all points from noon wherever your noon is at.
Because in UTC time, it is like. Well, for you it is noon, but we treat it being in the middle of the night.

I guess the other problem is, that Postgres does not store it this way.
So we should discuss the use case I guess briefly in the team.
Currently, we have just ignored it and used the library as is. But now it could be of interest :shrug:

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: We just found that in chrono `from_str` parses stuff with slightly more slack that rfcXYZ.
For instance, we allow negative date times.
But maybe we need just a definition of what we exactly want and then implement that?

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: I essentially want to prevent that no one just writes pub before that and uses that externally.

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: The reason I did not store a `DateTime<FixedOffset>` directly was for `chrono`, that `MIN` and `MAX` can no longer be `const`. But I guess for day of year the timezone matters?…

Moreover, maybe they don't have to be `const`… idk

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: Yes, unfortunately the old stuff works because %Y still gets passed to chrono and it works… Ideas for that?

datatypes/src/primitives/time_instance.rs
ChristianBeilschmidt: But isn't the Error for `TimeStep` more specific that this one? It essentially `None`s if it overflows.

datatypes/src/error.rs
ChristianBeilschmidt: [02b8676](https://github.com/geo-engine/geoengine/pull/523/commits/02b8676948adba13bcf7b18ac3a17157da225923)

datatypes/src/error.rs
ChristianBeilschmidt: [02b8676](https://github.com/geo-engine/geoengine/pull/523/commits/02b8676948adba13bcf7b18ac3a17157da225923)

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: [9b18fc7](https://github.com/geo-engine/geoengine/pull/523/commits/9b18fc7ba1c2a7fc0fbe2882cec07a2f7a2b1108)

[003ff00](https://github.com/geo-engine/geoengine/pull/523/commits/003ff00ac566c68568deb8c4c306f29e3646ad46)

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: [6cc01e7](https://github.com/geo-engine/geoengine/pull/523/commits/6cc01e7e8c81f7a5e6eda1570bc2b3ab8fdecc0e)

datatypes/src/primitives/datetime.rs
```diff
@@ -0,0 +1,638 @@
...
+    }
+}
+
+impl From<&DateTime> for chrono::DateTime<chrono::FixedOffset> {
+    fn from(datetime: &DateTime) -> Self {
+        datetime.datetime.into()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
```
michaelmattig, line 306: why is this not Deserialize? How is this constructed, aside from tests?

datatypes/src/primitives/datetime.rs
```diff
@@ -0,0 +1,638 @@
...
+pub struct DateTimeParseFormat {
+    fmt: String,
+    chrono_fmt: String,
```
michaelmattig, line 309: if this is for user inputs it should be independent of chrono I guess

datatypes/src/primitives/datetime.rs
michaelmattig: then we have to adjust all the dataset definitions?

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: There are custom impls

datatypes/src/primitives/datetime.rs
```diff
@@ -0,0 +1,668 @@
...
+
+                Ok(Self {
+                    datetime,
+                    offset: UTC,
+                })
+            }
+        }
+    }
+
+    pub fn parse_from_rfc3339(input: &str) -> Result<Self, DateTimeError> {
```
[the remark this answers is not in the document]
ChristianBeilschmidt: This is `Display` and `FromStr` now

datatypes/src/primitives/datetime.rs
ChristianBeilschmidt: We could do it, but we first need to be sure how we want it. Is is okay this way? Will we allow modifiers? Do we need to escape the chrono stuff so that only our Placeholders are really used?

michaelmattig: idk, maybe we have to discuss it. Your format is certainly more verbose than what we had before

ChristianBeilschmidt: [13762a7](https://github.com/geo-engine/geoengine/pull/523/commits/13762a70276d19b757a3de57880c813c2345afb8)

michaelmattig (approved):

michaelmattig:
bors merge

PR #530, opened 2022-04-25
Scale operator

jdroenner:


jdroenner:
[Offset/Scale aus Raster Metadaten anwenden](https://trello.com/c/9RKXaqVf/743-offset-scale-aus-raster-metadaten-anwenden)

datatypes/src/raster/operations/map_elements.rs
```diff
@@ -0,0 +1,364 @@
+use crate::raster::{
+    data_type::DefaultNoDataValue, EmptyGrid, Grid, GridOrEmpty, GridSize, NoDataValue,
+    RasterTile2D,
+};
+use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
+
+pub trait MapElements<In, Out, F: Fn(In) -> Option<Out>> {
+    type Output;
+    fn map_elements(self, map_fn: F, out_no_data: Option<Out>) -> Self::Output;
```
michaelmattig, line 9: does it make sense to call the function for every pixel or would it make more sense to have a function that maps all pixels at once?

datatypes/src/raster/operations/map_elements.rs
```diff
@@ -0,0 +1,364 @@
...
+        let out_no_data = out_no_data.unwrap_or(Out::DEFAULT_NO_DATA_VALUE);
+        let shape = self.shape.clone();
+        let data = self
+            .data
+            .iter()
+            .map(|x| {
+                if self.is_no_data(*x) {
+                    out_no_data
+                } else {
+                    map_fn(*x).unwrap_or(out_no_data)
```
michaelmattig, line 35: is this always meaningful or do we need a option (parameter) to return an error?

datatypes/src/raster/operations/scaling.rs
```diff
@@ -0,0 +1,631 @@
+use crate::raster::{
+    data_type::DefaultNoDataValue, EmptyGrid, Grid, GridOrEmpty, GridOrEmpty2D, GridSize,
+    MapElements, MapElementsParallel, RasterTile2D,
+};
+use num_traits::{CheckedAdd, CheckedDiv, CheckedMul, CheckedSub};
+use std::ops::{Add, Div, Mul, Sub};
+
+/// scales the value with `(value - offset_by) / scale_with`.
```
michaelmattig, line 8: I would have expected value * scale + offset tbh. wouldn't this here be more like "unscale" or normalize=

datatypes/src/raster/operations/scaling.rs
```diff
@@ -0,0 +1,631 @@
...
+    fn unscale_elements(self, scale_with: P, offset_by: P, out_no_data: P) -> Self::Output {
+        match self {
+            GridOrEmpty::Grid(g) => {
+                GridOrEmpty::Grid(g.unscale_elements(scale_with, offset_by, out_no_data))
+            }
+            GridOrEmpty::Empty(e) => GridOrEmpty::Empty(EmptyGrid::new(
+                e.shape,
+                e.no_data_value
+                    .unscale(scale_with, offset_by)
+                    .unwrap_or(P::DEFAULT_NO_DATA_VALUE),
```
michaelmattig, line 252: is this always meaningful or are there scenarios where we want to propagate this as an error?

datatypes/src/raster/data_type.rs
```diff
@@ -291,6 +292,51 @@ where
     }
 }
 
+pub trait DefaultNoDataValue {
+    /// The value that is used as default to represent no data for a data type
+    const DEFAULT_NO_DATA_VALUE: Self;
+}
+
+impl DefaultNoDataValue for u8 {
+    const DEFAULT_NO_DATA_VALUE: Self = Self::MAX;
```
michaelmattig, line 301: why not zero?

datatypes/src/raster/raster_properties.rs
```diff
@@ -21,10 +21,51 @@ impl RasterProperties {
         self.properties_map.keys().map(|m| m.domain.as_ref())
     }
 
+    pub fn set_scale(&mut self, scale: f64) {
+        self.scale = Some(scale);
```
michaelmattig, line 25: btw is this the value this was scaled with or is this the value we have to scale the pixels with to get measurements?

datatypes/src/raster/raster_properties.rs
```diff
@@ -34,9 +75,14 @@ impl RasterProperties {
     }
 
     pub fn string_property(&self, key: &RasterPropertiesKey) -> Result<String> {
+        if key.domain.is_none() && key.key == "band_name" {
```
michaelmattig, line 78: why is this the way it is, what does this mean? :D

operators/src/processing/raster_scaling.rs
```diff
@@ -0,0 +1,358 @@
...
+    ) -> Result<RasterTile2D<P>> {
+        let offset_by = Self::prop_value(&self.offset_by, &tile.properties)?;
+        let scale_with = Self::prop_value(&self.scale_with, &tile.properties)?;
+        let scaling_mode = self.scaling_mode;
+
+        let no_data_value = self.no_data_value;
+
+        let res_tile =
+            crate::util::spawn_blocking_with_thread_pool(pool, move || match scaling_mode {
+                ScalingMode::Scale => {
```
michaelmattig, line 185: could maybe also link this statically by making the processor generic over some T with a static scale/unscale method.

operators/src/processing/raster_scaling.rs
```diff
@@ -0,0 +1,358 @@
...
+        };
+
+        let stream = typed_processor
+            .raster_query(query, &query_ctx)
+            .await
+            .unwrap();
+
+        let results = stream.collect::<Vec<Result<RasterTile2D<u8>>>>().await;
+
+        // assert_eq!((&results).len(), 1);
```
michaelmattig, line 343: remove

operators/src/processing/raster_type_conversion.rs
```diff
@@ -0,0 +1,237 @@
...
+
+pub struct InitializedRasterTypeConversionOperator {
+    result_descriptor: RasterResultDescriptor,
+    source: Box<dyn InitializedRasterOperator>,
+}
+
+#[typetag::serde]
+#[async_trait]
+impl RasterOperator for RasterTypeConversionOperator {
+    async fn initialize(
```
michaelmattig, line 33: - should we ensure/enforce in and out datatype are different?
- do we want to throw errors if values are lost during conversion from larger to smaller datatype?

michaelmattig (changes requested):

datatypes/src/raster/operations/map_elements.rs
jdroenner: the idea was that this trait takes a fn and applies it to all pixels. so you only have to provide a fn for each pixel.

datatypes/src/raster/operations/map_elements.rs
jdroenner: i guess there might be a case where we need to return errors but we could add that later

datatypes/src/raster/operations/scaling.rs
jdroenner: i used the GDAL terms from: https://gdal.org/programs/gdal_translate.html

datatypes/src/raster/data_type.rs
jdroenner: i guess all choices are wrong so i can change it to zero if you like.

datatypes/src/raster/operations/scaling.rs
jdroenner: in this case i thing out of bounds values should cenvert to nodata

datatypes/src/raster/raster_properties.rs
jdroenner: it is the value the data was scaled with

datatypes/src/raster/raster_properties.rs
jdroenner: the key is a pair and there are fixed value fields in the properties. To be able to address them i added this. we could also remove the fixed keys

operators/src/processing/raster_type_conversion.rs
jdroenner: * it should be a no-op. I think we can add that when specialization is available. 
* i thought of this but it will cause a lot of "not real" errors. Users should be aware that out of bounds will turn to nodata (or clamp later?)

operators/src/processing/raster_scaling.rs
jdroenner: [d2194b3](https://github.com/geo-engine/geoengine/pull/530/commits/d2194b3637fd0e81c0613db4a2b3bf369c0e6518)

operators/src/processing/raster_scaling.rs
jdroenner: [f0e8659](https://github.com/geo-engine/geoengine/pull/530/commits/f0e86596f91cdee8832675c8fa8f4c7546aede1a)

datatypes/src/raster/operations/scaling.rs
jdroenner: [0f5694b](https://github.com/geo-engine/geoengine/pull/530/commits/0f5694b8b5b543dd192ecd08bf9651737e2f8afd)

datatypes/src/raster/operations/scaling.rs
michaelmattig: On the page it says

> -unscale[](https://gdal.org/programs/gdal_translate.html#cmdoption-gdal_translate-unscale)
>
>    Apply the scale/offset metadata for the bands to convert scaled values to unscaled values. It is also often necessary to reset the output datatype with the [-ot](https://gdal.org/programs/gdal_translate.html#cmdoption-gdal_translate-ot) switch.

isn't that what you're doing in your `scale` method? That's why I thought it should be called unscaled, but maybe I'm looking at it wrong.

datatypes/src/raster/raster_properties.rs
michaelmattig: maybe add some docs somewhere or refer to the Gdal docs?

datatypes/src/raster/operations/scaling.rs
jdroenner: For GDAL scaled data is not the real data. I also think that is easy to misunderstand. See the "-scale" option in gdal_edit: https://gdal.org/programs/gdal_edit.html#cmdoption-scale


"Scale and Offset are generally used together. For example, scale and offset might be used to store elevations in a unsigned 16bit integer file with a precision of 0.1, and starting from -100. True values would be calculated as: true_value = (pixel_value * scale) + offset"

datatypes/src/raster/raster_properties.rs
jdroenner: [ad30cb7](https://github.com/geo-engine/geoengine/pull/530/commits/ad30cb7840fa6e361f40e368a88b5148498b7d56)

michaelmattig (approved):

michaelmattig:
bors merge

jdroenner:
bors retry

PR #536, opened 2022-05-10
interpolation operator

michaelmattig:


operators/src/processing/interpolation/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+{
+    type Output = RasterTile2D<P>;
+    type SpatialBounds = SpatialPartition2D;
+
+    async fn query<'a>(
+        &'a self,
+        query: RasterQueryRectangle,
+        ctx: &'a dyn QueryContext,
+    ) -> Result<BoxStream<'a, Result<Self::Output>>> {
+        // TODO: check input resolution < output resolution
```
jdroenner, line 156: please add that.

operators/src/processing/interpolation/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+    tiling_specification: TilingSpecification,
+) -> impl Future<Output = Result<InterpolationAccu<T, I>>> {
+    // create an accumulator as a single tile that fits all the input tiles
+    crate::util::spawn_blocking(move || {
+        let tiling = tiling_specification.strategy(
+            query_rect.spatial_resolution.x,
+            -query_rect.spatial_resolution.y,
+        );
+
+        // TODO: enlarge query_rect to be sure to have all the neighbor pixels for the edges
```
jdroenner, line 306: is that still needed?

operators/src/processing/interpolation/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+where
+    T: Pixel,
+    I: InterpolationAlgorithm<T>,
+{
+    // get the time now because it is not known when the accu was created
+    accu.output_tile.time = tile.time;
+
+    // copy all input tiles into the accu to have all data for interpolation
+    let mut accu_input_tile = accu.input_tile.into_materialized_tile();
+    accu_input_tile.blit(tile)?;
```
jdroenner, line 386: i see the benefit of using blit here. However my suggestion would be to use a Buffer of 4 / 9 tiles and implement pixel access on that buffer. Then we can avoid the copy operation each blit requres.

operators/src/processing/interpolation/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+
+    Ok(InterpolationAccu::new(
+        accu_input_tile.into(),
+        accu.output_tile,
+        accu.pool,
+    ))
+}
+
+#[async_trait]
+pub trait InterpolationAlgorithm<P: Pixel>: Send + Sync + Clone + 'static {
```
jdroenner, line 396: we should move this to datatypes/raster/operations

jdroenner (changes requested):

operators/src/processing/interpolation/mod.rs
michaelmattig: this would avoid memcopy but destroy data locality as the rows will be split up so I don't think its 100% clear that this is more efficient so I wouldn't prematurely optimize this.

operators/src/processing/interpolation/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+        };
+
+        Ok(RasterSubQueryAdapter::<'a, P, _, _>::new(
+            &self.source,
+            query,
+            self.tiling_specification,
+            ctx,
+            sub_query,
+        )
+        .filter_and_fill(self.no_data_value)) // TODO: check if this is needed
```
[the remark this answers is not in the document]
michaelmattig: ok then let's keep it

operators/src/processing/interpolation/mod.rs
michaelmattig: I can move it, but then I would need to move the util/rayon.rs file as well because it's required for testing. Shall I do it?

operators/src/processing/interpolation/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+
+#[async_trait]
+pub trait InterpolationAlgorithm<P: Pixel>: Send + Sync + Clone + 'static {
+    /// interpolate the given input tile into the output tile
+    /// the output must be fully contained in the input tile and have an additional row and column in order
+    /// to have all the required neighbor pixels.
+    /// Also the output must have a finer resolution than the input
+    fn interpolate(input: &RasterTile2D<P>, output: &mut MaterializedRasterTile2D<P>)
+        -> Result<()>;
+}
```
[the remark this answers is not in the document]
michaelmattig: it should be possible to add it, but not "simply" because the bicubic interpolation needs 16 neighbor pixels instead of four, so we would need to make the creation of the accumulator dependent on the selected interpolation algorithm and make the query rectangle even larger for bicubic.

operators/src/processing/temporal_raster_aggregation/mean_aggregation_subquery.rs
```diff
@@ -125,13 +126,14 @@ impl<T> TemporalMeanTileAccu<T> {
...
 
+#[async_trait]
 impl<T> FoldTileAccu for TemporalMeanTileAccu<T>
 where
     T: Pixel,
 {
     type RasterType = T;
 
-    fn into_tile(self) -> RasterTile2D<Self::RasterType> {
+    async fn into_tile(self) -> Result<RasterTile2D<Self::RasterType>> {
```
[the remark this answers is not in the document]
michaelmattig: not necessarily. I do spawn a thread in the interpolation operator but it could as well perform some I/O operation or something else that can be `awaited`. I guess we could simplify some code in the subquery adapter by using async_trait instead of methods that produce Futures too.

operators/src/processing/interpolation/mod.rs
michaelmattig: 9fba3541e7ffdc6b094d5a270c28ddcc36909909

operators/src/processing/interpolation/mod.rs
michaelmattig: I removed it 9fba3541e7ffdc6b094d5a270c28ddcc36909909

operators/src/processing/interpolation/mod.rs
jdroenner: i think for the test you can simply create a rayon threadpool or use the global one.

michaelmattig: 7dea89ecba518bdd1f8afab5983e5d6abe1686f7

jdroenner (approved):

jdroenner:
bors merge

PR #537, opened 2022-05-11
fix blit creates lines

jdroenner:


jdroenner:
["Leere Linie" in projizierten Kacheln.](https://trello.com/c/LVFJsFVk/716-leere-linie-in-projizierten-kacheln)

datatypes/src/raster/geo_transform.rs
```diff
@@ -175,6 +154,61 @@ impl GeoTransform {
         }
     }
 
+    /// compute the index of the upper left pixel that is contained in the `partition`
+    pub fn upper_left_pixel_idx(&self, partition: &SpatialPartition2D) -> GridIdx2D {
+        // choose the epsilon relative to the pixel size
+        const EPSILON: f64 = 0.000_001;
```
ChristianBeilschmidt, line 160: What is this and does this work for all conditions?
Maybe we need a test for that for different resolutions.

datatypes/src/raster/geo_transform.rs
```diff
@@ -175,6 +154,61 @@ impl GeoTransform {
...
+        let epsilon: Coordinate2D =
+            (self.x_pixel_size() * EPSILON, self.y_pixel_size() * EPSILON).into();
+
+        let upper_left_coordinate = partition.upper_left() + epsilon;
```
ChristianBeilschmidt, line 164: Why should we do this?
I this is reasonable, please add a comment.

datatypes/src/raster/geo_transform.rs
```diff
@@ -175,6 +154,61 @@ impl GeoTransform {
...
+    }
+
+    /// compute the index of the lower right pixel that is contained in the `partition`
+    pub fn lower_right_pixel_idx(&self, partition: &SpatialPartition2D) -> GridIdx2D {
+        // as the lower right coordinate is not included in the partition we subtract an epsilon
+        // in order to not include the next pixel if the lower right coordinate is exactly on the
+        // edge of the next pixel
+
+        // choose the epsilon relative to the pixel size
+        const EPSILON: f64 = 0.000_001;
```
ChristianBeilschmidt, line 176: Fails if the pixel size is less than this, right?

datatypes/src/raster/geo_transform.rs
```diff
@@ -175,6 +154,61 @@ impl GeoTransform {
...
+        self.coordinate_to_grid_idx_2d(lower_right)
+    }
+
+    /// Transform a `SpatialPartition2D` into a `GridBoundingBox`
+    #[inline]
+    pub fn spatial_to_grid_bounds(
+        &self,
+        spatial_partition: &SpatialPartition2D,
+    ) -> GridBoundingBox2D {
+        //let snapped = spatial_partition.snap_to_grid(self.origin_coordinate, self.spatial_resolution());
```
ChristianBeilschmidt, line 199:
```suggestion
```

datatypes/src/raster/geo_transform.rs
```diff
@@ -380,6 +414,40 @@ mod tests {
...
+            geo_transform.lower_right_pixel_idx(&partition),
+            [8, 7].into()
+        );
+    }
+
+    #[test]
+    fn lower_right_pixel_index_inside() {
+        let geo_transform = GeoTransform::test_default();
+
+        let partition = SpatialPartition2D::new((1., 1.).into(), (7.5, -7.5).into()).unwrap();
```
ChristianBeilschmidt, line 444: We need some much smaller numbers wrt. the epsilon.

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
```diff
@@ -88,12 +88,17 @@ where
...
-            }))
+            let projected_bounds = bounds.reproject(&proj);
+
+            match projected_bounds {
+                Ok(pb) => Ok(Some(RasterQueryRectangle {
+                    spatial_bounds: pb,
+                    time_interval: TimeInterval::new_instant(start_time)?,
+                    spatial_resolution: self.in_spatial_res,
+                })),
+                Err(geoengine_datatypes::error::Error::OutputBboxEmpty { bbox: _ }) => Ok(None), // in some strange cases the reprojection can return an empty bbox. YWe ignore it since it contains no pixels.
```
ChristianBeilschmidt, line 99:
```suggestion
                // In some strange cases the reprojection can return an empty box.
                // We ignore it since it contains no pixels.
                Err(geoengine_datatypes::error::Error::OutputBboxEmpty { bbox: _ }) => Ok(None),
```

ChristianBeilschmidt (changes requested):

datatypes/src/raster/geo_transform.rs
jdroenner: i use it to move the Coordinate EPSILON * pixel_size to the reight/bottom. This solves the effect that a coordinate might be a µ pixel_size smaller that the border of the nearest pixel. We do the same thing for the lower right pixel of the box.
It should work in all conditions if the assumption that a coordinate that is a µ smaller then the border of a pixel should belong to that pixel. If that assumption does not hold then the calculation for the lower right is also wrong.

datatypes/src/raster/geo_transform.rs
jdroenner: see comment above

datatypes/src/raster/geo_transform.rs
jdroenner: no it is multiplied with the resolution in the next step. this is not new i just moved it from tiling_strategy

datatypes/src/raster/geo_transform.rs
jdroenner: the epsilon is a factor that is multiplied with the pixel_size. This is not new. I moved it from tiling_strategy

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #551, opened 2022-06-02
add window reference time as anchor point

michaelmattig:


operators/src/processing/temporal_raster_aggregation/temporal_aggregation_operator.rs
```diff
@@ -34,7 +34,9 @@ use super::min_max_first_last_subquery::{
 pub struct TemporalRasterAggregationParameters {
     aggregation: Aggregation,
     window: TimeStep,
-    // TODO: allow specifying window start instead of using query.start?
+    /// Define an anchor point for `window`
+    /// If `None`, the anchor point is `1970-01-01T00:00:00Z` by default
+    window_reference: Option<TimeInstance>,
```
michaelmattig, line 39: warning: this changes the semantics of the operator! Currently, the operator always starts aggregating at the start of the query time but now, if no window_reference is given, the exact same query will return a different query because we anchor it at1970-1-1. I designed it this way, because (1) I think it makes more sense, by default you want to align to full months and full days etc. and (2) the time projection for vectors works the same way and I think we should make it uniform.

jdroenner: yes i think that is very reasonable

operators/src/processing/temporal_raster_aggregation/temporal_aggregation_operator.rs
```diff
@@ -75,6 +77,10 @@ impl RasterOperator for TemporalRasterAggregation {
         let initialized_operator = InitializedTemporalRasterAggregation {
             aggregation_type: self.params.aggregation,
             window: self.params.window,
+            window_reference: self
+                .params
+                .window_reference
+                .unwrap_or_else(|| TimeInstance::from_millis_unchecked(0)),
```
jdroenner, line 83: we should add a constant for start of epoch.

jdroenner (changes requested):

operators/src/processing/temporal_raster_aggregation/temporal_aggregation_operator.rs
michaelmattig: 59341e43282f2e905b107ed01af3052be0bc683b

jdroenner (approved):

PR #554, opened 2022-06-15
Unify Layer listing

michaelmattig:


michaelmattig:
[LayerCollection für dynamische Workflows erweitern und ExternalDatasetProviders anpassen](https://trello.com/c/IpMFsNxb/829-layercollection-f%C3%BCr-dynamische-workflows-erweitern-und-externaldatasetproviders-anpassen)

services/src/layers/storage.rs
```diff
@@ -19,7 +21,7 @@ pub enum LayerDbError {
 }
 
 #[async_trait]
-pub trait LayerDb: Send + Sync {
+pub trait LayerDb: LayerCollectionProvider + Send + Sync {
```
ChristianBeilschmidt, line 24: Why is it good to move the two methods into a separate trait?
Isn't this trait implemented for every `LayerDb` and solely for `LayerDb`s?

services/src/layers/storage.rs
```diff
@@ -285,12 +293,21 @@ mod tests {
     async fn it_stores_layers() -> Result<()> {
         let db = HashMapLayerDb::default();
 
-        let workflow_id = WorkflowId::new();
+        let _workflow_id = WorkflowId::new();
```
ChristianBeilschmidt, line 296: remove?

services/src/pro/layers/postgres_layer_db.rs
```diff
@@ -74,7 +76,7 @@ where
                 &id,
                 &layer.name,
                 &layer.description,
-                &layer.workflow,
+                &serde_json::to_value(&layer.workflow).context(error::SerdeJson)?,
```
ChristianBeilschmidt, line 79: didn't you implement a ToSQL trait for the other types?

services/src/pro/layers/postgres_layer_db.rs
```diff
@@ -103,7 +105,7 @@ where
...
```
ChristianBeilschmidt, line 108: same here

services/src/pro/layers/postgres_layer_db.rs
```diff
@@ -137,7 +139,7 @@ where
             id,
             name: row.get(0),
             description: row.get(1),
-            workflow: row.get(2),
+            workflow: serde_json::from_value(row.get(2)).context(error::SerdeJson)?,
```
ChristianBeilschmidt, line 142: and here

services/src/workflows/workflow.rs
```diff
@@ -23,6 +23,15 @@ pub struct Workflow {
...
 }
 
+impl PartialEq for Workflow {
+    fn eq(&self, other: &Self) -> bool {
+        match (serde_json::to_string(self), serde_json::to_string(other)) {
+            (Ok(a), Ok(b)) => a == b,
+            _ => false,
+        }
+    }
+}
```
ChristianBeilschmidt, line 33: Hm, maybe we should have our operator spec `PartialEq`?

ChristianBeilschmidt (changes requested):

services/src/layers/storage.rs
michaelmattig: no, the LayerDB stores layers (inside Geo Engine). The LayerCollectionProviders provides access to layers. And they can be stored inside Geo Engine or be server from an external provider.

services/src/pro/layers/postgres_layer_db.rs
michaelmattig: no, we store complex values as json because we said it'S too much work to try to map the rust structs to SQL

services/src/layers/add_from_directory.rs
```diff
@@ -91,10 +83,12 @@ pub async fn add_layer_collections_from_directory<L: LayerDb>(db: &mut L, file_p
         }
         .validated()?;
 
-        db.add_collection_with_id(def.id, collection).await?;
+        // TODO: add only collections that aren't contained in any other collection to the root collection?
```
jdroenner, line 86: could we add a "leftover" or "unsorted aka. your desktop"? collection for this?

services/src/layers/external.rs
```diff
@@ -0,0 +1,73 @@
...
+pub trait ExternalLayerProvider: LayerCollectionProvider
+    + MetaDataProvider<MockDatasetDataSourceLoadingInfo, VectorResultDescriptor, VectorQueryRectangle>
+    + MetaDataProvider<OgrSourceDataset, VectorResultDescriptor, VectorQueryRectangle>
+    + MetaDataProvider<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
+    + Send
+    + Sync
+    + std::fmt::Debug
+{
+    // TODO: datasetId should be named something else because there are not external datasets anymore, only external layers
+    // TODO: rename trait ProvenanceProvider or smth?
```
jdroenner, line 68: good idea

services/src/layers/external.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+impl Clone for Box<dyn ExternalLayerProviderDefinition> {
+    fn clone(&self) -> Box<dyn ExternalLayerProviderDefinition> {
+        self.clone_boxed_provider()
+    }
+}
+
+/// A provider of layers that are not hosted by Geo Engine itself but some external party
+// TODO: Authorization: the provider needs to accept credentials for the external data source.
```
jdroenner, line 53: how will we handover credentials to the provider? i guess they need to be passed at some point?

services/src/layers/storage.rs
```diff
@@ -18,47 +25,99 @@ pub enum LayerDbError {
...
+    /// add new `layer` to the given `collection`
+    async fn add_layer(
+        &self,
+        layer: Validated<AddLayer>,
+        collection: &LayerCollectionId,
+    ) -> Result<LayerId>;
 
-    async fn get_layer(&self, id: LayerId) -> Result<Layer>;
+    /// add new `layer` with fixed `id` to the given `collection`
+    /// TODO: remove this method and allow stable names instead
```
jdroenner, line 45: what is this todo about?

services/src/pro/contexts/postgres.rs
```diff
@@ -105,29 +108,28 @@ where
...
+        add_providers_from_directory(&mut layer_provider_db, provider_defs_path.clone()).await;
+        add_providers_from_directory(&mut layer_provider_db, provider_defs_path.join("pro")).await;
 
         Ok(Self {
             user_db: Arc::new(PostgresUserDb::new(pool.clone())),
             project_db: Arc::new(PostgresProjectDb::new(pool.clone())),
             workflow_registry: Arc::new(workflow_db),
             dataset_db: Arc::new(dataset_db),
             layer_db: Arc::new(layer_db),
+            layer_provider_db: Arc::new(PostgresLayerProviderDb::new(pool.clone())),
```
jdroenner, line 132: why do we need layer_db, layer_provider_db and dataset_db?

jdroenner (commented):
i think i need another explanation of all the DBs

services/src/datasets/add_from_directory.rs
```diff
@@ -169,18 +61,16 @@ pub async fn add_datasets_from_directory<
     }
 }
 
-pub async fn add_providers_from_directory<D: DatasetDb<S>, S: MockableSession>(
-    db: &mut D,
-    file_path: PathBuf,
-) {
-    async fn add_provider_definition_from_dir_entry<D: DatasetDb<S>, S: MockableSession>(
+// TODO: move to layers source dir
```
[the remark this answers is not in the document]
michaelmattig: I meant moving this to src/layers because they are now layer providers and no longer dataset providers

services/src/layers/external.rs
michaelmattig: I think we should implement it in such a way, that the Context injects the user credentials into the Db (or a wrapper) s.th. we don't have to explicitly pass the session into the functions every time, because the Db already knows it. This makes it much easier as we don't need a generic for the session in the traits. However, IMHO, this should be done in a different task where we do that with all the Dbs.

services/src/layers/storage.rs
michaelmattig: right now we often have two methods to add something (a dataset, a layer, a provider) to the Db. One, where the Id of the newly created element is automatically assigned by the Db (as it should be) and one, where we pass the Id from the outside (because we need it to always be the same). This comment is about removing the latter one once when have stable names for items, that make stable ids obsolete.

services/src/layers/add_from_directory.rs
michaelmattig: ddfafc8adfbc119ea5a6f0c02257dae460ce2468

services/src/layers/external.rs
jdroenner: i think the naming is fine. Can you add a task for the unification of provenance?

jdroenner (approved):

jdroenner (approved):

jdroenner:
bors merge

jdroenner:
bors retry

jdroenner:
bors retry

PR #555, opened 2022-06-16
Tasks

ChristianBeilschmidt:
Create a simple task db in our `Context`.

- [x] Task and TaskDb trait
- [x] InMemory TaskDb
- [x] EBV overviews as tasks
- [ ] Postgres TaskDb
- [ ] Pro versions of TaskDbs


ChristianBeilschmidt:
[[GEO BON] Task-Runner für EBV-Overviews](https://trello.com/c/QliRig6m/850-geo-bon-task-runner-f%C3%BCr-ebv-overviews)

services/src/contexts/mod.rs
```diff
@@ -41,6 +42,8 @@ pub trait Context: 'static + Send + Sync + Clone {
     type LayerDB: LayerDb;
     type QueryContext: QueryContext;
     type ExecutionContext: ExecutionContext;
+    type TaskContext: TaskContext;
+    type TaskDb: TaskDb<Self::TaskContext>;
```
michaelmattig, line 46: it's not really a DB because we never want to persist the tasks right? Maybe we should call it differently? TaskManager?

services/src/tasks/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+use std::{fmt, sync::Arc};
+
+/// A database that allows registering and retrieving tasks.
+#[async_trait::async_trait]
+pub trait TaskDb<C: TaskContext>: Send + Sync {
+    async fn register(&self, task: Box<dyn Task<C>>) -> Result<TaskId, TaskError>;
+
+    async fn status(&self, task_id: TaskId) -> Result<TaskStatus, TaskError>;
+
+    async fn list(&self, options: TaskListOptions) -> Result<Vec<TaskStatusWithId>, TaskError>;
```
michaelmattig, line 18: should be `Validated<TaskListOptions>` for uniformity

services/src/handlers/tasks.rs
```diff
@@ -0,0 +1,271 @@
...
+            .service(web::resource("/list").route(web::get().to(list_handler::<C>)))
+            .service(web::resource("/{task_id}/status").route(web::get().to(status_handler::<C>))),
+    );
+}
+
+/// Create a task somewhere and respond with a task id and a url to query the task status.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct TaskResponse {
+    pub task_id: TaskId,
+    pub url: String,
```
michaelmattig, line 23: I don't like this idea of returning URLs. We do it nowhere else, why here? e.g. we don't create a project and return a URL to access it.

services/src/handlers/tasks.rs
```diff
@@ -0,0 +1,271 @@
...
+
+pub(crate) fn init_task_routes<C>(cfg: &mut web::ServiceConfig)
+where
+    C: Context,
+    C::Session: FromRequest,
+{
+    cfg.service(
+        web::scope("/tasks")
+            .service(web::resource("/list").route(web::get().to(list_handler::<C>)))
+            .service(web::resource("/{task_id}/status").route(web::get().to(status_handler::<C>))),
```
michaelmattig, line 15: the REST url should be singular task: `/task/id/status`

services/src/tasks/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+pub use error::TaskError;
+use geoengine_datatypes::{error::ErrorSource, identifier};
+pub use in_memory::{InMemoryTaskDb, InMemoryTaskDbContext};
+use serde::{Deserialize, Serialize, Serializer};
+use std::{fmt, sync::Arc};
+
+/// A database that allows registering and retrieving tasks.
+#[async_trait::async_trait]
+pub trait TaskDb<C: TaskContext>: Send + Sync {
+    async fn register(&self, task: Box<dyn Task<C>>) -> Result<TaskId, TaskError>;
```
michaelmattig, line 14: As this method also starts the task I think it should reflect that in the name

services/src/tasks/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+pub struct TaskListOptions {
+    #[serde(default)]
+    pub filter: Option<TaskFilter>,
+    #[serde(default)]
+    pub offset: u32,
+    #[serde(default = "task_list_limit_default")]
+    pub limit: u32,
+}
+
+fn task_list_limit_default() -> u32 {
```
michaelmattig, line 131: the other limits we read from the Settings.toml

services/src/tasks/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+pub trait TaskStatusInfo: erased_serde::Serialize + Send + Sync + fmt::Debug {
+    fn boxed(self) -> Box<dyn TaskStatusInfo>
+    where
+        Self: Sized + 'static,
+    {
+        Box::new(self)
+    }
+}
+
+erased_serde::serialize_trait_object!(TaskStatusInfo);
```
michaelmattig, line 116: why can you not use typetag?

services/src/tasks/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl TaskStatusInfo for () {}
+impl TaskStatusInfo for String {}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct TaskListOptions {
```
michaelmattig, line 122: what about `order` to get stable outputs?

services/src/tasks/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// A way to supply status updates from a `Task` to a `TaskDb`.
+#[async_trait::async_trait]
+pub trait TaskContext: Send + Sync {
+    /// Set the completion percentage (%) of the task.
+    /// This is a number between 0 and 100.
+    ///
+    /// Moreover, set a status message.
+    ///
+    async fn set_completion(&self, pct_complete: u8, status: Box<dyn TaskStatusInfo>);
```
michaelmattig, line 44: why is this a trait method on the context? shouldn't the task itself set its own status? Who else could know about its completion?

services/src/tasks/in_memory.rs
```diff
@@ -0,0 +1,150 @@
...
+        );
+        WriteLockAll {
+            handles,
+            tasks_by_id,
+            task_list,
+        }
+    }
+}
+
+struct TaskUptableStatusWithTaskId {
```
michaelmattig, line 40: what is a TaskUptable?

services/src/tasks/mod.rs
ChristianBeilschmidt: I can, but this leads to super weird infos, because every implementation, for instance, just a plain string needs a type definition. Here, it just needs to be serializable.

services/src/tasks/mod.rs
ChristianBeilschmidt: I don't know good ordering options, here. For now, it is stable, it is in the order of insertion.

services/src/tasks/mod.rs
ChristianBeilschmidt: It works this way. The task gets the context and can set the completion… or… do you mean something different?

services/src/handlers/tasks.rs
ChristianBeilschmidt: But it is our endpoint for tasks. Within, you have one taskId… right?
What is the guideline, maybe you can send me something, for this issue and my interest :o.

services/src/tasks/in_memory.rs
ChristianBeilschmidt: sql'scher Versprecher
[dbd850a](https://github.com/geo-engine/geoengine/pull/555/commits/dbd850a1883a97525514005abd31d62db604d8e8)

services/src/tasks/mod.rs
ChristianBeilschmidt: [cfa6775](https://github.com/geo-engine/geoengine/pull/555/commits/cfa6775646267d1e5f1726fc3173297c827a192f)

services/src/contexts/mod.rs
ChristianBeilschmidt: [599847d](https://github.com/geo-engine/geoengine/pull/555/commits/599847d1fafd57a440769b4b041f8a3efec1134f)

services/src/tasks/mod.rs
ChristianBeilschmidt: [905662a](https://github.com/geo-engine/geoengine/pull/555/commits/905662ae2828c81f18a9cde0de83cd8257717d4f)

services/src/handlers/tasks.rs
ChristianBeilschmidt: [ccd3657](https://github.com/geo-engine/geoengine/pull/555/commits/ccd3657e069e1c575890b2033728e91934dc0657)

services/src/tasks/mod.rs
ChristianBeilschmidt: [252d546](https://github.com/geo-engine/geoengine/pull/555/commits/252d546621c44fef01be509095cb247e48b9a7ed)

services/src/contexts/mod.rs
ChristianBeilschmidt: [0940ad4](https://github.com/geo-engine/geoengine/pull/555/commits/0940ad4d915a828e7ac6ef9608c7ff1b8f3e3116)

services/src/handlers/tasks.rs
michaelmattig: I think you are right
https://restfulapi.net/resource-naming/
https://cloud.google.com/apis/design/resource_names

I guess I have to rethink our whole API :D

services/src/tasks/mod.rs
michaelmattig: is "schedule" really better? I mean the tasks are not scheduled for some future execution, but they are started right then, or not? Why not call it run_task or something like that?

services/src/tasks/mod.rs
michaelmattig: the crate says

> This library should be considered a low-level building block for interacting with Serde APIs in an object-safe way. Most use cases will require higher level functionality such as provided by [typetag](https://github.com/dtolnay/typetag) which uses this crate internally.

so I am just wondering whether we want to depend on such a create. How much more work is it to use typetag instead? I have to admit I don't get how it works right now with erased_serde, but I haven't looked really into it either

services/src/tasks/mod.rs
michaelmattig: since when is iterating a hashmap stably sorted?

services/src/handlers/tasks.rs
ChristianBeilschmidt: :muscle:

services/src/tasks/mod.rs
ChristianBeilschmidt: Because in the future, we would also like to postpone running things if the server is packed.
Currently, we just run it directly.

services/src/tasks/mod.rs
ChristianBeilschmidt: We already do since `typetag` is implemented using `erased_serde`. This is why I chose it, to not introduce something really new.

services/src/tasks/mod.rs
ChristianBeilschmidt: I guess you forgot the implementation.

There is the following struct:

```Rust
pub struct SimpleTaskManager {
    handles: Db<HashMap<TaskId, JoinHandle<()>>>,
    tasks_by_id: Db<HashMap<TaskId, Db<TaskStatus>>>,
    task_list: Db<VecDeque<TaskUpdateStatusWithTaskId>>,
}
```

I use `tasks_by_id` for a fast lookup and `task_list` for listing the stuff. And a `VecDeque` has an order.

services/src/tasks/mod.rs
michaelmattig: but this is not apparent from the function name. Maybe it's not important, but I thought it was.

How would it be decided when to run the task anyhow? Will there be priorities or an additional parameter (e.g. enum WHEN { now, time {}, whenever }? Maybe we could introduce that already and allow only the now variant? Maybe also yagni though.

michaelmattig (approved):

michaelmattig:
bors merge

michaelmattig:
bors merge

PR #561, opened 2022-06-21
No data mask

jdroenner:
This PR changes the Geo Engine way of handling pixels in raster data where there is no "data"

First i introduced a new type of grid which stores each pixels value AND a validity mask that indicates if the pixel is valid:
```
#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize)]
pub struct MaskedGrid<D, T> {
    pub inner_grid: Grid<D, T>,
    pub validity_mask: Grid<D, bool>, // TODO: switch to bitmask or something like that (later)
}
```
A MaskedGrid is easy to generate from an EmptyGrid or a simple Grid. It will generate a matching mask. I also changed the GridOrEmpty type we use in tiles to hold an EmptyGrid or a MaskedGrid.

Then i added a new trait to access masked pixels values. It is implemented for EmptyGrid, MaskedGrid, GridOrEmpty, RasterTile. So you can always use that.
```
pub trait MaskedGridIndexAccessMut<T, I> {
    /// Sets a masked value at a grid index. None represents a missing value / no-data.
    ///
    /// # Errors
    ///
    /// The method fails if the grid index is out of bounds.
    ///
    fn set_masked_at_grid_index(&mut self, grid_index: I, value: Option<T>) -> Result<()>;

    /// Sets the value at a grid index
    fn set_masked_at_grid_index_unchecked(&mut self, grid_index: I, value: Option<T>);
}

/// Read elements stored in a Grid
pub trait MaskedGridIndexAccess<T, I> {
    /// Gets a reference to the value at a grid index. None represents a missing value / no-data.
    ///
    /// The method fails if the grid index is out of bounds.
    ///
    fn get_masked_at_grid_index(&self, grid_index: I) -> Result<Option<T>>;

    /// Gets a reference to the value at a grid index
    fn get_masked_at_grid_index_unchecked(&self, grid_index: I) -> Option<T>;
}
```

Then i added a trait for mapping a MaskedGrid into a new one. Use this when only the values are changed. It is implemented for MaskedGrid, GridOrEmpty and RasterTile.
```
pub trait MapMaskedElements<In, Out, F: Fn(Option<In>) -> Option<Out>> {
    type Output;
    fn map_or_mask_elements(self, map_fn: F) -> Self::Output;
}

pub trait MapMaskedElementsParallel<In, Out, F: Fn(Option<In>) -> Option<Out>> {
    type Output;
    fn map_or_mask_elements_parallel(self, map_fn: F) -> Self::Output;
}
```

I also added  a trait to map a pixel with its coordinate. This is very useful for operations like projection and interpolation. It is also implemented for MaskedGrid, GridOrEmpty and RasterTile.
```
pub trait MapIndexedElements<In, Out, GIdx, F: Fn(GIdx, Option<In>) -> Option<Out>> {
    type Output;
    fn map_index_elements(self, map_fn: F) -> Self::Output;
}

pub trait MapIndexedElementsParallel<In, Out, GIdx, F: Fn(GIdx, Option<In>) -> Option<Out>> {
    type Output;
    fn map_index_elements_parallel(self, map_fn: F) -> Self::Output;
}
```

I tried to change all operators to use the map traits. This way the parallelization stuff is contained in one place.

Exception: The expression stull uses a lot of zips. Now twice as much since there is also a Vec with bools for each input.  This will not scale. I had to remove the 7 and 8 input variant since the compiler did complain. Solution: also use the new approach with access via grid_index.

Other things:
* the ResultDescriptor no longer has a no_data_value!
* i added access to Iterators to MaskedGrid to loop over all pixels as Options. 
* The GDAL source will derive the no_data_value from the dataset if it is not set.
* The GeoTiff writer will convert masked grids into GeoTiffs with no_data_value if the metadata provided holds an no_data_value.
* The WCS request "GetCoverage" has a "nodatavalue" option. If it is provided it will set the GeoTiff metadata and will return a GeoTiff with no_data values.
* I found a (kind of) broken test in the WMS handler.

TODOS:
* Find redundant code since there is no longer a difference between grids with and without no_data_value.
* Use the GDAL method to get / set masked bands
* in a lot of operators / algorithms the iteration over pixels was not parallel. I change that to "map_index_elements/map_or_mask_elements" however, we should see if the parallel version is beneficial.
* At some places we generate a new tile from inputs. I replaced that with creating a (Masked)GRid and then using a map call. We should add a method to generate a Grid / MaskedGrid with a closure similar tot he map calls (but without initial data).
* Change docs: there is no no_data_value anymore
* The expression still has an output no_data_value. We should talk about that. It could also return an Option.
* Fix the test in WMS. Use the GDAL output as reference.
* ...


operators/src/plot/box_plot.rs
```diff
@@ -38,10 +38,9 @@ pub struct BoxPlotParams {
     /// Name of the (numeric) attributes to compute the box plots on.
     #[serde(default)]
     pub column_names: Vec<String>,
-
-    /// For rasters, we have the option to include no-data values
-    #[serde(default)]
-    pub include_no_data: bool,
+    // For rasters, we have the option to include no-data values
+    // #[serde(default)]
+    // pub include_no_data: bool,
```
michaelmattig, line 43: This makes no sense now right? I f no data has no value, we can no longer plot it right? If so, you can remove this line too

operators/src/plot/box_plot.rs
```diff
@@ -293,18 +281,10 @@ impl BoxPlotRasterQueryProcessor {
...
-                        accum.update(iter)?;
-                    },
                     // Ignore empty grids if no_data should not be included
                     GridOrEmpty::Empty(_) => {},
-                    GridOrEmpty::Grid(grid) if include_no_data => {
-                        accum.update(grid.data.iter().map(|x| (*x).as_()))?;
-                    },
                     GridOrEmpty::Grid(grid) => {
-                        accum.update(grid.data.iter().filter(|&x| !grid.is_no_data(*x)).map(|x| (*x).as_()))?;
+                        accum.update(grid.masked_copy_element_iterator().filter_map(|pixel_option| pixel_option.map(|p| { let v: f64 = p.as_(); v})))?; // TODO: if empty pixels shuld be included we need to add an enmpty pixel count in the accu
```
michaelmattig, line 287: I guess this was the include_nodata flag? But where would that be drawn in a boxplot?

Also: why do the elements have to be copied here?

datatypes/src/raster/operations/grid_blit.rs
```diff
@@ -66,14 +66,120 @@ where
...
+        + GridBounds<IndexArray = I>
+        + GridSpaceToLinearSpace<IndexArray = I>
+        + Clone
+        + PartialEq,
+    D2: GridSize<ShapeArray = A>
+        + GridBounds<IndexArray = I>
+        + GridSpaceToLinearSpace<IndexArray = I>
+        + Clone
+        + PartialEq
+        + PartialEq,
```
michaelmattig, line 117: redundant

operators/src/plot/statistics.rs
```diff
@@ -161,20 +161,15 @@ impl PlotQueryProcessor for StatisticsQueryProcessor {
...
-            }
-        }
-    } else {
-        for &value in &tile_grid.data {
+fn process_raster<I>(number_statistics: &mut NumberStatistics, data: I)
+where
+    I: Iterator<Item = Option<f64>>,
+{
+    for value_option in data {
+        if let Some(value) = value_option {
```
michaelmattig, line 169: question: for rasters with no no data value we didn't have to check whether they are no data. Now we have to check it always. Would it make sense to have a grid type where we know there is data everywhere (like in the empty case where we know there is only nodata) and implement operations more efficiently. Here for example I guess even on an empty grid we would iterate over a bunch of None pixels. Can we avoid that?

edit: nevermind, emptygrids are handlet completely difference ind line 149 I see now. Still the question with tiles with no nodata remains.

operators/src/processing/expression/mod.rs
```diff
@@ -43,7 +43,7 @@ pub struct ExpressionParams {
     #[serde(with = "float_with_nan")]
     pub output_no_data_value: f64, // TODO: check value is valid for given output type during deserialization
     pub output_measurement: Option<Measurement>,
-    pub map_no_data: bool,
+    pub map_no_data: bool, // TODO: remove since that has no use anymore!
```
michaelmattig, line 46: is that so? I think the param specified, whether the expression is evaluated at all if there is a single no data pixel among all inputs and this hasn't changed.

operators/src/processing/expression/mod.rs
```diff
@@ -284,6 +283,7 @@ impl RasterOperator for Expression {
             sources,
             expression,
             map_no_data: self.params.map_no_data,
+            output_no_data_value: self.params.output_no_data_value,
```
michaelmattig, line 286: why is this needed now? Can't we just put 0 there? I guess we really have to think about what the meaning of a pixel value is if it is masked as no data.

operators/src/processing/expression/mod.rs
```diff
@@ -43,7 +43,7 @@ pub struct ExpressionParams {
     #[serde(with = "float_with_nan")]
     pub output_no_data_value: f64, // TODO: check value is valid for given output type during deserialization
```
michaelmattig, line 44: we can remove this now right? We can just put 0 there, or not? I guess we really have to think about what the meaning of a pixel value is if it is masked as no data. I'd say it has no meaning and can be arbitrary.

operators/src/processing/expression/mod.rs
```diff
@@ -476,6 +474,8 @@ impl InitializedRasterOperator for InitializedExpression {
                     .boxed()
                 )
             }
+            // TODO: the zip method in the processor allows only a limited number of arrays. Use idx iterator and get_at_ instead?
```
michaelmattig, line 477: can we extend it to more inputs? We have to fix this somehow before merging.

operators/src/processing/expression/query_processor.rs
```diff
@@ -172,38 +174,46 @@ where
...
                 }
 
-                let result = expression(a.as_(), is_no_data, out_no_data.as_());
-                TO::from_(result)
+                let result = expression(a.as_(), !is_a_valid, out_no_data.as_());
+
+                let result_t = TO::from_(result);
+
+                let out_is_no_data =
+                    result_t == out_no_data || (out_no_data != out_no_data && result != result);
```
michaelmattig, line 210: what is happening here?

operators/src/processing/interpolation/mod.rs
```diff
@@ -124,8 +122,8 @@ where
     source: Q,
     params: InterpolationParams,
     tiling_specification: TilingSpecification,
-    no_data_value: P,
     interpolation: PhantomData<I>,
+    _phantom_data: PhantomData<P>,
```
michaelmattig, line 126: remove? Can't we bind this to the interpolation field?

operators/src/processing/temporal_raster_aggregation/mean_aggregation_subquery.rs
```diff
@@ -74,50 +72,48 @@ impl<T> TemporalMeanTileAccu<T> {
...
-                    .iter_mut()
-                    .zip(self.count_grid.data.iter_mut())
-                    .zip(in_tile_grid.data.iter())
-                {
-                    if in_tile_grid.is_no_data(*new_value) {
-                        // The input pixel value is nodata
-                        if !self.ignore_no_data {
-                            // once nodata always nodata
-                            *acc_count = 0;
+            GridOrEmpty::Grid(_) => {
```
michaelmattig, line 93: why do you no longer use the accu grid here?

operators/src/processing/temporal_raster_aggregation/mean_aggregation_subquery.rs
```diff
@@ -74,50 +72,48 @@ impl<T> TemporalMeanTileAccu<T> {
...
+                let accu_grid = self.value_grid.clone().into_materialized_grid(); // TODO do not clone!
```
michaelmattig, line 94: why do you use the tile as accu?

operators/src/processing/temporal_raster_aggregation/mean_aggregation_subquery.rs
```diff
@@ -42,15 +43,12 @@ pub struct TemporalMeanTileAccu<T> {
     time: TimeInterval,
     tile_position: GridIdx2D,
     global_geo_transform: GeoTransform,
-
-    value_grid: GridOrEmpty2D<f64>,
-    count_grid: Grid2D<u64>,
+    value_grid: GridOrEmpty2D<(f64, u64)>,
+    // count_grid: Grid2D<u64>,
```
michaelmattig, line 47: what happened to this?

operators/src/source/gdal_source/mod.rs
```diff
@@ -299,7 +300,7 @@ where
 {
     pub tiling_specification: TilingSpecification,
     pub meta_data: GdalMetaData,
-    pub no_data_value: Option<T>,
+    pub _phandom_data: PhantomData<T>,
```
michaelmattig, line 303: typo

operators/src/source/gdal_source/mod.rs
```diff
@@ -673,7 +644,21 @@ where
...
     )?;
-    Grid::new(tile_grid, buffer.data, no_data_value).map_err(Into::into)
+    let data = Grid::new(tile_grid, buffer.data)?;
+
+    let masked_grid = MaskedGrid::from(data).map_or_mask_elements(|e| {
+        // TODO: test if parallel map is better.
+        let e = e.unwrap();
+
+        if let Some(no_data) = no_data_value {
+            if no_data == e || (no_data != no_data && e != e) {
```
michaelmattig, line 654: again I am confused. Do you do that to check for NAN? maybe we should add a util method for that

operators/src/source/gdal_source/mod.rs
```diff
@@ -673,7 +644,21 @@ where
         (tile_x_size, tile_y_size),       // requested raster size
         None,                             // sampling mode
     )?;
-    Grid::new(tile_grid, buffer.data, no_data_value).map_err(Into::into)
+    let data = Grid::new(tile_grid, buffer.data)?;
+
+    let masked_grid = MaskedGrid::from(data).map_or_mask_elements(|e| {
+        // TODO: test if parallel map is better.
```
michaelmattig, line 650: yeah I guess we should test that. Also we should look into storing our dataset as masked raster bands to avoid doing this conversion over and over and over and over again.

Also: didn't you say Gdal allows somehow accessing a band as a masked band?

services/src/datasets/external/netcdfcf/mod.rs
```diff
@@ -454,7 +452,7 @@ impl NetCdfCfDataProvider {
             rasterband_channel: 0, // we calculate offsets in our source
             geo_transform,
             file_not_found_handling: FileNotFoundHandling::Error,
-            no_data_value: result_descriptor.no_data_value,
+            no_data_value: data_array.no_data_value(), // we could also leave this empty. The gdal source will try to get the correct one.
```
michaelmattig, line 455: I guess we should really remove this. We no longer need it for the operators, so it shouldn't be defined in the dataset definition. The only thing it would be useful though would be for guessing the output not data value during export. But we said we wouldn't do that.

services/src/handlers/wms.rs
```diff
@@ -361,13 +358,14 @@ mod tests {
     use actix_web::http::Method;
     use actix_web_httpauth::headers::authorization::Bearer;
     use geoengine_datatypes::operations::image::RgbaColor;
-    use geoengine_datatypes::primitives::SpatialPartition2D;
+    // use geoengine_datatypes::primitives::SpatialPartition2D;
```
michaelmattig, line 361: remove the commented out code

services/src/handlers/wms.rs
```diff
@@ -447,6 +445,8 @@ mod tests {
         check_allowed_http_methods(get_capabilities_test_helper, &[Method::GET]).await;
     }
 
+    // The result should be similar to the GDAL output of this command: gdalwarp -tr 1 1 -r near -srcnodata 0 -dstnodata 0  MOD13A2_M_NDVI_2014-01-01.TIFF MOD13A2_M_NDVI_2014-01-01_360_180_near_0.TIFF
+    /*
     #[tokio::test]
     async fn png_from_stream_non_full() {
```
michaelmattig, line 451: why is this test commented out?

services/src/handlers/workflows.rs
```diff
@@ -413,7 +412,7 @@ async fn create_dataset<C: Context>(
                 height: (info.query.spatial_bounds.size_y() / info.query.spatial_resolution.y)
                     .ceil() as usize,
                 file_not_found_handling: FileNotFoundHandling::Error,
-                no_data_value: result_descriptor.no_data_value,
+                no_data_value: None,      // TODO: how to handle this here?
```
michaelmattig, line 415: remove it from the GdalDataSetParameters

services/src/ogc/wcs/request.rs
```diff
@@ -77,6 +77,11 @@ pub struct GetCoverage {
     #[serde(default)]
     #[serde(deserialize_with = "from_str_option")]
     resy: Option<f64>,
+
+    // Geo Engine specific
+    #[serde(default)]
+    #[serde(deserialize_with = "from_str_option")]
+    pub nodatavalue: Option<f64>,
```
michaelmattig, line 84: We need to document this in our docs

datatypes/src/operations/image/to_png.rs
```diff
@@ -78,24 +92,42 @@ where
     }
 }
 
-fn create_rgba_image<P: Pixel + RgbaTransmutable, N: Fn(P) -> bool>(
+fn create_rgba_image_from_grid<P: Pixel + RgbaTransmutable>(
```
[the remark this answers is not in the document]
jdroenner: There might be a case where we want to have grids without mask. So i keept this method.

datatypes/src/plots/histogram.rs
```diff
@@ -208,16 +208,14 @@ impl Histogram {
...
-    pub fn add_raster_data<P: Pixel>(&mut self, data: &[P], no_data_value: Option<P>) {
-        if let Some(no_data_value) = no_data_value {
-            for &value in data {
-                self.handle_data_item(value.as_(), value == no_data_value);
+    pub fn add_raster_data<P: Pixel, I: Iterator<Item = Option<P>>>(&mut self, data: I) {
+        data.for_each(|pixel_option| {
+            if let Some(p) = pixel_option {
+                self.handle_data_item(p.as_(), false);
+            } else {
+                self.handle_data_item(0., true); // TODO: remove not needed zero
```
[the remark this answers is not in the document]
jdroenner: or add a "inc_no_data_count()". There is something like this somewhere else

datatypes/src/raster/operations/grid_blit.rs
```diff
@@ -66,14 +66,120 @@ where
     }
 }
 
-impl<D, T> GridBlit<EmptyGrid<D, T>, T> for Grid2D<T>
+impl<D, T> GridBlit<Grid<D, T>, T> for Grid3D<T>
```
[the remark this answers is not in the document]
jdroenner: it helps to validate that the approach also works for more dimensions.

datatypes/src/raster/operations/grid_blit.rs
```diff
@@ -66,14 +66,120 @@ where
...
+    fn grid_blit_from(&mut self, other: &MaskedGrid<D1, T>) {
+        // easy part: blit the data
+        self.inner_grid.grid_blit_from(other.as_ref());
+        self.validity_mask.grid_blit_from(other.mask_ref());
+    }
+}
+
+impl<D1, D2, T, A, I> GridBlit<EmptyGrid<D1, T>, T> for MaskedGrid<D2, T>
+where
+    D1: GridSize<ShapeArray = A>
```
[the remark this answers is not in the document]
jdroenner: We could think about that. I don't think we need the combination that often.

datatypes/src/raster/operations/grid_blit.rs
```diff
@@ -66,14 +66,120 @@ where
...
+    T: Copy + Sized + Default,
+    Grid<D2, bool>: GridBlit<EmptyGrid<D1, T>, T>,
+    Grid<D2, bool>: GridBlit<Grid<D1, bool>, bool>,
+{
+    fn grid_blit_from(&mut self, other: &EmptyGrid<D1, T>) {
+        self.mask_mut().grid_blit_from(other);
+    }
+}
+
+impl<D1, D2, T, A, I> GridBlit<Grid<D1, T>, T> for MaskedGrid<D2, T>
```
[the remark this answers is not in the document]
jdroenner: the implementation from maskedgrid uses this.

datatypes/src/raster/operations/interpolation.rs
```diff
@@ -37,40 +39,21 @@ where
         let out_x_size = info_out.global_geo_transform.x_pixel_size();
         let out_y_size = info_out.global_geo_transform.y_pixel_size();
 
-        let parallelism = rayon::current_num_threads();
-        let rows_per_task =
-            num::integer::div_ceil(output.grid_array.shape.axis_size_y(), parallelism);
-
-        let chunk_size = output.grid_array.shape.axis_size_x() * rows_per_task;
+        let map_fn = |gidx: GridIdx2D, _current_pixel_value: Option<P>| {
```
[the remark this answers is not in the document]
jdroenner: yes. i suggested this in the summary.

datatypes/src/raster/operations/interpolation.rs
```diff
@@ -37,40 +39,21 @@ where
...
+            let GridIdx([y, x]) = gidx;
+            let out_y_coord = out_upper_left.y + y as f64 * out_y_size;
```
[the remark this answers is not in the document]
jdroenner: this is one (MAC/MAD) instruction for the CPU. I don't want to write and validate the same logic all the time to save that.

datatypes/src/raster/operations/interpolation.rs
```diff
@@ -37,40 +39,21 @@ where
...
-                                .get_at_grid_index_unchecked([nearest_in_y_idx, nearest_in_x_idx]);
-
-                            *pixel = value;
-                        });
-                    });
-            });
+            .clone()
+            .map_index_elements_parallel(map_fn);
+
+        output.grid_array = out_grid; // TODO: add a trait for update elements?
```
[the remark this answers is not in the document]
jdroenner: we generate a new grid (new memory) and discard the old one. We could also have an implementation for "&mut Grid"

datatypes/src/raster/operations/interpolation.rs
```diff
@@ -126,58 +109,53 @@ where
...
+            let a_v = input.get_masked_at_grid_index_unchecked([in_y_idx, in_x_idx]);
+
+            let b_v = input.get_masked_at_grid_index_unchecked([in_y_idx + 1, in_x_idx]);
+
+            let c_v = input.get_masked_at_grid_index_unchecked([in_y_idx, in_x_idx + 1]);
+
+            let d_v = input.get_masked_at_grid_index_unchecked([in_y_idx + 1, in_x_idx + 1]);
+
+            let value = match (a_v, b_v, c_v, d_v) {
+                (Some(a), Some(b), Some(c), Some(d)) => Some(Self::bilinear_interpolation(
```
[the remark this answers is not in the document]
jdroenner: i guess so

datatypes/src/raster/operations/map_indexed_elements.rs
```diff
@@ -0,0 +1,198 @@
...
+                        let grid_idx = data.shape.grid_idx_unchecked(lin_idx);
+
+                        let in_masked_value = if *mask { Some(*i) } else { None };
+
+                        let out_value_option = map_fn(grid_idx, in_masked_value);
+
+                        *mask = out_value_option.is_some();
+
+                        if let Some(out_value) = out_value_option {
+                            *out = out_value;
```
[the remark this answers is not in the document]
jdroenner: the mask is set with the "*mask = ..."

datatypes/src/raster/operations/map_masked_elements.rs
```diff
@@ -0,0 +1,234 @@
...
+            .into_par_iter()
+            .with_min_len(data.shape.axis_size_x())
+            .zip(
+                validity_mask
+                    .data
+                    .into_par_iter()
+                    .with_min_len(validity_mask.shape.axis_size_x()),
+            )
+            .map(|(i, m)| {
+                let in_value = if m { Some(i) } else { None };
```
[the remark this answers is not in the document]
jdroenner: No it is not clear. This is the way arrow does it. And we should test that. We could also use a bitmask instead of a groid of booleans. However, i think optimization of this is a new task

datatypes/src/raster/operations/map_masked_elements.rs
```diff
@@ -0,0 +1,234 @@
...
+                if let Some(o) = map_fn(in_value) {
+                    (o, m)
+                } else {
+                    (Out::default(), false)
+                }
+            })
+            .collect();
+
+        MaskedGrid::new(
+            Grid::new(shape.clone(), new_data).expect("Grid creation failed before"),
```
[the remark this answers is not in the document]
jdroenner: before this point.

datatypes/src/raster/grid_or_empty.rs
```diff
@@ -18,14 +19,15 @@ pub type GridOrEmpty3D<T> = GridOrEmpty<GridShape3D, T>;
 #[serde(rename_all = "camelCase")]
 #[serde(tag = "type")]
 pub enum GridOrEmpty<D, T> {
-    Grid(Grid<D, T>),
+    Grid(MaskedGrid<D, T>),
     Empty(EmptyGrid<D, T>),
+    // MaskedGrid(MaskedGrid<D,T>),
```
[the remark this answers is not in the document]
jdroenner: i left it as reminder that we could also have a Grid(Grid))

datatypes/src/raster/grid_traits.rs
```diff
@@ -136,6 +136,31 @@ pub trait GridIndexAccess<T, I> {
...
 }
 
+pub trait MaskedGridIndexAccessMut<T, I> {
+    /// Sets a masked value at a grid index. None represents a missing value / no-data.
+    ///
+    /// # Errors
+    ///
+    /// The method fails if the grid index is out of bounds.
+    ///
+    fn set_masked_at_grid_index(&mut self, grid_index: I, value: Option<T>) -> Result<()>;
```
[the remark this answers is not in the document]
jdroenner: this trait already exists.

datatypes/src/raster/grid_traits.rs
```diff
@@ -136,6 +136,31 @@ pub trait GridIndexAccess<T, I> {
...
+
+    /// Sets the value at a grid index
+    fn set_masked_at_grid_index_unchecked(&mut self, grid_index: I, value: Option<T>);
+}
+
+/// Read elements stored in a Grid
+pub trait MaskedGridIndexAccess<T, I> {
```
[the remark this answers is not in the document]
jdroenner: I think it is not a good idea to start working around the no data stuff this early. We should check if we store Option < T > if the access to both arrays is to expensive.

This trait exists for grid. you can use if you get the grid from the masked grid.

datatypes/src/raster/masked_grid.rs
```diff
@@ -0,0 +1,274 @@
...
+    D: GridBounds<IndexArray = I> + Clone,
+    T: Clone,
+    GridBoundingBox<I>: GridSize,
+    GridIdx<I>: Add<Output = GridIdx<I>> + From<I> + Clone,
+{
+    type Output = MaskedGrid<GridBoundingBox<I>, T>;
+
+    fn shift_by_offset(self, offset: GridIdx<I>) -> Self::Output {
+        MaskedGrid {
+            inner_grid: self.inner_grid.shift_by_offset(offset.clone()),
```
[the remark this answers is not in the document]
jdroenner: yes. we can rewrite the MaskedGrid. Currently it contaisn of a grid with values and a grid with booleans. We can change that but the task would be to find the optimal storage. Option < T >, Grid + bitmask ...

datatypes/src/util/test.rs
```diff
@@ -82,350 +79,207 @@ pub fn save_test_bytes(bytes: &[u8], filename: &str) {
...
-        assert!(grid_eq_with_no_data(&r1, &r2));
+        assert!(grid_eq(&r1, &r2));
     }
 
     #[test]
-    fn test_grid_eq_with_no_data_integral_fail_dim() {
+    fn test_grid_eq_with_integral_fail_dim() {
         let d1: GridShape2D = [4, 1].into();
         let d2: GridShape2D = [2, 2].into();
+        let r1 = Grid2D::new(d1, vec![1, 2, 3, 42]).unwrap();
```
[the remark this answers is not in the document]
jdroenner: no this is the value in the data grid. in the mask grid there is a false. Or we could/and should use the iterator that generates options.

operators/src/adapters/raster_subquery/raster_subquery_reprojection.rs
```diff
@@ -305,45 +300,22 @@ where
...
-        let parallelism = pool.current_num_threads();
-        let par_chunk_split =
-            num::integer::div_ceil(axis_size_y, parallelism).max(min_rows_in_par_chunk); // don't go below MIN_ROWS_IN_PAR_CHUNK lines per chunk.
-        let par_chunk_size = axis_size_x * par_chunk_split;
+        let map_fn = |grid_idx, accu_value: Option<T>| {
+            let lookup_coord = coords.get_at_grid_index_unchecked(grid_idx);
+            let lookup_value = lookup_coord
+                .filter(|coord| tile_bounding_box.contains_coordinate(coord))
+                .and_then(|coord| tile.pixel_value_at_coord_unchecked(coord));
+            lookup_value.or(accu_value)
```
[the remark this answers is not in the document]
jdroenner: in the old solution the pixel was only changed if there was a new value. With the method, we need to return a value to store it. If we return None then none is set.

operators/src/plot/box_plot.rs
jdroenner: yes

operators/src/plot/statistics.rs
jdroenner: good question. We can add the Grid(Grid) variant to the GridOrEmpty enum. The first time a Grid(Grid) is used in an operator where empty pixels can be a result it has to be converted into a MAskedGrid in order to be able to use parallel processing. Then we can check if all validity flags are true and turn it back. i think it is not worth it.

operators/src/processing/expression/mod.rs
jdroenner: nope we still need it. we have to change the return type of the expression to be either Option< T > or (T, bool). If we set it to 0 the all results of the expression that are 0 eill be flaged as invalid/no_data.

operators/src/processing/expression/mod.rs
jdroenner: is that so? then we can keep it.

operators/src/processing/expression/mod.rs
jdroenner: i answered that above

operators/src/processing/expression/mod.rs
jdroenner: yes and no. we need to remove the number of zips. so we can either combine zips into tuples early or we can use the approach i used for the map stuff.

operators/src/processing/expression/query_processor.rs
jdroenner: we need to check if the result value is the no data value the user set. so the result is either equal to the no_data value or both result and no_data are NAN.

operators/src/processing/interpolation/mod.rs
jdroenner: good catch. will check that,

operators/src/processing/temporal_raster_aggregation/mean_aggregation_subquery.rs
jdroenner: i don't the value_grid is the grid in the accu

operators/src/source/gdal_source/mod.rs
jdroenner: 1. yes. there are a lot of places where we could use the parallel map. we should do that in an new task.
2. gdal can do that. we have to implement it first.

operators/src/source/gdal_source/mod.rs
jdroenner: yes this checks for NAN

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -206,6 +207,21 @@ impl<P: Pixel + GdalType> GdalDatasetWriter<P> {
     }
 
     fn write_tile(&self, tile: RasterTile2D<P>) -> Result<()> {
+        // FIXME: transform masks to no-data values if the gdal_tiff_metadata no-data value is set. This might need to move somewhere else. Also: write the mask if no no-data value is set.
+        let tile = if let Some(replace_no_data_value) = self.gdal_tiff_metadata.no_data_value {
```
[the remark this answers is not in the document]
jdroenner: i would like to avoid adding no data values anywhere. I guess this is the best way to do it.

services/src/handlers/wms.rs
jdroenner: we will need it if the test is fixed i guess

services/src/handlers/wms.rs
jdroenner: i told you about that in slack. the test is broken. With the old implementation it is fine. However, the result does not match the gdal result. The changes in this PR return the gdal result which is not identical to the test data.

services/src/ogc/wcs/request.rs
jdroenner: yes. we need to change a lot of docs for this PR

datatypes/src/raster/operations/grid_blit.rs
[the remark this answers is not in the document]
jdroenner: i tried to remove non needed traits.

datatypes/src/raster/operations/interpolation.rs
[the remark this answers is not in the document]
jdroenner: i think it should be fine like this atm. The compiler should be smart enough to figure out that we do not read the vale. However, we need to set it so the memory is alocated and accessed anyway.

datatypes/src/raster/operations/interpolation.rs
```diff
@@ -126,58 +109,53 @@ where
         let out_x_size = info_out.global_geo_transform.x_pixel_size();
         let out_y_size = info_out.global_geo_transform.y_pixel_size();
 
-        let parallelism = rayon::current_num_threads();
-        let rows_per_task =
-            num::integer::div_ceil(output.grid_array.shape.axis_size_y(), parallelism);
+        let map_fn = |g_idx: GridIdx2D, _current_pixel_value: Option<P>| {
```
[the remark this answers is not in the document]
jdroenner: since wa cant have a "&mut pixel" because the pixel is currently a value and a bool we need to pass the current value in. It is a copy type so there should be no harm. The funtion then updates the value with the result value.

datatypes/src/raster/operations/interpolation.rs
```diff
@@ -126,58 +109,53 @@ where
...
+            let GridIdx([y_idx, x_idx]) = g_idx;
+
+            let out_y = out_upper_left.y + y_idx as f64 * out_y_size;
+            let in_y_idx = ((out_y - in_upper_left.y) / in_y_size).floor() as isize;
+
+            let a_y = in_upper_left.y + in_y_size * in_y_idx as f64;
```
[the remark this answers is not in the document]
jdroenner: like i said. this is a single instruction. i dont think it is worth the overhead. someone can write a benchmark for that...

datatypes/src/raster/operations/map_elements.rs
```diff
@@ -0,0 +1,267 @@
...
+    G: GridSize + Clone + Send + Sync,
+    F: Fn(In) -> Out + Sync + Send,
+{
+    type Output = Grid<G, Out>;
+    fn map_elements_parallel(self, map_fn: F) -> Self::Output {
+        let shape = self.shape.clone();
+        let data = self
+            .data
+            .into_par_iter()
+            .with_min_len(self.shape.axis_size_x())
```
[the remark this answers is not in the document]
jdroenner: tHe key here is the ".with_min_len(self.shape.axis_size_x())". The RAyon docs say:

"Sets the minimum length of iterators desired to process in each thread. Rayon will not split any smaller than this length, but of course an iterator could already be smaller to begin with.

Producers like zip and interleave will use greater of the two minimums. Chained iterators and iterators inside flat_map may each use their own minimum length."

So i interprete that as the same logic we added with the chunks.

datatypes/src/raster/grid_or_empty.rs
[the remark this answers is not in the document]
jdroenner: removed it

datatypes/src/raster/masked_grid.rs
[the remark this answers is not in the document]
michaelmattig: maybe create a task or a todo in the code

datatypes/src/raster/operations/interpolation.rs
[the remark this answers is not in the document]
michaelmattig: but the logic was already there and I think the code was still quite readable (especially in comparison to the abominations we have elsewhere in code). Also, it's a lookup, an addition and a multiplication, so not quite one instruction. And it will be performed redundantly thousands of times. Also, I think it should be possible to encapsulate this to reduce redundant code, where you provide some code that is done for each row and then some code that is done for each cell. 

For now, I don't find this solution better than what was there previously because we have a redundant parameter in the map function and redundant computations.

datatypes/src/raster/operations/interpolation.rs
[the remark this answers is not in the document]
michaelmattig: couldn't we just have to types of map iterations? One that takes index and pixel value and one that gives only the index so we don't carry over unneeded data?

datatypes/src/raster/operations/map_masked_elements.rs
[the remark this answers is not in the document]
michaelmattig: is that a good message though? Shouldn't it say why you expect it to succeed?

datatypes/src/plots/histogram.rs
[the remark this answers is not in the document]
michaelmattig: maybe you can add this method. Otherwise it will probably stay like this forever.

operators/src/processing/expression/mod.rs
michaelmattig: yes we should change the return type of the expression then

operators/src/util/raster_stream_to_geotiff.rs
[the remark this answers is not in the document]
michaelmattig: if we just use it for output I think it would be much cleaner. At least add a todo here please because I think this is rather hacky.

operators/src/processing/expression/query_processor.rs
```diff
@@ -389,46 +399,52 @@ macro_rules! impl_expression_tuple_processor {
...
 
-                let min_batch_size = rasters[0].grid_array.grid_shape().axis_size_x();
+                let shape = rasters[0].grid_shape().clone();
+
+                let mut out = GridOrEmpty::from(EmptyGrid2D::<TO>::new(shape));
 
-                let [ $($TILE),* ] = rasters.map(|raster| raster.into_materialized_tile());
+                let map_fn = move |lin_idx: usize, _empty: Option<TO>| {
+                    $(
+                        let $PIXEL = rasters[$I].get_at_grid_index_unchecked(lin_idx);
```
michaelmattig, line 414: why can you just use a lin_idx as a gridIndex here? Aren't grid index n-dimensional?

operators/src/processing/temporal_raster_aggregation/mean_aggregation_subquery.rs
```diff
@@ -74,50 +72,40 @@ impl<T> TemporalMeanTileAccu<T> {
...
+                            let delta = ivf - acc_value;
+                            let new_acc_value = acc_value + delta / (new_acc_count as f64);
+                            Some((new_acc_value, new_acc_count))
                         }
+                        (None, Some(_) | None) | (Some(_), None) => None,
                     }
-                }
+                };
+
+                g.update_indexed_elements(map_fn); // TODO: make this patallel?
```
michaelmattig, line 108: typo

operators/src/plot/statistics.rs
michaelmattig: so, should we add this as a card or consider it a moonshot and get to it whenever?

operators/src/source/gdal_source/mod.rs
michaelmattig: can you add some docs to the NoDataValueGrid? At first glance once could think that it's a grid full of no data.

datatypes/src/raster/raster_traits.rs
```diff
@@ -0,0 +1,140 @@
...
+impl<G, P> CoordinatePixelAccess<P> for BaseTile<G>
+where
+    G: GridSize + Clone,
+    P: Pixel,
+    Self: GridIndexAccess<Option<P>, GridIdx2D>,
+{
+    type Output = Option<P>;
+
+    fn pixel_value_at_coord(&self, coordinate: Coordinate2D) -> Result<Option<P>> {
+        // TODO: benchmark the impact of creating the `GeoTransform`s
```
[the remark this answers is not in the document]
michaelmattig: what task do you mean?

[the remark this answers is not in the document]
jdroenner: uh i guess i got lost while writing the comment. We discussed to change the way how we work with the location of raster data. For that we suggested to change access to Grids via GridIdx only. If we do that the geo_transform will move away from Grid / RasterTile into the TilingSpec or re something thats static for a workflow execution. If we do this we have to "remove" this access method anyway.

operators/src/source/gdal_source/mod.rs
jdroenner: [47ea273](https://github.com/geo-engine/geoengine/pull/561/commits/47ea27352a728b4d7f22efc52bbc45d594f3f0cc)

CHANGELOG.md
```diff
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added traits for updating / mapping the pixels of grid types. Also added a trait for creating grid types using a function for each pixels location.
+
+  - https://github.com/geo-engine/geoengine/pull/561
+  - There is now a trait "`UpdateElements`" for updating the pixels of a grid type and a parallel version "`UpdateElementsParallel`".
```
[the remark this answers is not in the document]
jdroenner: [7e9a1ef](https://github.com/geo-engine/geoengine/pull/561/commits/7e9a1efd1b5b266b4df9d20dc5a9743f88c1ec33)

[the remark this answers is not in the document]
jdroenner: [7e9a1ef](https://github.com/geo-engine/geoengine/pull/561/commits/7e9a1efd1b5b266b4df9d20dc5a9743f88c1ec33)

datatypes/src/raster/no_data_value_grid.rs
```diff
@@ -5,6 +5,8 @@ pub type NoDataValueGrid1D<T> = NoDataValueGrid<GridShape1D, T>;
 pub type NoDataValueGrid2D<T> = NoDataValueGrid<GridShape2D, T>;
 pub type NoDataValueGrid3D<T> = NoDataValueGrid<GridShape3D, T>;
 
+/// A `NoDataValueGrid` is an n-dmensional array.
+/// In this grid type each element/pixel can be invalid. This is the case if the element/pixel value is equal to the value storef in the `no_data_value` field.
```
michaelmattig, line 12: typo "storef"

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added traits for updating / mapping the pixels of grid types. Also added a trait for creating grid types using a function for each pixels location.
```
michaelmattig, line 12: "Added methods"? You implemented the functionality too

jdroenner: [80d8a47](https://github.com/geo-engine/geoengine/pull/561/commits/80d8a4701b42eec34ec0d075b9523292a48c9b6f)

datatypes/src/raster/no_data_value_grid.rs
jdroenner: [b154e04](https://github.com/geo-engine/geoengine/pull/561/commits/b154e04d0b7c40afde418b281cdd52078d2909f0)

michaelmattig (approved):

michaelmattig (approved):

michaelmattig:
bors merge

PR #577, opened 2022-07-15
Read masked data

jdroenner:


jdroenner:
[Gdal reads masked data](https://trello.com/c/4WcJSHIJ/901-gdal-reads-masked-data)

operators/src/source/gdal_source/mod.rs
```diff
@@ -610,31 +615,55 @@ where
...
+        let no_data_value = dataset_params
+            .no_data_value
+            .or_else(|| rasterband.no_data_value())
+            .and_then(FromPrimitive::from_f64);
+        let no_data_value_grid = NoDataValueGrid::new(data_grid, no_data_value);
+        let grid_or_empty = GridOrEmpty::from(no_data_value_grid);
+        return Ok(grid_or_empty);
+    }
+
+    debug!("use mask based no-data handling.");
```
michaelmattig, line 638: maybe throw error when the mask flag indicates alpha channel?

operators/src/source/gdal_source/mod.rs
```diff
@@ -592,15 +593,19 @@ fn read_grid_from_raster<
...
+    dataset_params: &GdalDatasetParameters,
 ) -> Result<GridOrEmpty<D, T>>
 where
-    T: Pixel + GdalType + Default,
+    T: Pixel + GdalType + Default + FromPrimitive,
     D: PartialEq + Clone,
 {
+    let raster_band_index = dataset_params.rasterband_channel;
+
+    let rasterband = dataset.rasterband(raster_band_index as isize)?;
```
michaelmattig, line 607: the rasterband is now opened twice(?) (if there is propertiesMapping). Do you know whether there is some cost in doing so?

jdroenner: i dont know. We can change it back once we add the mask band getter to the gdal crate

operators/src/source/gdal_source/mod.rs
jdroenner: i think many tiffs use the alpha channel even if it is only true, false

operators/src/util/gdal.rs
```diff
@@ -29,6 +29,105 @@ use crate::{
...
+    if raster_band_ptr.is_null() {
+        return Err(Error::Gdal {
+            source: gdal::errors::GdalError::NullPointer {
+                method_name: "GDALGetRasterBand",
+                msg: "Could not open RasterBand".to_string(),
+            },
+        });
+    }
+    let band_mask_flags = unsafe { gdal_sys::GDALGetMaskFlags(raster_band_ptr) };
+    // TODO: if 0 that should be an error...
```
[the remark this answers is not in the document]
jdroenner: gdal says: Applications should never destroy GDALRasterBands directly, instead destroy the GDALDataset. (see the destructor here: https://gdal.org/doxygen/classGDALRasterBand.html)

operators/src/util/gdal.rs
```diff
@@ -29,6 +29,105 @@ use crate::{
...
+        let raster_band_ptr = gdal_sys::GDALGetRasterBand(dataset.c_dataset(), raster_band_index);
+        if raster_band_ptr.is_null() {
+            return Err(Error::Gdal {
+                source: gdal::errors::GdalError::NullPointer {
+                    method_name: "GDALGetRasterBand",
+                    msg: "Could not open RasterBand".to_string(),
+                },
+            });
+        }
+        let mask_band_ptr = gdal_sys::GDALGetMaskBand(raster_band_ptr);
```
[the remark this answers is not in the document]
jdroenner: gdal says: Applications should never destroy GDALRasterBands directly, instead destroy the GDALDataset. (see the destructor here: https://gdal.org/doxygen/classGDALRasterBand.html)

operators/src/util/gdal.rs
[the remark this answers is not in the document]
jdroenner: https://github.com/geo-engine/geoengine/pull/577/commits/1b595b199c1a55039aa3d65a7e0ac930b6e18a0d

operators/src/source/gdal_source/mod.rs
michaelmattig: do you mean that there are a lot of tiffs where the mask flag is wrong? Maybe then we should add an override like for the nodata value and still throw an error if there is no override and the mask flag is alpha channel

operators/src/source/gdal_source/mod.rs
michaelmattig: ok. can you add a todo in the code?

operators/src/source/gdal_source/mod.rs
michaelmattig: By override I meant, that, like for no data values, the user should be able to define a _concrete_ GdalMaskFlag for the dataset that overrides the maskflag that is stored in the tiff. Then, in case the tiff says `GMF_ALPHA` we can override it with `GMF_NODATA` and use the alpha values as a nodata mask. Sorry if my comment was unclear. I think this would be more flexible than your current solution.

michaelmattig:
bors merge

PR #581, opened 2022-07-29
hierarchical listing of netcdfcf ebv layers

michaelmattig:


michaelmattig:
[[GEO BON] 1.x LayerAPI umsetzen](https://trello.com/c/W4pWeL0N/919-geo-bon-1x-layerapi-umsetzen)

services/src/datasets/external/netcdfcf/mod.rs
```diff
@@ -42,15 +47,19 @@ use geoengine_operators::{
...
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use walkdir::{DirEntry, WalkDir};
 
-mod error;
+#[cfg(feature = "ebv")]
+mod ebvportal_api;
+#[cfg(feature = "ebv")]
+mod ebvportal_provider;
```
ChristianBeilschmidt, line 61: Do we need all the cfgs? I thought this whole module only gets inserted if `ebv` is activated.

services/src/datasets/external/netcdfcf/mod.rs
```diff
@@ -1,14 +1,19 @@
+#[cfg(feature = "ebv")]
```
ChristianBeilschmidt, line 1: Same here.

services/src/datasets/external/netcdfcf/mod.rs
```diff
@@ -840,103 +855,270 @@ impl DataProvider for NetCdfCfDataProvider {
...
+    groups: &[String],
+) -> crate::error::Result<Option<NetCdfGroup>> {
+    if groups.is_empty() {
+        return Ok(None);
+    }
+
+    let mut group_stack = groups.iter().collect::<Vec<_>>();
+    let mut group = netcdf_groups
+        .into_iter()
+        .find(|g| g.name == *group_stack.remove(0))
```
ChristianBeilschmidt, line 931: If you always do remove(0), shouldn't you just once reverse the vector and then do pop?

services/src/datasets/external/netcdfcf/mod.rs
```diff
@@ -840,103 +855,270 @@ impl DataProvider for NetCdfCfDataProvider {
...
-        );
+        let id: NetCdfLayerCollectionId = serde_json::from_str(&collection.0)?;
 
-        let mut dir = tokio::fs::read_dir(&self.path).await?;
-
-        let mut datasets = vec![];
-        while let Some(entry) = dir.next_entry().await? {
-            if !entry.path().is_file() {
-                continue;
+        match id {
```
ChristianBeilschmidt, line 1021: Do you check somewhere that you cannot escape the NetCDF path?
E.g., that you cannot provide `../../hidden.nc` or `/media/secret/hidden.nc`?

I did this at some points by checking the the absolute file path I will load has the absolute provider path as its prefix or something like this.

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
```diff
@@ -0,0 +1,555 @@
...
+                entity: _,
+            } => return Err(Error::InvalidLayerCollectionId),
+        })
+    }
+
+    async fn root_collection_id(&self) -> Result<LayerCollectionId> {
+        EbvCollectionId::Classes.try_into()
+    }
+
+    async fn get_layer(&self, id: &LayerId) -> Result<Layer> {
```
ChristianBeilschmidt, line 417: Why can't you call get_layer from the normal NetCDF provider?

ChristianBeilschmidt (changes requested):

services/src/datasets/external/netcdfcf/mod.rs
```diff
@@ -840,103 +855,270 @@ impl DataProvider for NetCdfCfDataProvider {
...
+    Group {
+        path: PathBuf,
+        groups: Vec<String>,
+    },
+    Entity {
+        path: PathBuf,
+        groups: Vec<String>,
+        entity: usize,
+    },
+}
```
ChristianBeilschmidt, line 872: Two comments on this:

1. This enum misses `#[serde(tag = "type", rename_all = "camelCase")]`

2. Woudln't it make sense to reflect the hierarchies in the path?
 So why create a json file for that hierarchy, if we could say `/layers/community/species/ecostress/metric1/scenario1/entity5`?

ChristianBeilschmidt (changes requested):

services/src/layers/layer.rs
```diff
@@ -71,10 +75,13 @@ pub struct LayerCollection {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
 pub struct LayerCollectionListing {
     pub id: ProviderLayerCollectionId,
     pub name: String,
     pub description: String,
+    pub entry_label: Option<String>, // a common label for the collection's entries, if there is any
+    pub properties: Option<Vec<(String, String)>>,
```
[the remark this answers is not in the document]
michaelmattig: it makes initializing a bit more complicated but would make using it easier I guess

services/src/layers/layer.rs
```diff
@@ -29,13 +31,15 @@ pub struct Layer {
...
     pub symbology: Option<Symbology>,
+    pub properties: Option<HashMap<String, String>>,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub struct LayerListing {
     pub id: ProviderLayerId,
     pub name: String,
     pub description: String,
+    pub properties: Option<HashMap<String, String>>,
```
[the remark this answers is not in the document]
michaelmattig: it's because in the collection item listing the properties are stuff that is rendered as strings as part of the listing. In the LayerListing it contains stuff that it isn't displayed but used for actually loading the data, like timeSteps. There, I need to access the stuff based on the key instead of just rendering it (in order). Maybe we should call it differently here. metadata?

services/src/datasets/external/netcdfcf/ebvportal_api.rs
```diff
@@ -0,0 +1,218 @@
...
+    Ok(EbvHierarchy {
+        provider_id: NETCDF_CF_PROVIDER_ID,
+        tree: listing,
+    })
+}
+
+pub struct NetCdfCfDataProviderPaths {
+    pub provider_path: PathBuf,
+    pub overview_path: PathBuf,
+}
```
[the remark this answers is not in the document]
michaelmattig: I don't think it makes sense to test something here. The structs are modeled after the ebv portal API json response and putting it in here would test exactly nothing. What would make sense is to test that the EbvPortalProvider makes a correct listing, given correct responses from the ebv portal API. I can add that if you want.

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
```diff
@@ -0,0 +1,555 @@
...
+    async fn initialize(self: Box<Self>) -> crate::error::Result<Box<dyn DataProvider>> {
+        Ok(Box::new(EbvPortalDataProvider {
+            netcdf_cf_provider: NetCdfCfDataProvider {
+                path: self.path,
+                overviews: self.overviews,
+            },
+        }))
+    }
+
+    fn type_name(&self) -> String {
```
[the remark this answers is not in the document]
michaelmattig: in the providers we currently have, this is fix

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
```diff
@@ -0,0 +1,555 @@
...
+                        collection_id: EbvCollectionId::Dataset {
+                            class: class.to_string(),
+                            ebv: ebv.to_string(),
+                            dataset: d.id,
+                        }
+                        .try_into()?,
+                    },
+                    name: d.name.clone(),
+                    description: d.description,
+                    entry_label: Some("Metric".to_string()),
```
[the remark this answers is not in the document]
michaelmattig: I thought the EBV hierarchy was always class->ebv->dataset->metric->entity for the portal

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
```diff
@@ -0,0 +1,555 @@
...
+                                class: class.to_string(),
+                                ebv: ebv.to_string(),
+                                dataset: dataset.to_string(),
+                                groups: out_groups,
+                            }
+                            .try_into()?,
+                        },
+                        name: group.title,
+                        description: group.description,
+                        entry_label: None,
```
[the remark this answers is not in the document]
michaelmattig: do you have an example where the label is missing?

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
```diff
@@ -0,0 +1,555 @@
...
+                    .data_range
+                    .unwrap_or_else(|| (colorizer.min_value(), colorizer.max_value()));
+
+                Ok(Layer {
+                    id: ProviderLayerId {
+                        provider_id: EBV_PROVIDER_ID,
+                        layer_id: id.clone(),
+                    },
+                    name: format!("{} > {} > {} > {}", class, ebv, ebv_hierarchy.tree.title, /*netcdf_metric.title, */netcdf_entity.name),
+                    description: format!("{} > {} > {} > {}", class, ebv, ebv_hierarchy.tree.title, /* netcdf_metric.title,*/ netcdf_entity.name),
```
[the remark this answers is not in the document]
michaelmattig: this is only shown in the layer list which is not present in the data portals

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: same what?

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: 1. there is nothing to be camlized
2. yeah that would be great but I don't know if this is possible with actix as we would have to caputre arbitrary url paths. Also the path already contains slashes we can't put that into the url directly.

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: because it is less code this way

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: the netcdfdf mod has no conditional compiliation flag

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
```diff
@@ -0,0 +1,555 @@
...
+    async fn meta_data(
+        &self,
+        _id: &DataId,
+    ) -> Result<
+        Box<dyn MetaData<OgrSourceDataset, VectorResultDescriptor, VectorQueryRectangle>>,
+        geoengine_operators::error::Error,
+    > {
+        Err(geoengine_operators::error::Error::NotYetImplemented)
+    }
+}
```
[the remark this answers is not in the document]
michaelmattig: a92ebe88f86821a470b90502d4cb8e370cce7373

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: sounds like a good idea. Can you point me to the code that did this, because I couldn't find it.

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
michaelmattig: because for the ebv portal the path to the netcdf file is retrieved from the dataset via id and in the raw netcdfprovider the path is passed as part of the LayerCollectionId

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: if you mean this: https://github.com/geo-engine/geoengine/blob/c2406c742467d6f002ab0ab7bc2e2715868ccf9f/services/src/datasets/external/netcdfcf/mod.rs#L399 then either I don't understand it or it doesn't work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fcbe04de9cbaacd7bcc36f4fe7051bec

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: We have two test datasets, one has metric->ebv_cube and the other one scenario->metric->ebv_cube. We need to support at least our test data. They once encoded the ordering it into a variable in the file, but its gone. So we should assume the ordering of our test data.

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: There is never scenario, e.g. for `dataset_sm.nc`

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: If you rely on that the entity is always called ebv_cube, it is not a generic NetCDF provider, right?

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: cf. above

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: Then just use reverse once and pop instead of remove(0)?

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: Good point, we could use `join`, then `canonicalize`, and then check for `starts_with`?

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
ChristianBeilschmidt: Is it now only working when we have an overview? Because we have to discuss that with idiv.

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: 1. Cmon… `tag = "type"` is not used currently. `rename_all` just for the future attributes that we will miss then.
2. The slashes would be part of the path, wouldn't they?

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: How likely is it that we ever implement a provider where we generate its type?

services/src/datasets/external/netcdfcf/ebvportal_api.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: https://github.com/geo-engine/geoengine/blob/86d9754c3584ac17cd40f680f9093522d5b6f951/services/src/datasets/external/netcdfcf/ebvportal_provider.rs

…has no tests, or are you referring to something else?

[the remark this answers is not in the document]
michaelmattig: I see tests https://github.com/geo-engine/geoengine/blob/a92ebe88f86821a470b90502d4cb8e370cce7373/services/src/datasets/external/netcdfcf/ebvportal_provider.rs

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
michaelmattig: I can't give you a number here.

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
michaelmattig: so I'd have to open the netcdf here and look if the top level group has groups. If so, it's a scenario and otherwise a metric?

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
michaelmattig: I meant in the EbvPortal. Would help to have a dataset for testing.

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: do you mean NetCDF_CF_? Idk actually. But what has this to do with the compilation flags?

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: that would be more code and I doubt it makes a difference performance-wise on a vector of length 3, but I can do it if you insist.

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
michaelmattig: no, this method builds the layer from a `NetCdfOverview` which I thought to be more like a NetCfdSummary and something else than the metdata.json in the overview folder?

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: 1. since when do we preemptively define proc macros that might be used at some point?
2. how would you know whether a part of the url is a foldername or a scenarioname?

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: 9d0604e03b5a4ad399543e26b02c0f5a2490073b

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: I see no other way than to look into the (meta)data.

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: cf. Slack

> Hier wäre es statisch:
https://github.com/geo-engine/geoengine/blob/38aeafb68f13f41b87b822e437b7193d2f908a87/operators/src/engine/operator.rs#L203
https://github.com/geo-engine/geoengine/blob/38aeafb68f13f41b87b822e437b7193d2f908a87/operators/src/plot/box_plot.rs#L217
https://github.com/geo-engine/geoengine/blob/36582bd457e09bf2a0a11ea81b9ae53f0c3544e8/operators/src/source/gdal_source/mod.rs#L489
geht anscheinend auch ganz cool als const type
Ist halt die Frage, ob es hier anders sein kann, aber ich dachte nicht :schulterzucken:

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: I guess this one: https://portal.geobon.org/ebv-detail?id=28

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: It would make sense to have a generic NetCDF-CF provider.
However, relying on the EBV standard (EBV 1.0) means we always have entities and they are always encoded in the array `ebv_cube`. This is not generic, so I would pack it under the `ebv` flag.

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: Okay, keep it.

```Rust
use std::collections::VecDeque;

use criterion::{black_box, Criterion};
use criterion::{criterion_group, criterion_main};

pub fn vec_benchmark(c: &mut Criterion) {
    fn bench(values: [String; 3]) -> Vec<String> {
        let mut v: Vec<String> = Vec::from(values);
        while !v.is_empty() {
            v.remove(0);
        }
        v
    }

    c.bench_function("vec", |b| {
        b.iter(|| {
            bench(black_box([
                "foo".to_string(),
                "bar".to_string(),
                "baz".to_string(),
            ]))
        })
    });
}

pub fn vecdeque_benchmark(c: &mut Criterion) {
    fn bench(values: [String; 3]) -> VecDeque<String> {
        let mut v: VecDeque<String> = VecDeque::from(values);
        while !v.is_empty() {
            v.pop_front();
        }
        v
    }

    c.bench_function("vecdeque", |b| {
        b.iter(|| {
            bench(black_box([
                "foo".to_string(),
                "bar".to_string(),
                "baz".to_string(),
            ]))
        })
    });
}

pub fn vec_rev_benchmark(c: &mut Criterion) {
    fn bench(values: [String; 3]) -> Vec<String> {
        let mut v: Vec<String> = Vec::from(values);
        v.reverse();
        while !v.is_empty() {
            v.pop();
        }
        v
    }

    c.bench_function("vec_reverse", |b| {
        b.iter(|| {
            bench(black_box([
                "foo".to_string(),
                "bar".to_string(),
                "baz".to_string(),
            ]))
        })
    });
}

criterion_group!(
    benches,
    vec_benchmark,
    vecdeque_benchmark,
    vec_rev_benchmark
);
criterion_main!(benches);

```

```
Gnuplot not found, using plotters backend
vec                     time:   [32.511 ns 32.568 ns 32.642 ns]                 
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

vecdeque                time:   [38.337 ns 38.381 ns 38.425 ns]                      
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) low mild

vec_reverse             time:   [31.618 ns 31.656 ns 31.697 ns]                         
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
```

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: 1. You need `tag = "type"`, it is not like we serialize the other stuff currently.
2. The provider needs to know how to traverse or extract, I guess.

michaelmattig: 1. ok
2. shall I look into this then? I don't know if that is possible with actix and this would then also mean more additional logic to navigate the layer because we know have to figure out what a path is and what not. I think it would make the API much cleaner but I am not sure whether we need this in order for this PR to get merged.

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: e42fe9a87e42e94869a4a247f12f0baf073e80ae

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: e42fe9a87e42e94869a4a247f12f0baf073e80ae

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
michaelmattig: the last example was done this way because Rust wouldn't allow a static method on a trait that would be used as a boxed object.

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
michaelmattig: eccd39aabd804b43fc4cba1dbc3bf8769e8e24a3

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
michaelmattig: eccd39aabd804b43fc4cba1dbc3bf8769e8e24a3

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: 1. I guess this is better, because for groups you have `Groups: { … }`.
2. https://github.com/actix/actix-web/issues/1595

services/src/datasets/external/netcdfcf/ebvportal_provider.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: I see, will you change it then to be like the other APIs?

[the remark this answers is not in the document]
michaelmattig: d688d27d058520590acd7f996c120375b45bffeb

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: 143e2530400b1f0118d231d0b767c347c1982788

ChristianBeilschmidt (dismissed):

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #592, opened 2022-08-15
remove overviews

ChristianBeilschmidt:


ChristianBeilschmidt:
[[GEO BON] 1.x Verbesserungen bei Overviews](https://trello.com/c/r8WG7OtS/874-geo-bon-1x-verbesserungen-bei-overviews)

services/src/handlers/ebv.rs
```diff
@@ -51,8 +51,9 @@ where
             web::resource("/dataset/{id}/subdatasets")
                 .route(web::get().to(get_ebv_subdatasets::<C>)),
         )
-        .service(web::resource("/create_overviews").route(web::post().to(create_overviews::<C>)))
-        .service(web::resource("/create_overview").route(web::post().to(create_overview::<C>)));
+        .service(web::resource("/createOverviews").route(web::post().to(create_overviews::<C>)))
+        .service(web::resource("/createOverview").route(web::post().to(create_overview::<C>)))
+        .service(web::resource("/removeOverview").route(web::post().to(remove_overview::<C>)));
```
michaelmattig, line 56: why not use `DELETE` http method?

michaelmattig, line 56: why not use DELETE http method?

services/src/handlers/ebv.rs
```diff
@@ -1464,4 +1545,98 @@ mod tests {
...
+        assert_eq!(res.status(), 200, "{:?}", res.response());
+
+        let task_response =
+            serde_json::from_str::<TaskResponse>(&read_body_string(res).await).unwrap();
+
+        wait_for_task_to_finish(ctx.tasks(), task_response.task_id).await;
+
+        let status = ctx.tasks().status(task_response.task_id).await.unwrap();
+
+        assert_eq!(
```
michaelmattig, line 1634: maybe also check if the files were actually deleted

services/src/datasets/external/netcdfcf/error.rs
```diff
@@ -169,4 +169,14 @@ pub enum NetCdfCf4DProviderError {
     CannotGenerateLoadingInfo {
         source: Box<dyn ErrorSource>,
     },
+    CannotCreateInProgressFlag {
+        source: Box<dyn ErrorSource>,
+    },
+    CannotRemoveInProgressFlag {
+        source: Box<dyn ErrorSource>,
+    },
+    CannotRemoveOverviewsWhileCreationIsInProgress,
```
michaelmattig, line 178: one could argue that we should abort the creation process in that case but I guess it's ok to return an error

services/src/handlers/ebv.rs
```diff
@@ -616,6 +628,75 @@ impl<C: Context> Task<C::TaskContext> for EvbOverviewTask<C> {
 
 impl TaskStatusInfo for NetCdfCfOverviewResponse {}
 
+#[derive(Debug, Deserialize)]
+struct RemoveOverviewParams {
+    file: PathBuf,
```
michaelmattig, line 633: wouldn't it be better to give the dataset name or id?

ChristianBeilschmidt: Currently, the Path is the identification… it is not ideal, but that is also used for lookups currenty…

services/src/datasets/external/netcdfcf/error.rs
ChristianBeilschmidt: Yeah, so maybe you want that or maybe not… I would leave it as is if it is okay for you.

services/src/handlers/ebv.rs
ChristianBeilschmidt: [bd83581](https://github.com/geo-engine/geoengine/pull/592/commits/bd835810a808f93fa3666b9d715a6b03f9c3ea6f)

services/src/handlers/ebv.rs
ChristianBeilschmidt: [bd83581](https://github.com/geo-engine/geoengine/pull/592/commits/bd835810a808f93fa3666b9d715a6b03f9c3ea6f)

michaelmattig (approved):

datatypes/src/util/any.rs
```diff
@@ -13,3 +13,20 @@ where
         self
     }
 }
+
+/// Easy `Any`-casting for `Arc`s by propagating the call to the underlying implementor
+pub trait AsAnyArc {
+    /// Returns the required Arc type for calling `Arc::downcast`
+    fn as_any_arc(self: Arc<Self>) -> Arc<(dyn Any + Send + Sync)>;
```
michaelmattig, line 20: producing an Arc only makes sense if we need more than one reference right? Do you really need that? Otherwise a &dyn Any would suffice, or would it not?

services/src/datasets/external/netcdfcf/overviews.rs
```diff
@@ -467,6 +467,17 @@ pub fn remove_overviews(dataset_path: &Path, overview_path: &Path, force: bool)
 
     fs::remove_dir_all(&out_folder_path).boxed_context(error::CannotRemoveOverviews)?;
 
+    // traverse up and remove folder if empty
```
michaelmattig, line 470: can you not just defer the creation of the folder until you successfully opened the gdal dataset?

ChristianBeilschmidt: [74e9b52](https://github.com/geo-engine/geoengine/pull/592/commits/74e9b52580aa2221f3c645b177c20cf5f84d8fff)

datatypes/src/util/any.rs
ChristianBeilschmidt: We don't produce one. We consume one and basically just cast.
And I can't say self: `&Arc<Self>`, or can I?

michaelmattig (approved):

michaelmattig:
bors merge

PR #593, opened 2022-08-16
fix DateTime SQL

michaelmattig:


michaelmattig:
[In der Pro-Version gibt die Session-API falsche Datumswerte aus](https://trello.com/c/M6RmwixO/891-in-der-pro-version-gibt-die-session-api-falsche-datumswerte-aus)

ChristianBeilschmidt:
Can you add a test that verifies this fix? It is always concerning if we change behavior and no tests need to be adjusted :smile:.

michaelmattig:
> Can you add a test that verifies this fix? It is always concerning if we change behavior and no tests need to be adjusted 😄.

I just delegate the method and the actual implementation is tested in the postgres_types crate. Doesn't this suffice?

michaelmattig:
> Can you add a test that verifies this fix? It is always concerning if we change behavior and no tests need to be adjusted 😄.

unfortunately, I can't. The config for the postgres isn't accessible in the datatypes crate. I guess it's ok though because I only delegate the method to the implementation in postgres_types.

ChristianBeilschmidt:
From Slack: Test the functionality in the services crate.

michaelmattig:
I indirectly test it now as part of the method where we first noticed the error: 773c15bffb399b731d74d5dff5eea73e6c5d3f2b

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #597, opened 2022-08-18
Test_suggest

michaelmattig:


michaelmattig:
[Test suggest api method](https://trello.com/c/sWKW8bhZ/954-test-suggest-api-method)

services/src/handlers/datasets.rs
```diff
@@ -1619,4 +1619,139 @@ mod tests {
...
+                    }
+                  },
+                  "time": null,
+                  "bbox": null
+                }
+              }
+            })
+        );
+
+        drop(test_data);
```
ChristianBeilschmidt, line 1754: You don't need to call drop here because of the RAII principle.
It drops when the function returns.

ChristianBeilschmidt (changes requested):

services/src/handlers/datasets.rs
michaelmattig: f92b59b66fd6324df6d5949d740245d9227a411a

ChristianBeilschmidt (approved):

PR #598, opened 2022-08-18
Openapi

michaelmattig:


michaelmattig:
[API-Docs erzeugen](https://trello.com/c/g0zsro2u/738-api-docs-erzeugen)

services/Cargo.toml
```diff
@@ -76,6 +76,8 @@ url = { version = "2.2", features = ["serde"] }
 uuid = { version = "1.1", features = ["serde", "v4", "v5"] } # must be compatible with `bb8-postgres`
 walkdir = "2.3"
 zip = "0.6"
+utoipa = { version = "1.1.0", features = ["actix_extras", "uuid"] }
+utoipa-swagger-ui = { version = "1.1.0", features = ["actix-web"] }
```
ChristianBeilschmidt, line 80: please order ist lexicographically

services/src/handlers/plots.rs
```diff
@@ -294,7 +294,7 @@ mod tests {
         ];
         let req = actix_web::test::TestRequest::get()
             .uri(&format!(
-                "/plot/{}/?{}",
+                "/plot/{}?{}",
```
ChristianBeilschmidt, line 297: why does our API change with this PR?

services/src/server.rs
```diff
@@ -99,7 +102,6 @@ where
                     .handler(http::StatusCode::METHOD_NOT_ALLOWED, render_405),
             )
             .wrap(TracingLogger::<CustomRootSpanBuilder>::new())
-            .wrap(middleware::NormalizePath::trim())
```
ChristianBeilschmidt, line 102: 1. We should be careful removing all normalization, especially since you did not remove that from the pro server.

2. Can't we specify a normalization that makes sense for us?
Moreover, can we put swagger outside of the normalization?

services/src/server.rs
```diff
@@ -112,7 +114,11 @@ where
             .configure(handlers::wcs::init_wcs_routes::<C>)
             .configure(handlers::wfs::init_wfs_routes::<C>)
             .configure(handlers::wms::init_wms_routes::<C>)
-            .configure(handlers::workflows::init_workflow_routes::<C>);
+            .configure(handlers::workflows::init_workflow_routes::<C>)
+            .service(
+                SwaggerUi::new("/swagger-ui/{_:.*}")
+                    .url("/api-doc/openapi.json", ApiDoc::openapi()),
```
ChristianBeilschmidt, line 120: Is this name part of a standard? If not, we should camelCase it.

services/src/util/tests.rs
```diff
@@ -191,7 +191,6 @@ pub async fn send_test_request<C: SimpleContext>(
                     .handler(http::StatusCode::NOT_FOUND, render_404)
                     .handler(http::StatusCode::METHOD_NOT_ALLOWED, render_405),
             )
-            .wrap(middleware::NormalizePath::trim())
```
ChristianBeilschmidt, line 194: cf. other comment about normalization

datatypes/src/primitives/query_rectangle.rs
```diff
@@ -38,3 +38,48 @@ impl From<QueryRectangle<BoundingBox2D>> for QueryRectangle<SpatialPartition2D>
         }
     }
 }
+
+/// manual implementation, because derivation can't handle the `SpatialBounds` generic (yet)
+impl utoipa::Component for QueryRectangle<SpatialPartition2D> {
```
ChristianBeilschmidt, line 43: is it reasonable to use utoipa in the datatypes crate?
Or does this reflect a bad design of us? Actually, it should only be part of services?

operators/src/engine/operator.rs
```diff
@@ -169,6 +169,43 @@ pub enum TypedOperator {
     Plot(Box<dyn PlotOperator>),
 }
 
+impl utoipa::Component for TypedOperator {
```
ChristianBeilschmidt, line 172: same for the operators crate

services/src/apidoc.rs
```diff
@@ -0,0 +1,133 @@
...
+struct ApiDocInfo;
+
+impl Modify for ApiDocInfo {
+    fn modify(&self, openapi: &mut utoipa::openapi::OpenApi) {
+        openapi.info.title = "Geo Engine API".to_string();
+
+        openapi.info.contact = Some(
+            utoipa::openapi::ContactBuilder::new()
+                .name(Some("Geo Engine Developers"))
+                .email(Some("dev@geoengine.de"))
```
ChristianBeilschmidt, line 120: Should/can we link to the repo here?

ChristianBeilschmidt (changes requested):

services/src/handlers/plots.rs
michaelmattig: it doesn't change, the test does. In production we already use it without a terminating slash.

services/src/server.rs
michaelmattig: 1. the normalization regards only the trailing slashes. IMHO it's okay that it works only as we define it in our API, either with, or without a trailing slash, and not the other way around.

2. Not via a middleware, because middleware are application-wide

services/src/server.rs
michaelmattig: > The OpenAPI Wiki [recommends](https://github.com/OAI/OpenAPI-Specification/wiki#server) using an /api-docs endpoint, at least for server APIs. I've seen several sites in the wild that use that, and it's our shop standard.

(https://stackoverflow.com/questions/41660658/openapi-or-swagger-json-auto-discovery)

it seems to be the default. We could just put it under /openapi.json directly. or /doc/openapi.json. If we do a non-standard path anyway I would go with one of the two instead of apiDocs

datatypes/src/primitives/query_rectangle.rs
michaelmattig: it's definitely suboptimal that our API data types are scattered the way they are. However, I am not sure if we should really duplicate everything we need in services. I guess we have to discuss this.

services/src/apidoc.rs
michaelmattig: it's possible to add a url. But I am not sure if we want to link the repo, or our website. Especially in deployments we might want to point to the operator of the instance 🤷

services/Cargo.toml
michaelmattig: bc560471977af3d817753652fa8d016ac4eaf23b

services/src/server.rs
ChristianBeilschmidt: Okay, but then we should do it at `/api-docs` and not `/api-doc`, right?

michaelmattig: 2692ca2c8fa4cdc131ad683c53c7654ab5a6259d

services/src/apidoc.rs
```diff
@@ -1,89 +1,92 @@
...
+            TypedResultDescriptor,
+            PlotResultDescriptor,
+            RasterResultDescriptor,
+            VectorResultDescriptor,
+            VectorColumnInfo,
+            RasterDatasetFromWorkflow,
+            RasterDatasetFromWorkflowResult,
+            RasterQueryRectangle,
+            // VectorQueryRectangle,
+            // PlotQueryRectangle,
```
ChristianBeilschmidt, line 88: I guess you forgot those.

services/src/datasets/external/pangaea/mod.rs
```diff
@@ -420,17 +422,17 @@ mod tests {
 
         let meta: Box<
             dyn MetaData<OgrSourceDataset, VectorResultDescriptor, VectorQueryRectangle>,
-        > = provider.meta_data(&id).await.unwrap();
+        > = provider.meta_data(&id.clone().into()).await.unwrap();
```
ChristianBeilschmidt, line 425: Is it possible to implement Borrow or AsRef here for the id?

services/src/handlers/mod.rs
```diff
@@ -78,7 +78,7 @@ pub fn get_token(req: &HttpRequest) -> Result<SessionId> {
     let scheme = Bearer::parse(header).map_err(|_| Error::Authorization {
         source: Box::new(Error::InvalidAuthorizationScheme),
     })?;
-    SessionId::from_str(scheme.token()).map_err(|err| Error::Authorization {
-        source: Box::new(Error::InvalidUuid { source: err }),
+    SessionId::from_str(scheme.token()).map_err(|_err| Error::Authorization {
+        source: Box::new(Error::InvalidUuid),
```
ChristianBeilschmidt, line 82: Why don't we need the inner error anymore?
Could be good for debugging?

services/src/handlers/workflows.rs
```diff
@@ -348,13 +353,13 @@ async fn resolve_provenance<C: Context>(
 }
 
 /// parameter for the dataset from workflow handler (body)
-#[derive(Clone, Debug, Deserialize, Serialize, Component)]
-#[component(example = json!({"name": "foo", "description": null, "query": {"spatialBounds": {"upperLeftCoordinate": {"x": -10.0, "y": 80.0}, "lowerRightCoordinate": {"x": 50.0, "y": 20.0}}, "timeInterval": {"start": 1_388_534_400_000_i64, "end": 1_388_534_401_000_i64}, "spatialResolution": {"x": 0.1, "y": 0.1}}}))]
+#[derive(Clone, Debug, Deserialize, Serialize, ToSchema)]
```
ChristianBeilschmidt, line 356: Just for curiosity: What is the difference?

services/src/pro/apidoc.rs
```diff
@@ -0,0 +1,147 @@
...
+            TypedResultDescriptor,
+            PlotResultDescriptor,
+            RasterResultDescriptor,
+            VectorResultDescriptor,
+            VectorColumnInfo,
+            RasterDatasetFromWorkflow,
+            RasterDatasetFromWorkflowResult,
+            RasterQueryRectangle,
+            // VectorQueryRectangle,
+            // PlotQueryRectangle,
```
ChristianBeilschmidt, line 99: Again, forgot those.

services/src/projects/project.rs
```diff
@@ -119,10 +121,28 @@ impl Project {
...
+//             "x": 180.0,
+//             "y": 90.0
+//         }
+//     },
+//     "spatialReference": "EPSG:4326",
+//     "timeInterval": {
+//         "end": 1_388_534_400_000_i64,
+//         "start": 1_388_534_400_000_i64
+//     }
+// }))]
```
ChristianBeilschmidt, line 145: I don't quite get it. This has nothing to do with database stuff, right?

ChristianBeilschmidt (changes requested):

services/src/apidoc.rs
michaelmattig: I commented them out because swaggerhub complains about these types being unused. But I already added the schema definition and we will use them in the future, so I commented them out instead of deleting. But I can also delete the lines.

services/src/datasets/external/pangaea/mod.rs
michaelmattig: I don't think it is possible because it expects reference to something that owns the fields (of the `DataId`), but I could only create something that borrows the fields. But if you know a solution that would be nice.

services/src/handlers/mod.rs
michaelmattig: I couldn't keep it, because the source error previously was from the datatypes (or operators) crate, but now it has to be in the services crate and that would mean I'd have to box it. I can do it, but basically this error is thrown if a string is no a uuid, so there is no real information to get here.

services/src/handlers/workflows.rs
michaelmattig: it's just renamed in utoipa to more closely match the OpenAPI specification on what a component and what a schema is.

services/src/projects/project.rs
michaelmattig: utoipa's dervice macro doesn't work together with the To/FromSql macros. I opened an issue at utoipa regarding this: https://github.com/juhaku/utoipa/issues/266

ChristianBeilschmidt: can you reference the issue in the TODO?

services/src/handlers/mod.rs
ChristianBeilschmidt: Why? It wasn't part of the API because it was not part of the display string.

services/src/apidoc.rs
ChristianBeilschmidt: Does this result in the project not being builded?

michaelmattig: no the project builds fine but the resulting openapi.json generates warnings when I upload them to swaggerhub. Because you aren't supposed to include unreferenced components

services/src/handlers/mod.rs
michaelmattig: the error returned from this function was a operators-crate error. Now, because of the changes in the IDs etc., it is a services-crate error. And that one I can no longer use as a source field in the error, because that would be self-referential, aka not allowed because you can no longer compute the size of the Error enum. As I said, I can box it and put it back but I don't see why we would need to know the UUID error, it should suffice to know that the input token is not a valid UUID

services/src/projects/project.rs
michaelmattig: bdce6dee4fb6d21e20c40435a4695c572eacac14

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors merge

PR #602, opened 2022-08-26
Gdal Source read up-side-down data

jdroenner:


jdroenner:
[Enable GDAL source to read up-side-down rasters](https://trello.com/c/Q1hYywdL/959-enable-gdal-source-to-read-up-side-down-rasters)

datatypes/src/raster/grid.rs
```diff
@@ -319,6 +319,29 @@ where
     pub fn inner_ref(&self) -> &Vec<T> {
         &self.data
     }
+
+    /// reverse this grid along the y-axis. Returns an "up-side-down" `Grid`.
+    #[must_use]
+    pub fn reversed_y_axis_grid(&self) -> Grid<D, T>
```
michaelmattig, line 325: Are there also files where the x axis is inverted in the tiff? Then we would need a method to invert that as well

operators/src/source/gdal_source/mod.rs
```diff
@@ -144,6 +161,95 @@ pub struct GdalDatasetGeoTransform {
...
+            ((coord.y - self.origin_coordinate.y) / self.y_pixel_size).floor() as isize;
+
+        [grid_y_index, grid_x_index].into()
+    }
+
+    fn spatial_partition_to_read_window(
+        &self,
+        spatial_partition: &SpatialPartition2D,
+    ) -> GdalReadWindow {
+        // choose the epsilon relative to the pixel size
```
michaelmattig, line 213: maybe explain why you do this. I guess in order to not read the next pixel, because gdals reads inclusive and our partitions are exlusive? If so we should need the epsilon only on the bottom right corner.

operators/src/source/gdal_source/mod.rs
```diff
@@ -144,6 +161,95 @@ pub struct GdalDatasetGeoTransform {
     pub y_pixel_size: f64,
 }
 
+impl GdalDatasetGeoTransform {
+    pub fn standardized_spatial_partition(
+        &self,
+        x_size: usize,
+        y_size: usize,
+    ) -> SpatialPartition2D {
+        // The geo transform of a gdal dataset might have flipped axis. So correct that.
```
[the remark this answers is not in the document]
michaelmattig: also: can't you, instead, just compute the window without the epsilon and check if the resulting pixels fall out of the the partition and subtract one then?

operators/src/source/gdal_source/mod.rs
```diff
@@ -288,6 +387,7 @@ impl GdalRasterLoader {
         tile_time: TimeInterval,
     ) -> Result<RasterTile2D<T>> {
         match dataset_params {
+            // TODO: discuss if we need this check here. The metadata provider should only pass on loading infos if the query intersects the datasets bounds! And the tiling strategy should only generate tiles that intersect the querys bbox.
```
michaelmattig, line 390: we could ensure it in the metadataprovider and only do an debug_assert here. But I don't think we do that so might as well check it here and be safe?

operators/src/source/gdal_source/mod.rs
```diff
@@ -450,22 +547,32 @@ where
             start.elapsed()
         );
 
-        // TODO: evaluate if there are GeoTransforms with positive y-axis
-        // The "Pixel-space" starts at the top-left corner of a `GeoTransform`.
+        // A `GeoTransform` maps pixel space to world space.
+        // We are not aware of spatial reference systems where the x-axis points tot he right.
```
michaelmattig, line 551: typo "tot he"

operators/src/source/gdal_source/mod.rs
```diff
@@ -450,22 +547,32 @@ where
...
+        // However, there are spatial reference systems where the y-axis points downwards.
```
michaelmattig, line 552: do you mean upwards? I thought downwards was the norm.

operators/src/source/gdal_source/mod.rs
```diff
@@ -588,33 +695,36 @@ impl InitializedRasterOperator for InitializedGdalSourceOperator {
...
-        None,                             // sampling mode
+        read_window.gdal_window_start(), // pixelspace origin
+        read_window.gdal_window_size(),  // pixelspace size
+        gdal_out_shape,                  // requested raster size
+        None,                            // sampling mode
     )?;
-    let data_grid = Grid::new(tile_grid.clone(), buffer.data)?;
+    let data_grid = Grid::new(out_shape.clone(), buffer.data)?;
+
+    let data_grid = if flip_y_axis {
```
michaelmattig, line 723: would it make sense to do this outside of this function, and only do the actual reading here? Also: maybe implement the reverse on the MaskedGrid so that it only has to be called once?

operators/src/source/gdal_source/mod.rs
```diff
@@ -683,15 +806,21 @@ where
...
     tile_info: TileInformation,
     dataset: &gdal::Dataset,
-    dataset_bounds: SpatialPartition2D,
-    dataset_geo_transform: GeoTransform,
     dataset_params: &GdalDatasetParameters,
 ) -> Result<GridOrEmpty2D<T>>
 where
     T: Pixel + GdalType + Default + FromPrimitive,
 {
+    // this are the bounds from the datasets def!
```
michaelmattig, line 814: typo "this" -> "these"

michaelmattig: but maybe just remove the commented out code

operators/src/source/gdal_source/mod.rs
```diff
@@ -701,21 +830,37 @@ where
...
 
-    let output_geo_transform = tile_info.tile_geo_transform();
-    let dataset_grid_bounds =
-        dataset_geo_transform.spatial_to_grid_bounds(&dataset_intersection_area);
+    let tile_geo_transform = tile_info.tile_geo_transform();
+
+    let gdal_read_window =
+        gdal_dataset_geotransform.spatial_partition_to_read_window(&dataset_intersection_area);
+
+    let is_y_axis_flipped = tile_geo_transform.y_pixel_size().is_sign_negative()
```
michaelmattig, line 838: maybe use the "y_axis_is_increasing" method instead

michaelmattig: and: why is do you do the comparison? I don't understand it.

operators/src/source/gdal_source/mod.rs
```diff
@@ -1616,4 +1752,141 @@ mod tests {
...
+        let sb = gt.standardized_spatial_partition(10, 10);
+
+        let exp = SpatialPartition2D::new(Coordinate2D::new(5., 0.), Coordinate2D::new(10., -5.))
+            .unwrap();
+
+        assert_eq!(sb, exp);
+    }
+
+    #[test]
+    fn gdal_read_window_a() {
```
michaelmattig, line 1821: I guess we need a test that reads from a flipped tiff in order to test if it all actually works

operators/src/source/gdal_source/mod.rs
jdroenner: we could remove the geotransform and dataset size from the json files if we remove it here and rely on the dataset bounds.

operators/src/source/gdal_source/mod.rs
jdroenner: i thought about this and if we do that we need to reverse the mask also. even if it was generated from no-data value or if the data is always valid.

operators/src/source/gdal_source/mod.rs
jdroenner: you mean this:     `let is_y_axis_flipped = tile_geo_transform.y_pixel_size().is_sign_negative()!= gdal_dataset_geotransform.y_pixel_size.is_sign_negative();` ? This is how we find out if the data is stored up-side-down. The sign of the y-axis in the geotransform indicates how the pixels are ordered. If the order is not the same it is up-side-down.

datatypes/src/primitives/spatial_partition.rs
```diff
@@ -232,6 +232,16 @@ impl SpatialPartition2D {
             ),
         }
     }
+
+    /// Returns true if the x value of the left edge is smaller then the x value of the right edge. Spatial x-axis "points right".
+    pub fn x_axis_is_increasing(&self) -> bool {
+        self.upper_left_coordinate.x <= self.lower_right_coordinate.x
```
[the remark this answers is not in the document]
jdroenner: i removed it since the axis order is not related to the SpatialPartition but to the GeoTransform. [feb6a53](https://github.com/geo-engine/geoengine/pull/602/commits/feb6a534eb0c19d7b3f55e8c6714186ac1e17150)

datatypes/src/raster/grid.rs
jdroenner: i have not seen that. However we can add it the same way if we see some in the wild.

operators/src/source/gdal_source/mod.rs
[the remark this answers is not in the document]
jdroenner: [2ca4d35](https://github.com/geo-engine/geoengine/pull/602/commits/2ca4d35f8820be9271bbe1b9f678efd207605762)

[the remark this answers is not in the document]
jdroenner: [8fd3b20](https://github.com/geo-engine/geoengine/pull/602/commits/8fd3b20b83382d3b413047c666ec86f744e593b1)

operators/src/source/gdal_source/mod.rs
```diff
@@ -144,6 +161,95 @@ pub struct GdalDatasetGeoTransform {
...
+    fn spatial_partition_to_read_window(
+        &self,
+        spatial_partition: &SpatialPartition2D,
+    ) -> GdalReadWindow {
+        // choose the epsilon relative to the pixel size
+        const EPSILON: f64 = 0.000_001;
+        let epsilon: Coordinate2D =
+            (self.x_pixel_size * EPSILON, self.y_pixel_size * EPSILON).into();
+
+        let (near_origin_coord, far_origin_coord) = if self.y_pixel_size.is_sign_negative() {
```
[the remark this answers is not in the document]
jdroenner: [806f037](https://github.com/geo-engine/geoengine/pull/602/commits/806f037021d196ad8cc9c337617bf9f9b10f0f5a)

operators/src/source/gdal_source/mod.rs
jdroenner: [8fd3b20](https://github.com/geo-engine/geoengine/pull/602/commits/8fd3b20b83382d3b413047c666ec86f744e593b1)

operators/src/source/gdal_source/mod.rs
jdroenner: i think i removed that in one of the commits

operators/src/source/gdal_source/mod.rs
jdroenner: [313204f](https://github.com/geo-engine/geoengine/pull/602/commits/313204fa2ba923767410b29e2bbc6bafc5014a92)

operators/src/source/gdal_source/mod.rs
jdroenner: [feb6a53](https://github.com/geo-engine/geoengine/pull/602/commits/feb6a534eb0c19d7b3f55e8c6714186ac1e17150)

operators/src/source/gdal_source/mod.rs
```diff
@@ -115,7 +114,7 @@ pub enum TimeReference {
 pub struct GdalDatasetParameters {
     pub file_path: PathBuf,
     pub rasterband_channel: usize,
-    pub geo_transform: GdalDatasetGeoTransform,
+    pub geo_transform: GdalDatasetGeoTransform, // TODO: discuss if we need this at all
```
michaelmattig, line 117: Maybe to be able to read raster files (images) without geotransform? 
Or: to avoid reading the metadata from disk every time. But maybe gdal reads it anyway.

operators/src/source/gdal_source/mod.rs
```diff
@@ -144,6 +163,120 @@ pub struct GdalDatasetGeoTransform {
...
+        // World coordinates and pixel sizes use float values. Since the float imprecision might cause overflowing into the next pixel we use an epsilon to correct values very close the pixel borders. This logic is the same as used in [`GeoTransform::grid_idx_to_pixel_upper_left_coordinate_2d`].
+        const EPSILON: f64 = 0.000_001;
+        let epsilon: Coordinate2D =
+            (self.x_pixel_size * EPSILON, self.y_pixel_size * EPSILON).into();
+
+        /*
+        The read window is relative to the transform of the gdal dataset. The `SpatialPartition` is oriented at axis of the spatial SRS. This usually causes this situation:
+
+        The gdal data is stored with negative pixel size. The "ur" coordinate of the `SpatialPartition` is neareest to the origin of the gdal raster data.
+        ur                      ul
```
michaelmattig, line 219: why is `ur` in the left?

operators/src/source/gdal_source/mod.rs
michaelmattig: how does this get rid of the option?

operators/src/source/gdal_source/mod.rs
```diff
@@ -442,31 +558,62 @@ where
...
+
+        if let Some(data_spatial_bounds) = result_descriptor.bbox {
+            if !data_spatial_bounds.intersects(&query.spatial_bounds) {
+                debug!("query does not intersect spatial data bounds");
+                empty = true;
+            }
+        }
+
+        // TODO: use the time bounds to early return.
+        /*
```
michaelmattig, line 598: why is this commented out?

operators/src/source/gdal_source/mod.rs
michaelmattig: but in Geo Engine the origin of a tile is always in the top left corner or not? Then tile_geo_transform y_pixelsize should always be negative and we only need to look at the dataset_geo_transform?

And what about the case where the CRS axis are flipped?

operators/src/source/gdal_source/mod.rs
```diff
@@ -1616,4 +1789,199 @@ mod tests {
...
+
+        assert!(!grid.is_empty());
+
+        let grid = grid.into_materialized_masked_grid();
+
+        assert_eq!(grid.inner_grid.data.len(), 64);
+        assert_eq!(
+            grid.inner_grid.data,
+            &[
+                255, 255, 255, 255, 255, 255, 255, 255, 255, 75, 37, 255, 44, 34, 39, 32, 255, 86,
```
michaelmattig, line 1977: How did you verify the data? Maybe you could add a dataset definition so that we could see in the UI that it's correct?

operators/src/source/gdal_source/mod.rs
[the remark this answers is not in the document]
michaelmattig: what do you think about the suggestion of having two separate data structures for SpatialPartitions? Or at least having a `is_standardized()` method and debug assert everywhere, where we expect it to not be flipped, that it isn't.

michaelmattig (commented):
.

operators/src/source/gdal_source/mod.rs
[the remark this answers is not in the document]
jdroenner: the SpatialPartition is never flipped. As I see it the SpatialPartition does not know about directions. Only a GeoTransform can indicate the direction the pixels are stored. We have to sperate a few things here.
A GeoTransform translate from pixel space to world space. The pixel size sign (y) indicates how the raster rows are stored in your memory layout.
SRS coordinates in map space are usually arranged like your normal coordinate system. However, there might be the case that a SRS has an up -> down y-axis. Then our SpatialPartition will not work. Well, it will work but the edge at lower left is actually the edge with the smallest values. It will then be at the top-left and the SpatialPartition is up-side-down.

operators/src/source/gdal_source/mod.rs
jdroenner: Well i would argue that there might be a strange SRS where the y-axis is pointing down. Then the pixel size in the GeoTransform we use for our context will be positive. And there might still be a tiff where the datas geotransform has a pixel size with a negative sign

operators/src/source/gdal_source/mod.rs
jdroenner: i do not know if gdal supports data without a geotransform. It guess it will report a pixel size of 1,-1. How can we decide what is right? Or will the json overrule the dataset? I guess this is part of the discussion what a dataset def really is.

operators/src/source/gdal_source/mod.rs
```diff
@@ -133,9 +132,29 @@ pub struct GdalDatasetParameters {
...
+
+    fn gdal_window_size(&self) -> (usize, usize) {
+        (self.read_size_x, self.read_size_y)
+    }
+}
+
 /// A user friendly representation of Gdal's geo transform. In contrast to [`GeoTransform`] this
 /// geo transform allows arbitrary pixel sizes and can thus also represent rasters where the origin is not located
 /// in the upper left corner. It should only be used for loading rasters with Gdal and not internally.
+/// The GDAL pixel space is usually anchored at the "top-left" corner of the data spatial bounds. Therefore the raster data is stored with spatial coordinate y-values decreasing with the rasters rows. This is represented by a negative pixel size.
```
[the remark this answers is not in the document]
jdroenner: as i wrote above. I think it is not correct to limit it to negative signs.

operators/src/source/gdal_source/mod.rs
jdroenner: it will generate too large empty time-intervals.

operators/src/source/gdal_source/mod.rs
```diff
@@ -442,31 +558,62 @@ where
...
+            if !data_time_bounds.intersects(&query.time_interval) {
+                debug!("query does not intersect temporal data bounds");
+                empty = true;
+            }
+        }
+        */
+
+        let loading_iter = if empty {
+            GdalLoadingInfoTemporalSliceIterator::Static {
+                parts: vec![].into_iter(),
```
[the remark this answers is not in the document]
jdroenner: no the type will not be the same.

operators/src/source/gdal_source/mod.rs
jdroenner: it is a flipped version of the dataset we use for all the other tests. it reads the flipped data and produces the same output as the non-flipped data.

operators/src/source/gdal_source/mod.rs
jdroenner: currently the dataset bounds are stored twice in the json file. once as bbox in the result descriptors and as geotransform and size in pixels.

jdroenner: and for time this should be possible using the time bound of the dataset. if the query does not intersect the time interval it should produce empty data. At the same time, the metadata loadinginfo loading provider provider should skip empty tiles. The emtpy tile adapter will fill it anyway.

operators/src/source/gdal_source/mod.rs
jdroenner: it simply needs more work to enable the check

operators/src/source/gdal_source/mod.rs
jdroenner: [44bd6b5](https://github.com/geo-engine/geoengine/pull/602/commits/44bd6b52f5b3a1e8f344edb0cd17692038662067)

operators/src/source/gdal_source/mod.rs
jdroenner: [add ndvi flipped y-axis def](https://github.com/geo-engine/geoengine/pull/602/commits/1ad08331349d8e7505cabfd1f9f9d6a4af1ff4f7)

operators/src/source/gdal_source/mod.rs
michaelmattig: still: why do you need to check the tile_geo_transform in order to determine that the y_axis of the tiff is flipped?

test_data/dataset_defs/ndvi_flipped_y_axis.json
```diff
@@ -0,0 +1,329 @@
...
+    },
+    "timePlaceholders": {
+      "%_START_TIME_%": {
+        "format": "%Y-%m-%d",
+        "reference": "start"
+      }
+    },
+    "dataTime": {
+      "start": "2014-01-01T00:00:00.000Z",
+      "end": "2014-07-01T00:00:00.000Z"
```
michaelmattig, line 322: should be 2014-02-01

operators/src/source/gdal_source/mod.rs
jdroenner: the GeoTransform of the tile is what we want and the GEoTransform of the data is what we get. If they don't have the same sign the data is flipped. Currently it will always be negative for the tile GeoTransform but do you really hurt to check that? We have to check the sign for the data GeoTransform anyway.

test_data/dataset_defs/ndvi_flipped_y_axis.json
jdroenner: [682bfce](https://github.com/geo-engine/geoengine/pull/602/commits/682bfce1f0a6c100a4f8e7e152aa43e377dc62e2)

michaelmattig:
bors merge

michaelmattig (approved):

michaelmattig:
bors merge

PR #603, opened 2022-08-26
change layer collection api to return collections instead of items

michaelmattig:


michaelmattig:
[[GEO BON] 1.x LayerAPI umsetzen](https://trello.com/c/W4pWeL0N/919-geo-bon-1x-layerapi-umsetzen)

services/src/layers/layer.rs
```diff
@@ -67,12 +66,15 @@ pub struct LayerDefinition {
...
-    id: LayerCollectionId,
-    name: String,
-    description: String,
-    items: Vec<CollectionItem>,
+    pub id: ProviderLayerCollectionId,
+    pub name: String,
+    pub description: String,
+    pub items: Vec<CollectionItem>,
+    pub entry_label: Option<String>, // a common label for the collection's entries, if there is any // TODO: separate labels for collections and layers?
+    pub properties: Vec<(String, String)>,
```
ChristianBeilschmidt, line 77: Why did you move this?

ChristianBeilschmidt (changes requested):

services/src/layers/layer.rs
michaelmattig: it makes much more sense this way, because they belong to the collection that is currently being listed. When I render a collection in the UI, I want to print the properties and I want to display the entryLabel. Before the change I had to take the information from the parent collection's corresponding child entry.

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #610, opened 2022-09-05
add generic scale operator to replace radiance op

jdroenner:


jdroenner:
[Offset/Scale aus Raster-Metadaten anwenden](https://trello.com/c/9RKXaqVf/743-offset-scale-aus-raster-metadaten-anwenden)

datatypes/src/raster/operations/checked_scaling.rs
```diff
@@ -0,0 +1,395 @@
...
+
+pub trait Scale
+where
+    Self: Sized,
+{
+    /// scales with `(self - offset_by) / scale_with`. Overflows produce `None`.
+    fn scale(self, scale_with: Self, offset_by: Self) -> Option<Self>;
+}
+
+macro_rules! impl_scale_conv {
```
michaelmattig, line 56: can't you implement this any differently? I mean, all the Pixel types should implement the add, mul and div traits so isn't it possible to implement this generically without macros?

jdroenner: i can write it for each type...
i use checked_div, checked_mul ... for the int types since the type can overflow. And i can't impl the trait for a generic since all types implement + - * / and i can not override it (specialization). I also can not impl it in a generic way since there is an error that there might be a checked_... impl for f32, f64 later...

michaelmattig (approved):

michaelmattig:
bors merge

PR #614, opened 2022-09-08
Calculate statistics for vector data

dbrandenstein:


domBrandenstein:
[Zählstatistiken für Vektordaten](https://trello.com/c/Sx6cTpIF/977-z%C3%A4hlstatistiken-f%C3%BCr-vektordaten)

operators/src/plot/statistics.rs
```diff
@@ -41,79 +49,228 @@ impl PlotOperator for Statistics {
...
+            .collect();
+
+        call_on_generic_vector_processor!(&self.vector, processor => {
+            let mut query = processor.query(query, ctx).await?;
+
+            while let Some(collection) = query.next().await {
+                let collection = collection?;
+
+                for (column, stats) in &mut number_statistics {
+                    collection.data(column).expect("checked in param").float_options_iter().for_each(
```
ChristianBeilschmidt, line 241: Maybe we can code defensive here and say:
If it does not exist, just `continue` (instead of panic).

operators/src/plot/statistics.rs
```diff
@@ -173,10 +330,35 @@ where
...
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct VectorColumnStatisticsOutput {
+    pub value_count: usize,
+    pub nan_count: usize,
+    pub min: f64,
+    pub max: f64,
+    pub mean: f64,
+    pub stddev: f64,
+}
```
ChristianBeilschmidt, line 343: We thought of unifying the output for rasters and vectors.
Our suggestion:

```suggestion
struct StatisticsOutput {
    pub value_count: usize,
    pub valid_count: usize,
    pub min: f64,
    pub max: f64,
    pub mean: f64,
    pub stddev: f64,
}
```

So, we use (1) value for pixel for rasters and (2) valid_count instead of nan_count, which means, we have to calculate `number_statistics.count() - number_statistics.nan_count()`.

ChristianBeilschmidt (changes requested):

operators/src/plot/statistics.rs
domBrandenstein: Isn't `number_statistics.count()` already sort of a "valid_count", because it excludes NaN values?
Wouldn't it then make more sense to have "value_count" be `number_statistics.count() + number_statistics.nan_count()` and "valid_count" be `number_statistics.count()`?

operators/src/plot/statistics.rs
domBrandenstein: Can this even happen after checking that the column exists in line 127?

operators/src/plot/statistics.rs
domBrandenstein: ... like this https://github.com/geo-engine/geoengine/pull/614/commits/dcc29d1618711ec7ca0f13d42ec7056e7129b1a4

operators/src/plot/statistics.rs
```diff
@@ -41,79 +49,228 @@ impl PlotOperator for Statistics {
...
-        Ok(initialized_operator.boxed())
+                let rasters = join_all(rasters.into_iter().map(|s| s.initialize(context))).await;
+                let rasters = rasters.into_iter().collect::<Result<Vec<_>>>()?;
+
+                let in_descriptors = rasters
+                    .iter()
+                    .map(InitializedRasterOperator::result_descriptor)
+                    .collect::<Vec<_>>();
+
+                if rasters.len() > 1 {
```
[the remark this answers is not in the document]
domBrandenstein: I added a test which confirms that an empty raster input vector leads to an empty output without reporting an error: https://github.com/geo-engine/geoengine/pull/614/commits/7d1ac4eeabf5fc180d7b2400b1c79d34edb9c794

operators/src/plot/statistics.rs
ChristianBeilschmidt: Well, it is in the result descriptor and thus in 99,999999% of the cases should be part of the feature collection (if there is no bug, etc.). But if it wouldn't be there for whatever reasons, it is defensive to code it to not panic.

Sometimes it is hard to do so, but here it should be very easy.

operators/src/plot/statistics.rs
ChristianBeilschmidt: You are right, didn't have the implementation in mind.

operators/src/plot/statistics.rs
domBrandenstein: NaN is now assumed for all values in an unexpectedly missing column (https://github.com/geo-engine/geoengine/pull/614/commits/b12212ea766ec55d63406194a8b1109136421add).
Let me know if you prefer to completely exclude the missing column values from the statistics (total count).

Also note that some other plot operators (scatter plot, box plot, ...) panic in this case. It should probably also be changed there.

ChristianBeilschmidt (dismissed):

ChristianBeilschmidt:
You need to merge the master again, sry.

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt (approved):

PR #618, opened 2022-09-09
add resolution to result descriptor

michaelmattig:


michaelmattig:
[[RESPECT M.1] Dateninterpolation](https://trello.com/c/666qNh0F/973-respect-m1-dateninterpolation)

operators/src/processing/expression/mod.rs
```diff
@@ -265,6 +265,17 @@ impl RasterOperator for Expression {
         let time = time_interval_extent(in_descriptors.iter().map(|d| d.time));
         let bbox = partitions_extent(in_descriptors.iter().map(|d| d.bbox));
 
+        let resolution = in_descriptors
+            .iter()
+            .map(|d| d.resolution)
+            .reduce(|a, b| match (a, b) {
+                (Some(a), Some(b)) => {
+                    Some(SpatialResolution::new_unchecked(a.x.min(b.x), a.y.min(b.y)))
```
ChristianBeilschmidt, line 273: That is actually strange, but I have no better solution.
You would not be able to query without having some interpolation… I don't know…

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -585,6 +585,7 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
             measurement: Measurement::Unitless,
             time: None,
             bbox: None,
+            resolution: None,
```
ChristianBeilschmidt, line 588: Don't we have one for sentinel? It should be 10m? Or do we get this from metadata?

operators/src/processing/interpolation/mod.rs
```diff
@@ -159,17 +185,16 @@ where
...
-                input_resolution: self.params.input_resolution,
-            }
-        );
+        // do not interpolate if the source resolution is already fine enough
+        if query.spatial_resolution.x >= self.input_resolution.x
+            && query.spatial_resolution.y >= self.input_resolution.y
+        {
+            // TODO: should we use the query or the input resolution here?
+            return self.source.query(query, ctx).await;
+        }
```
ChristianBeilschmidt, line 194: I guess query as you did, otherwise we cannot load less data depending on the query input.
This would mean, we always query with the lowest pyramid level right?

operators/src/processing/expression/mod.rs
michaelmattig: What do you mean?

The idea is taking the finest resolution of all inputs as the finest resolution of the output. This is the only thing that makes sense to me. What does it have to do with interpolation, though?

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: I don't know. We should maybe investigate this in a separate task.

operators/src/processing/reprojection.rs
```diff
@@ -251,6 +251,7 @@ impl RasterOperator for Reprojection {
             measurement: in_desc.measurement.clone(),
             time: in_desc.time,
             bbox,
+            resolution: None, // TODO: calculate resolution
```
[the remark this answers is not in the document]
michaelmattig: I think it is very problematic because the resolution can highly vary, depending on the location. We also have this problem when determining the resolution with which we actually query the source in the reprojection. 

I think a first approach would be to build special methods for known conversions which are very important to us. E.g. 4326<->3857 or 4326<->UTM32N. I can't think of any generic solutions that would work between arbitrary projections... Maybe reprojecting the bounds of every pixel and taking the minimum width and height, but that would be very expensive.

operators/src/processing/interpolation/mod.rs
```diff
@@ -30,10 +30,17 @@ use snafu::{ensure, Snafu};
...
-    pub input_resolution: SpatialResolution,
+    pub input_resolution: InputResolution,
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+#[serde(rename_all = "camelCase", tag = "type")]
+pub enum InputResolution {
+    Value(SpatialResolution),
+    Native,
```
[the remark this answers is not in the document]
michaelmattig: I mean the finest resolution that the operator can produce. Query it with any finer resolution and you get no more data but only interpolated data. Thus I called it `native` because it is the native resolution of the data. We can also call it `max` or `original`.

I don't think `derived` quite captures that, because you could also derive something else than the maximum reasonable resolution.

operators/src/processing/interpolation/mod.rs
michaelmattig: no, it means we query with the pyramid level that has the next finer resolution as the query. Could also be the highest pyramid level or some level in the middle.

operators/src/processing/expression/mod.rs
ChristianBeilschmidt: cf. chat, we can leave it for now

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
ChristianBeilschmidt: Can you add a TODO then?

operators/src/processing/reprojection.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Johannes said something about an algorithm from GDAL…

operators/src/processing/interpolation/mod.rs
ChristianBeilschmidt: Why the next finer? The input resolution is fixed, right?

operators/src/processing/interpolation/mod.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: cf. Slack

> Christian
  [18:27 Uhr](https://agdbs.slack.com/archives/C02JDR82DAP/p1663000043475689)
[@Michael M.](https://agdbs.slack.com/team/UBBE024JG)
 im UI nennst du das ja auch "Source Resolution". Vielleicht ist das besser als Native, denn Native ist ja komisch, wenn man eine setzt nach der Expression.


> Michael M.
  [18:32 Uhr](https://agdbs.slack.com/archives/C02JDR82DAP/p1663000372527469)
jo


> Christian
  [18:35 Uhr](https://agdbs.slack.com/archives/C02JDR82DAP/p1663000537185969)
Dann würde ich dir das noch mal assignen und danach kann man das erst einmal mergen.

operators/src/processing/interpolation/mod.rs
michaelmattig: If we would use it then yes. I don't know why it would correspond to pyramid levels, though. The user could specify any inputResolution. It's independent of any pyramids.

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: ecb4ae36da7c1cb88d8d66ecaee478c080493c56

operators/src/processing/interpolation/mod.rs
```diff
@@ -44,14 +51,9 @@ pub enum InterpolationMethod {
 #[snafu(visibility(pub(crate)), context(suffix(false)), module(error))]
 pub enum InterpolationError {
     #[snafu(display(
-        "The query resolution ({:?}) must be smaller than the input resolution ({:?})",
-        query_resolution,
-        input_resolution
+        "The input resolution was defined as `native` but the source resolution is unknown.",
```
ChristianBeilschmidt, line 54: Please check the errors

```suggestion
        "The input resolution was defined as `source` but the source resolution is unknown.",
```

ChristianBeilschmidt (changes requested):

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #626, opened 2022-09-15
derive output resolution after reprojection

michaelmattig:


michaelmattig:
[[RESPECT M.1] Dateninterpolation](https://trello.com/c/666qNh0F/973-respect-m1-dateninterpolation)

operators/src/processing/reprojection.rs
```diff
@@ -1276,4 +1294,43 @@ mod tests {
...
+                spatial_reference: out_proj.into(),
+                measurement: Measurement::Unitless,
+                time: None,
+                bbox: out_proj
+                    .area_of_use_projected::<SpatialPartition2D>()
+                    .unwrap()
+                    .into(),
+                // TODO: y resolution should be double the x resolution, but currently we only compute a uniform resolution
+                resolution: Some(SpatialResolution::new_unchecked(
+                    14_084.951_977_350_345,
```
jdroenner, line 1330: how did you verify the resolution? Compared with GDAL? Please add a comment how to do this.

operators/src/processing/reprojection.rs
```diff
@@ -47,6 +47,42 @@ pub struct RasterReprojectionState {
...
+            )?;
+
+            Some(bbox.reproject_clipped(&projector)?)
+        } else {
+            None
+        };
+
+        let out_res = match (in_desc.bbox, in_desc.resolution, out_bbox) {
+            (Some(in_bbox), Some(in_res), Some(out_bbox)) => {
+                suggest_pixel_size_from_diag_cross_projected(in_bbox, out_bbox, in_res).ok()
```
jdroenner, line 70: If i remember correct we had some problems with inconsistent resolutions in the projection when using other BBOXes then the SRS ones. Is there a similar problem here? Will the resolution be the same as the one the operator actually uses? 
Also can we use the data BBOX in the operator and fall back to the SRS BBOXes?

operators/src/processing/reprojection.rs
michaelmattig: it's very similar to what gdal would do:

> $ gdalwarp -t_srs EPSG:3857 test_data/raster/modis_ndvi/MOD13A2_M_NDVI_2014-01-01.TIFF ndvi_3857.tif
> $ gdalinfo ndvi_3857.tif
> ...
> Pixel Size = (14052.950258048738760,-14057.881117788405390)

I just used what we had like you suggested, so there isn't really a way to verify anything.

operators/src/processing/reprojection.rs
michaelmattig: > If i remember correct we had some problems with inconsistent resolutions in the projection when using other BBOXes then the SRS ones.

I only remember that we had problems when we tried to partition the space and find the max resolution of all partitions. But we no longer do this.

> Will the resolution be the same as the one the operator actually uses?

What do you mean by this? The only place where we use this resolution (for now) is as a parameter in the interpolation. If the input resolution is set to `source` then this resolution here would be used to query the projection with.

> Also can we use the data BBOX in the operator and fall back to the SRS BBOXes?

I think the in_bbox and out_bbox have to be the same area, but only in different resolutions, for pixel size suggestion to work. So I guess we could compute the intersection of the in_srs_bbox and out_srs_bbox and compute the resolution on this. Shall I do it?

Now that I think about it: don't we have to do this for the case where a data bbox is available, too? (clipping the in_bbox to the area of out_bbox)

michaelmattig: I did some brainstorming with myself and implemented three things now

- in the query method, use the data bounds from the source result descriptor for calculating the input resolution. Only if it is not given fall back to the srs area of use
- in the derivation of the output result descriptor, I fall back to the srs area of use if the input descriptor has no bbox defined but does have a resolution defined
- in both methods, I reproject the output bbox back to the input srs now. My thinking was: we want the bbox in both projections and they should cover the same area of the world. If we only `reproject_clipped` then it might happen that the output bbox is much smaller than the input box. However, I noticed that the computed output resolution for my test drifted farther apart from what gdal does. So maybe we shouldn't do this after all? Maybe we have to TALK about it.

operators/src/processing/reprojection.rs
```diff
@@ -453,13 +461,25 @@ where
     ) -> Result<BoxStream<'a, Result<Self::Output>>> {
         // calculate the intersection of input and output srs in both coordinate systems
         // TODO: do this in initialization?
-        let valid_bounds_in = Self::valid_bounds::<SpatialPartition2D>(self.from, self.to)?;
-        let valid_bounds_out = Self::valid_bounds::<SpatialPartition2D>(self.to, self.from)?;
+        let (valid_bounds_in, valid_bounds_out) = if let Some(data_bounds) = self.data_bounds {
```
jdroenner, line 464: isn't this the same as in line 58?

operators/src/processing/reprojection.rs
jdroenner: yes this is also what i was thinking about. The deviation to GDAL is something that we need to talk about. GDAL also allows to process data outside of the "valid" bounds. Since we now have data bounds we could try to also allow that... However, this is something for another day.

operators/src/processing/reprojection.rs
michaelmattig: it's the same logic but the other way around. Here, we calculate the resolution for **querying** the source and in line 58 we calculate the maximum **output** resolution.

operators/src/processing/reprojection.rs
michaelmattig: As we discussed I changed the computation of the corresponding bboxes for the pixel size estimation. It now takes the intersection on both boxes in WGS84. It produces exactly the same, at least in the test case.

datatypes/src/spatial_reference.rs
```diff
@@ -119,6 +119,30 @@ impl SpatialReference {
     pub fn srs_string(&self) -> String {
         format!("{}:{}", self.authority, self.code)
     }
+
+    /// Compute the bounding box of this spatial reference that is also valid in the `other` spatial reference
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    pub fn valid_bounds<T>(&self, other: &SpatialReference) -> Result<T>
```
jdroenner, line 125: i think it is fine to return a WGS84 bbox since the area of use is also wgs84

datatypes/src/operations/reproject.rs
```diff
@@ -474,10 +476,45 @@ pub fn reproject_query<S: AxisAlignedRectangle>(
     })
 }
 
+/// Reproject a bounding box to the `target` projection and return the input and output bounding box
+/// as a pair where both elements cover the same area of the original `source_bbox` in WGS84.
+pub fn reproject_and_unify_bbox<T: AxisAlignedRectangle>(
```
jdroenner, line 481: let's keep the other variant. this is way to complex.

jdroenner (changes requested):

datatypes/src/spatial_reference.rs
michaelmattig: do you mean returning the intersection of the spatial references in wgs84? But we need it in the projection of `self` in the ReprojectionOperator. We could do it as an extra step but then I guess this method here doesn't serve much purpose.

datatypes/src/operations/reproject.rs
michaelmattig: 4cc7328bae331ccfa33462032ed881ca5f52c09f

jdroenner (dismissed):

jdroenner (approved):

jdroenner:
bors merge

PR #635, opened 2022-09-28
add docs for layer collection api

michaelmattig:

---

add docs for layer collection api

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           # An additional cache key that is added alongside the automatic `job`-based
           # cache key and can be used to further differentiate jobs.
           # Default: empty
-          key: 2_${{ matrix.features }}_${{ matrix.build }}
+          key: 3_${{ matrix.features }}_${{ matrix.build }}
 
           # A whitespace separated list of env-var *prefixes* who's value contributes
           # to the environment cache key.
@@ -89,11 +89,15 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-targets ${{ matrix.features }} ${{ matrix.build }} -- -D warnings
+        env:
+          CARGO_TARGET_DIR: /home/runner/work/geoengine/geoengine/target
       - name: Run tests
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: ${{ matrix.features }} ${{ matrix.build }} --verbose
+        env:
+          CARGO_TARGET_DIR: /home/runner/work/geoengine/geoengine/target
 
   build:
     runs-on: ${{matrix.os}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,11 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-targets ${{ matrix.features }} ${{ matrix.build }} -- -D warnings
         env:
+          # utoipa packages the UI part of the SwaggerUI and the build.rs of utoipa moves this into a "wrong" folder.
+          # The folder where it is moved is not covered by the cache.
+          # However, utoipa tries to access the UI files and this crashes the test/build process.
+          # Using the CARGO_TARGET_DIR we can force cargo to move all build dependecies into a specific folder.
+          # TODO: remove this when utoipa is fixed
           CARGO_TARGET_DIR: /home/runner/work/geoengine/geoengine/target
       - name: Run tests
         uses: actions-rs/cargo@v1
@@ -97,6 +102,7 @@ jobs:
           command: test
           args: ${{ matrix.features }} ${{ matrix.build }} --verbose
         env:
+          # TODO: remove this when utoipa is fixed
           CARGO_TARGET_DIR: /home/runner/work/geoengine/geoengine/target
 
   build:
@@ -161,4 +167,5 @@ jobs:
           command: build
           args: ${{ matrix.features }} ${{ matrix.build }} --verbose --target-dir=target
         env:
+          # TODO: remove this when utoipa is fixed
           CARGO_TARGET_DIR: /home/runner/work/geoengine/geoengine/target
--- a/datatypes/src/operations/image/colorizer.rs
+++ b/datatypes/src/operations/image/colorizer.rs
@@ -517,7 +517,7 @@ pub type Breakpoints = Vec<Breakpoint>;
 /// It is assumed that is has at least one and at most 256 entries.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(try_from = "SerializablePalette", into = "SerializablePalette")]
-pub struct Palette(HashMap<NotNan<f64>, RgbaColor>);
+pub struct Palette(pub HashMap<NotNan<f64>, RgbaColor>);
 
 /// A type that is solely for serde's serializability.
 /// You cannot serialize floats as JSON map keys.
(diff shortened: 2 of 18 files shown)


michaelmattig:
[API-Docs erzeugen](https://trello.com/c/g0zsro2u/738-api-docs-erzeugen)

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -177,19 +177,21 @@ impl SentinelS2L2aCogsDataProvider {
...
-                                    (10_000.0, RgbaColor::black())
-                                        .try_into()
-                                        .expect("valid breakpoint"),
-                                ],
-                                RgbaColor::transparent(),
-                                RgbaColor::transparent(),
-                            )
-                            .expect("valid colorizer"),
+                            colorizer:
+                                geoengine_datatypes::operations::image::Colorizer::linear_gradient(
```
ChristianBeilschmidt, line 181: Why do we need the full path here?

datatypes/src/operations/image/colorizer.rs
```diff
@@ -517,7 +517,7 @@ pub type Breakpoints = Vec<Breakpoint>;
 /// It is assumed that is has at least one and at most 256 entries.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(try_from = "SerializablePalette", into = "SerializablePalette")]
-pub struct Palette(HashMap<NotNan<f64>, RgbaColor>);
+pub struct Palette(pub HashMap<NotNan<f64>, RgbaColor>);
```
ChristianBeilschmidt, line 520: Why do we need to make this pub?

datatypes/src/operations/image/colorizer.rs
```diff
@@ -550,7 +550,7 @@ impl TryFrom<SerializablePalette> for Palette {
 
 /// `RgbaColor` defines a 32 bit RGB color with alpha value
 #[derive(Copy, Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
-pub struct RgbaColor([u8; 4]);
+pub struct RgbaColor(pub [u8; 4]);
```
ChristianBeilschmidt, line 553: Why do we need to make this pub?

.github/workflows/ci.yml
```diff
@@ -89,11 +89,15 @@ jobs:
...
           args: --all-targets ${{ matrix.features }} ${{ matrix.build }} -- -D warnings
+        env:
+          CARGO_TARGET_DIR: /home/runner/work/geoengine/geoengine/target
       - name: Run tests
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: ${{ matrix.features }} ${{ matrix.build }} --verbose
+        env:
+          CARGO_TARGET_DIR: /home/runner/work/geoengine/geoengine/target
```
ChristianBeilschmidt, line 100: is this necessary or was it just the cache?

services/src/handlers/layers.rs
```diff
@@ -171,6 +266,168 @@ async fn list_collection_handler<C: Context>(
...
+        )
+    ),
+    params(
+        ("provider" = DataProviderId, description = "Data provider id"),
+        ("layer" = LayerCollectionId, description = "Layer id"),
+    ),
+    security(
+        ("session_token" = [])
+    )
+)]
```
ChristianBeilschmidt, line 430: Maybe you can make sense of this error

grafik

ChristianBeilschmidt (changes requested):
Maybe take a look at the checkbox

.github/workflows/ci.yml
jdroenner: The problem is that utoipa packages the UI part of the SwaggerUI and the build.rs of utoipa moves this into a "wrong" folder. The folder where it is moved is not covered by the cache. However, utoipa tries to access the UI files and this crashes the test/build process.
Using the CARGO_TARGET_DIR we can force cargo to move all build dependecies into a specific folder. And the build.rs of utoipa also uses this env variable to determine the folder whrere the ui stuff is moved into.

datatypes/src/operations/image/colorizer.rs
michaelmattig: the Palette is part of our API but defined in the datatypes crate. Thus I copied it to the services crate and implemented a From-Conversion. But that requires access to the map.

datatypes/src/operations/image/colorizer.rs
michaelmattig: the RgbaColor is part of our API but defined in the datatypes crate. Thus I copied it to the services crate and implemented a From-Conversion. But that requires access to the map.

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
michaelmattig: The `linear_gradient` method is only available on the Colorizer of the datatypes crate. But for including it into the API I have to use the services crate counterpart. To make it apparent that this Colorizer is not the one used in the API, but only converted `into` the actual Colorizer, I put the full path here.

services/src/handlers/layers.rs
michaelmattig: you switched data and layer provider id. it has to be the other way around.

datatypes/src/operations/image/colorizer.rs
ChristianBeilschmidt: cf. above

.github/workflows/ci.yml
ChristianBeilschmidt: Can you link the utoipa issue as a TODO?

datatypes/src/operations/image/colorizer.rs
michaelmattig: if you have an idea, please let me know. I couldn't think of one.

ChristianBeilschmidt: At least create a special constructor for that. Then, you don't access the fields differently now.

.github/workflows/ci.yml
ChristianBeilschmidt: Could you create one and link it?

michaelmattig: unfortunately I can't, because I don't understand the problem well enough. Johannes would have to do that.

datatypes/src/operations/image/colorizer.rs
ChristianBeilschmidt: `pub` means you can access the internal HashMap, etc. directly.
That wasn't the case before.

.github/workflows/ci.yml
ChristianBeilschmidt: Since he already commented, let's include him again?

datatypes/src/operations/image/colorizer.rs
michaelmattig: so do you want an *accessor*  or a constructor?

ChristianBeilschmidt: You can remove your pub declaration of the internals of the struct. This means, you can only work with the struct like before and not access the internals directly.

datatypes/src/operations/image/colorizer.rs
michaelmattig: 4cc60a0b64b1d9fb4b45e664ff15b6cc92a8de3e

.github/workflows/ci.yml
michaelmattig: 4e2e5fddb02b1e8a80060d89894e7261adc621cf

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #639, opened 2022-09-30
Feature/xgboost

MarWeUMR:
- [ x] I added an entry to [`CHANGELOG.md`](CHANGELOG.md) if knowledge of this change could be valuable to users.

---

This PR is intended to add functionality for using `XGBoost` with a pre-trained model to generate predictions based on some raster tiles.

There are three main parts:

- the operator itself: xgboost.rs
- the xgboost-sys crate to generate the bindings and build the library
- the bindings module where all the higher level functionality is added to the bindings

In the operator is a test method located, which uses the `marburg`dataset to predict land usage. This test takes some time to finish and should be replaced with an easier dataset just for testing purpose.


MarWeUMR:
[XGBoost Operator implementieren](https://trello.com/c/QoxXOMwb/1048-xgboost-operator-implementieren)

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+        }
+    }
+
+    async fn process_tile_async(
+        &self,
+        bands_of_tile: Vec<Result<Tile<P>, crate::error::Error>>,
+        pool: Arc<ThreadPool>,
+    ) -> Result<RasterTile2D<PixelOut>> {
+        // gather information
+        let reference_tile = bands_of_tile.first().unwrap();
```
michaelmattig, line 137: perform proper error handling (error variant + ?-operator)

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+        let n_bands = bands_of_tile.len() as i32;
+        let time = tile.time;
+        let tile_position = tile.tile_position;
+        let global_geo_transform = tile.global_geo_transform;
+
+        // extract the actual data from the tiles
+        let rasters: Vec<Vec<f32>> = bands_of_tile
+            .into_iter()
+            .map(|band| {
+                let band_ok = band.unwrap();
```
michaelmattig, line 152: error handling

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+                let mat_tile = band_ok.into_materialized_tile();
+
+                mat_tile
+                    .grid_array
+                    .inner_grid
+                    .data
+                    .into_iter()
+                    .map(num_traits::AsPrimitive::as_)
+                    .collect::<Vec<f32>>()
```
michaelmattig, line 161: you already converted all the source rasters into f64, why do you convert again? Can't you convert them to f32 directly? Then you could use the `data` here as it is

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+                    .unwrap()
+                    .to_owned();
+                row_data.push(pxl);
+            }
+            pixels.extend_from_slice(&row_data);
+        }
+
+        let model = self.model_file.clone();
+        let predicted_grid: geoengine_datatypes::raster::Grid<GridShape<[usize; 2]>, f32> =
+            crate::util::spawn_blocking(move || {
```
michaelmattig, line 183: I think you should do all the stuff above in spawn_blocking. You shouldn't do too much work in `async` functions

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+            })
+            .collect();
+
+        let mut pixels: Vec<f32> = Vec::new();
+
+        for row in 0..(n_rows * n_cols) {
+            let mut row_data: Vec<f32> = Vec::new();
+            for col in 0..n_bands {
+                let pxl = rasters
+                    .get(col as usize)
```
michaelmattig, line 171: you can directly access the raster wirh `[i]` because you know the values exist

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+        };
+
+        let op = RasterOperator::boxed(xg).initialize(&ctx).await.unwrap();
+
+        let processor = op.query_processor().unwrap().get_f32().unwrap();
+
+        let ctx = MockQueryContext::test_default();
+        let result_stream = processor.query(qry_rectangle, &ctx).await.unwrap();
+
+        // TODO: .take(1) for a quick test run
```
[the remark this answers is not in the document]
MarWeUMR: This was a badly worded reminder, that the .take(1) call needs to be removed once a proper test dataset is used.
Changed in [1834aac](https://github.com/geo-engine/geoengine/pull/639/commits/1834aacd06d45aa78f7fcbe3069c18b8562f3381).

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+        let result: Vec<crate::util::Result<RasterTile2D<f32>>> =
+            result_stream.take(1).collect().await;
+
+        let mut all_pixels = Vec::new();
+
+        // TODO: .take(1) for a quick test run
```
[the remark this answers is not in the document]
MarWeUMR: See my comment above.

operators/Cargo.toml
```diff
@@ -48,6 +48,8 @@ tokio = { version = "1.15", features = ["macros", "signal", "sync", "rt-multi-th
 tracing = "0.1"
 typetag = "0.2"
 uuid = { version = "1.1", features = ["serde", "v4", "v5"] }
+# xgboost deps
```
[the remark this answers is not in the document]
MarWeUMR: Changed with [cbdd559](https://github.com/geo-engine/geoengine/pull/639/commits/cbdd55987bcb946624d440cfa7b362f1f2ac65af).

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added a new operator `XGBoost`. This operator allows to use a pretrained model in order to make predictions based on some set of raster tile data.
```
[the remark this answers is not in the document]
MarWeUMR: Changed with [cbdd559](https://github.com/geo-engine/geoengine/pull/639/commits/cbdd55987bcb946624d440cfa7b362f1f2ac65af).

operators/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub mod source;
 pub mod util;
 
 /// Compiles Geo Engine Pro
-#[cfg(feature = "pro")]
+// #[cfg(feature = "pro")]
```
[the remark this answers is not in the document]
MarWeUMR: Chaged with [cbdd559](https://github.com/geo-engine/geoengine/pull/639/commits/cbdd55987bcb946624d440cfa7b362f1f2ac65af).

operators/src/pro/mod.rs
```diff
@@ -1 +1,3 @@
 // This is an inclusion point of Geo Engine Pro
+// #[cfg(feature="pro")]
```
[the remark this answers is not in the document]
MarWeUMR: Changed with [cbdd559](https://github.com/geo-engine/geoengine/pull/639/commits/cbdd55987bcb946624d440cfa7b362f1f2ac65af).

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+
+                let mut out_dim: u64 = 0;
+                let bst = Booster::load_buffer(model_file).unwrap();
+                // measure time for prediction
+                let result = bst.predict_from_dmat(
+                    &xg_matrix,
+                    &[n_parallel_pixels as u64, n_bands as u64],
+                    &mut out_dim,
+                );
+                result.unwrap()
```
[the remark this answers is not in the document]
MarWeUMR: Changed in [ff0886b](https://github.com/geo-engine/geoengine/pull/639/commits/ff0886b6292e3fe8cf3e4addb4f5579c05cbecd4).

operators/src/pro/ml/xgboost.rs
MarWeUMR: Changed in [ff0886b](https://github.com/geo-engine/geoengine/pull/639/commits/ff0886b6292e3fe8cf3e4addb4f5579c05cbecd4).

operators/src/pro/ml/xgboost.rs
MarWeUMR: Changed in [ff0886b](https://github.com/geo-engine/geoengine/pull/639/commits/ff0886b6292e3fe8cf3e4addb4f5579c05cbecd4).

operators/Cargo.toml
```diff
@@ -48,6 +48,8 @@ tokio = { version = "1.15", features = ["macros", "signal", "sync", "rt-multi-th
...
+xgboost = {git = "https://github.com/MarWeUMR/xgboost-rs"}
```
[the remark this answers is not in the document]
MarWeUMR: Done [license](https://github.com/MarWeUMR/xgboost-rs/blob/main/LICENSE)

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+            .par_chunks(chunk_size)
+            .map(|elem| {
+                // get xgboost style matrices
+                let xg_matrix = DMatrix::from_col_major_f32(
+                    elem,
+                    mem::size_of::<f32>() * n_bands,
+                    mem::size_of::<f32>(),
+                    n_parallel_pixels,
+                    n_bands,
+                    -1,
```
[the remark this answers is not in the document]
MarWeUMR: This is the number of threads parameter for xgboost. It's optional and could be removed if there shouldn't be an option to restrict the number of threads from the geoengine.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+use TypedRasterQueryProcessor::F32 as QueryProcessorOut;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct XgboostParams {
+    model_file_path: String,
+}
+
+pub type XgboostOperator = Operator<XgboostParams, MultipleRasterSources>;
+pub type Tile<P> = BaseTile<GridOrEmpty<GridShape<[usize; 2]>, P>>;
```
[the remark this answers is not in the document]
MarWeUMR: Updated in https://github.com/geo-engine/geoengine/pull/639/commits/6293d0f17193beb5b44941d4483d73200499f4f8

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+        .await?;
+
+        let input = init_rasters.get(0).unwrap();
+        let in_desc = input.result_descriptor();
+
+        let out_desc = RasterResultDescriptor {
+            data_type: RasterOut,
+            time: in_desc.time,
+            bbox: in_desc.bbox,
+            resolution: in_desc.resolution,
```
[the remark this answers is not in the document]
MarWeUMR: I added a routine for that in https://github.com/geo-engine/geoengine/pull/639/commits/6293d0f17193beb5b44941d4483d73200499f4f8#diff-ee4ef753360f7a08623dea3d2f3283835ae7f5d06414f4105fab5c25184f3f6cR148. Not sure if the error variant was intended for cases like that, though.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+        )
+        .await?;
+
+        let input = init_rasters.get(0).unwrap();
+        let in_desc = input.result_descriptor();
+
+        let out_desc = RasterResultDescriptor {
+            data_type: RasterOut,
+            time: in_desc.time,
+            bbox: in_desc.bbox,
```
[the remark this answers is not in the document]
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/5927879ce4601e18e31e87a860066f5d6794581d

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+                .map(|raster| raster.clone().initialize(context)),
+        )
+        .await?;
+
+        let input = init_rasters.get(0).unwrap();
+        let in_desc = input.result_descriptor();
+
+        let out_desc = RasterResultDescriptor {
+            data_type: RasterOut,
+            time: in_desc.time,
```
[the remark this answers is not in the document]
MarWeUMR: Same implementation as above.  
https://github.com/geo-engine/geoengine/pull/639/commits/5927879ce4601e18e31e87a860066f5d6794581d

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+}
+
+impl InitializedRasterOperator for InitializedXgboostOperator {
+    fn result_descriptor(&self) -> &RasterResultDescriptor {
+        &self.result_descriptor
+    }
+
+    fn query_processor(&self) -> Result<TypedRasterQueryProcessor> {
+        // given a set of raster bands, if they have different types,
+        // where should they be converted?
```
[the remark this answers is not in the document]
MarWeUMR: I'm sorry, that comment was old. Please ignore.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+
+impl<Q, P> XgboostProcessor<Q, P>
+where
+    Q: RasterQueryProcessor<RasterType = P>,
+    P: Pixel,
+{
+    pub fn new(sources: Vec<Q>, model_file_path: String) -> Self {
+        Self {
+            sources,
+            model_file: std::fs::read_to_string(model_file_path).unwrap(),
```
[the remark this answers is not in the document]
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/5927879ce4601e18e31e87a860066f5d6794581d

operators/src/pro/ml/xgboost.rs
MarWeUMR: I added a new method for the query_processor in order to achieve that. https://github.com/geo-engine/geoengine/pull/639/commits/5927879ce4601e18e31e87a860066f5d6794581d
I hope it's ok like that, or would there be another way?

operators/src/pro/ml/xgboost.rs
MarWeUMR: Restructured in https://github.com/geo-engine/geoengine/pull/639/commits/5927879ce4601e18e31e87a860066f5d6794581d

operators/src/pro/ml/xgboost.rs
MarWeUMR: See comment below

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+                    .data
+                    .into_iter()
+                    .map(num_traits::AsPrimitive::as_)
+                    .collect::<Vec<f32>>()
+            })
+            .collect();
+
+        let mut pixels: Vec<f32> = Vec::new();
+
+        for row in 0..(n_rows * n_cols) {
```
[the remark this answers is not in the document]
MarWeUMR: Rewrote the logic in https://github.com/geo-engine/geoengine/pull/639/commits/be1683071ae1ef41bb8ca48470545cbb65ffca2c

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+fn process_tile(
+    bands_of_tile: &Vec<f32>,
+    pool: &ThreadPool,
+    model_file: &[u8],
+    grid_shape: GridShape<[usize; 2]>,
+    n_bands: usize,
+) -> geoengine_datatypes::raster::Grid<GridShape<[usize; 2]>, f32> {
+    pool.install(|| {
+        // to get one row of data means taking (n_pixels * n_bands) elements
+        let n_parallel_pixels = grid_shape.shape_array[0] * grid_shape.shape_array[1];
```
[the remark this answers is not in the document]
MarWeUMR: Applied in https://github.com/geo-engine/geoengine/pull/639/commits/be1683071ae1ef41bb8ca48470545cbb65ffca2c

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+            .map(|elem| {
+                // get xgboost style matrices
+                let xg_matrix = DMatrix::from_col_major_f32(
+                    elem,
+                    mem::size_of::<f32>() * n_bands,
+                    mem::size_of::<f32>(),
+                    n_parallel_pixels,
+                    n_bands,
+                    -1,
+                    0.0,
```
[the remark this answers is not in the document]
MarWeUMR: Yes, correct. This is the nan/no data value that should be used in xgboost. I've added a parameter for that now, but I'm not sure where it should be set by the user.
Furthermore, what would be an appropriate default value here (if any)?

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+        let in_desc = input.result_descriptor();
+
+        let out_desc = RasterResultDescriptor {
+            data_type: RasterOut,
+            time: in_desc.time,
+            bbox: in_desc.bbox,
+            resolution: in_desc.resolution,
+            spatial_reference: in_desc.spatial_reference,
+
+            // where do we get this information from?
```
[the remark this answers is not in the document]
MarWeUMR: Is this the part you are referring to? https://github.com/geo-engine/geoengine/blob/8bf2d34936f0d35506271621e8b9e0d0c23bdb4b/operators/src/processing/expression/mod.rs#L281-L292

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+
+type PixelOut = f32;
+
+#[typetag::serde]
+#[async_trait]
+impl RasterOperator for XgboostOperator {
+    async fn initialize(
+        self: Box<Self>,
+        context: &dyn ExecutionContext,
+    ) -> Result<Box<dyn InitializedRasterOperator>> {
```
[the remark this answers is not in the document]
MarWeUMR: Do you mean like done in the expression operator here?
https://github.com/geo-engine/geoengine/blob/8bf2d34936f0d35506271621e8b9e0d0c23bdb4b/operators/src/processing/expression/mod.rs#L256-L265

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+                },
+            };
+            src_vec.push(mrs.boxed());
+        }
+        src_vec
+    }
+
+    #[tokio::test]
+    async fn xg_op_test() {
+        // TODO: This test needs a much smaller test data set.
```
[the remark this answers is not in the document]
MarWeUMR: Is there an appropriate dataset available, or should we come up with a simple one that would be good to test if the xgboost integration is actually behaving as expected within the geoengine workflow?

operators/Cargo.toml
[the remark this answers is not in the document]
michaelmattig: maybe create a tag, or specify a commit here so that we have a fixed version of your library

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: yes. I think all the input rasters must have the same projection, otherwise the pixel you put into the ML algorithm don't match, right?

operators/Cargo.toml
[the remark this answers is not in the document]
michaelmattig: maybe create a tag, or specify a commit here so that we have a fixed version of your library

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: now that I think about it: inheriting won't make sense I guess. We could take it as a parameter like in the snippet you posted, but then again, I guess the measurement is always the same for a specific learned model, right? So maybe we should store it in (or alongside) the model.json. This would mean users will have to specify the measurement during learning or at least during import of the model. What do you think?

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+    ) -> Result<Box<dyn InitializedRasterOperator>> {
+        let init_rasters = future::try_join_all(
+            self.sources
+                .rasters
+                .iter()
+                .map(|raster| raster.clone().initialize(context)),
+        )
+        .await?;
+
+        let input = init_rasters.get(0).unwrap();
```
michaelmattig, line 73: you should throw an error like, XgBoost requires at least 1 raster input

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+        let in_desc = input.result_descriptor();
+
+        // TODO: provide a sane default for the none case
+        // ASSUMPTION: the shape is always square?
+        let mut fin_res = SpatialResolution::new(512., 512.).unwrap();
```
michaelmattig, line 78: please have a look at this: https://github.com/geo-engine/geoengine/blob/d2a12abe9916f31a1e47daf069522bf3408c2858/operators/src/processing/expression/mod.rs#L267 you can reuse this.

if one of the input result descriptors has `None` for bbox, time or resolution, the output descriptor also has to be `None`.

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: cf. my other comment above

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+        let n_bands = bands_of_tile.len() as i32 + 1; // +1 because of .pop()
+        let props = tile.properties.clone(); // = &tile.properties;
+        let time = tile.time;
+        let tile_position = tile.tile_position;
+        let global_geo_transform = tile.global_geo_transform;
+
+        let model = self.model_file.clone();
+        let ndv = self.no_data_value.or(Some(f32::NAN));
+
+        let predicted_grid: geoengine_datatypes::raster::Grid<GridShape<[usize; 2]>, f32> =
```
michaelmattig, line 204: you can also use the `Grid2d` type here

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+                            .grid_array
+                            .inner_grid
+                            .data
+                            .into_iter()
+                            .map(num_traits::AsPrimitive::as_)
+                            .collect::<Vec<f32>>()
+                    })
+                    .collect();
+
+                // let pixels: Vec<_> = rasters.into_iter().collect();
```
michaelmattig, line 231: remove?

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+    model_file_path: String,
+}
+
+pub type XgboostOperator = Operator<XgboostParams, MultipleRasterSources>;
+pub type Tile<P> = BaseTile<GridOrEmpty<GridShape<[usize; 2]>, P>>;
+
+pub struct InitializedXgboostOperator {
+    result_descriptor: RasterResultDescriptor,
+    sources: Vec<Box<dyn InitializedRasterOperator>>,
+    model_file_path: String,
```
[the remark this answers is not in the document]
michaelmattig: I guess (in the future) we need a proper `MLModelDb` (or something like this) where we manage the learned models. When applying the operator we would only specify the ID of the learned model and the operator can get the model parameters and metadata like the output unit from this database.

In the meantime, I think we should e.g. have a configurable base directory where we place all the model jsons. The operator then doesn't get the full path to the file, but only the path *relative* to this base directory. Then the operator must only check that the input path (that comes from the user) doesn't escape this path. We already have some logic for that. I think `canonicalize_subpath` is that function.

This is all for ensuring that the user can't just read any file on disk, but only the ones in the folder with the models.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+                // put the tile back into place
+                bands_of_tile.insert(0, Ok(tile));
+                // extract the actual data from the tiles
+                let pixels: Vec<_> = bands_of_tile
+                    .into_iter()
+                    .flat_map(|band| {
+                        let band_ok = match band {
+                            Ok(x) => x,
+                            Err(err) => {
+                                panic!("Failed to get BaseTile with error msg: \n{:?}", err)
```
michaelmattig, line 215: we don't want to crash the whole program because of an error tile. I guess if one tile is Error, the whole `process_tile_async` should just return Error?

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+                            .inner_grid
+                            .data
+                            .into_iter()
+                            .map(num_traits::AsPrimitive::as_)
+                            .collect::<Vec<f32>>()
+                    })
+                    .collect();
+
+                // let pixels: Vec<_> = rasters.into_iter().collect();
+                let model_content = std::fs::read_to_string(model).unwrap();
```
michaelmattig, line 232: you should never use unwrap, but always handle the error. Or: if you are sure that it can never occure, use `expect` and explain why it can never fail.

Here it CAN fail and it should result in an Error and not a crash. How large is the model? Depending on the size I would suggest to load the model during `initialize` of the Operator and just keep it in memory. We don't want to read it for every tile. 

If it can be very large then maybe only load it on `query` because if we load it during `initialize` we may have loaded in unnecessarily when the processor is never actually queried.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+
+                process_tile(
+                    &pixels,
+                    &pool,
+                    model_content.as_bytes(),
+                    grid_shape,
+                    n_bands as usize,
+                    ndv.unwrap(),
```
michaelmattig, line 240: also here: don't unwrap but return an Error. Maybe you can also check this during initialization and unpack the option there so that you know here that there is a value.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+                )
+            })
+            .await
+            .unwrap();
```
michaelmattig, line 244: error handling

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+impl InitializedRasterOperator for InitializedXgboostOperator {
+    fn result_descriptor(&self) -> &RasterResultDescriptor {
+        &self.result_descriptor
+    }
+
+    fn query_processor(&self) -> Result<TypedRasterQueryProcessor> {
+        let vec_of_rqps: Vec<Box<dyn RasterQueryProcessor<RasterType = f32>>> = self
+            .sources
+            .iter()
+            .map(|init_raster| init_raster.query_processor().unwrap().into_f32())
```
michaelmattig, line 153: error handling

michaelmattig, line 153: error handling

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: cf my other comment above

operators/src/pro/ml/xgboost.rs
michaelmattig: ok but why do you you still convert the data here? They are already f32

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: add a TODO to maybe make this configurable in the Settings.toml

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: When we still NoData values in our rasters we could have used them. But we have masked rasters now. Without looking at the data we don't know which value is safe to use. So I think the user has to specify it.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,560 @@
...
+                    mem::size_of::<f32>(),
+                    n_parallel_pixels,
+                    n_bands,
+                    -1,
+                    0.0,
+                )
+                .unwrap();
+
+                let mut out_dim: u64 = 0;
+                let bst = Booster::load_buffer(model_file).unwrap();
```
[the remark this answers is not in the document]
michaelmattig: please don't panic but throw a new Error like "Could not load Booster buffer" and let the query fail instead of crashing the program.

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: I don't know one. Maybe we could just use some random rasters as input, and calculate the truth based on some simple formula, which the model then learns to predict?

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+                    elem,
+                    mem::size_of::<f32>() * n_bands,
+                    mem::size_of::<f32>(),
+                    grid_shape.number_of_elements(),
+                    n_bands,
+                    -1,
+                    nan_val,
+                ) {
+                    Ok(matrix) => matrix,
+                    Err(err) => panic!(
```
michaelmattig, line 285: return error instead of panic

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+                        "XGBoost failed to create DMatrix with error msg: \n{:?}",
+                        err
+                    ),
+                };
+
+                let mut out_dim: u64 = 0;
+                let bst = match Booster::load_buffer(model_file) {
+                    Ok(booster) => booster,
+                    Err(err) => panic!(
```
michaelmattig, line 294: Ok(matrix) => matrix,
                    Err(err) => panic!(

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+
+                // measure time for prediction
+                match bst.predict_from_dmat(
+                    &xg_matrix,
+                    &[grid_shape.number_of_elements() as u64, n_bands as u64],
+                    &mut out_dim,
+                ) {
+                    Ok(res) => res,
+                    Err(err) => {
+                        panic!("XGBoost failed to predict data with error msg: \n{:?}", err)
```
michaelmattig, line 308: return error instead of panic

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
MarWeUMR: I think it would be good to require the class labels, because as of right now I had to use a routine to get the class labels and create a mapping for them which is passed to xgboost in the training process (the corresponding code is not part of the PR). Having a config for that would make things easier.

operators/src/pro/ml/xgboost.rs
MarWeUMR: I might be missing something obvious here, but `data` is a `Vec<P>` and I need that conversion to get stuff over to xgboost, which doesn't know the `Pixel` trait. What other way could I use here?

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,628 @@
...
+
+    async fn query<'a>(
+        &'a self,
+        query: RasterQueryRectangle,
+        ctx: &'a dyn QueryContext,
+    ) -> Result<BoxStream<'a, Result<Self::Output>>> {
+        let mut band_buffer = Vec::new();
+
+        for band in &self.sources {
+            let stream = band.query(query, ctx).await.unwrap();
```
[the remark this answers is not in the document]
MarWeUMR: https://github.com/geo-engine/geoengine/commit/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/4c95927f876b2b966e402d64135a45e44dd92a9f

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
MarWeUMR: https://github.com/geo-engine/geoengine/commit/84fb951c583cb7141ff50f1196f90504d2ef8a90

operators/Cargo.toml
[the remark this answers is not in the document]
MarWeUMR: Added with https://github.com/geo-engine/geoengine/pull/639/commits/70e5ac4c6e5181ab3f180f532aee6528a1521b69.
Reference: https://github.com/MarWeUMR/xgboost-rs/tags

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
MarWeUMR: Added the changes with https://github.com/geo-engine/geoengine/pull/639/commits/4c95927f876b2b966e402d64135a45e44dd92a9f

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/5a44b4fef9cb96d82488102cfbcf4e38d95e649a

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
MarWeUMR: Ok, so I made it a hard requirement. https://github.com/geo-engine/geoengine/pull/639/commits/503b946b55d52e2613f672271a28f1bb79ca5f17

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
MarWeUMR: yes, that would be a good option. I will have a look at some of the other operators and see if I can get a few ideas how to achieve that.

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/5815f849de667bc249ec87211fc8ecf941a99970

operators/src/pro/ml/xgboost.rs
MarWeUMR: sorry about that.
https://github.com/geo-engine/geoengine/pull/639/commits/57b4aed25ecc0f872fa197e2f62b0ae1546a0b75

operators/src/pro/ml/xgboost.rs
MarWeUMR: I removed the option from that with https://github.com/geo-engine/geoengine/pull/639/commits/503b946b55d52e2613f672271a28f1bb79ca5f17

operators/src/pro/ml/xgboost.rs
MarWeUMR: I changed the error handling with https://github.com/geo-engine/geoengine/commit/84fb951c583cb7141ff50f1196f90504d2ef8a90.

There is an ugly problem with the model file. I need to load the model for each tile, because the booster instance has to be created distinctly and can not be used across the multiple tiles. I encountered this issue a while ago and am not entirely sure anymore, but I think it had to do with the unsafe part of the booster code. 

I will try to recreate the situation, maybe we can sort it out then.

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: better move the `canonicalize_subpath` code to the datatypes crate and use it here

operators/src/pro/ml/xgboost.rs
michaelmattig: ok but you can still read the file once and only recreate the booster instance for each tile, right?

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,694 @@
...
+
+    fn query_processor(&self) -> Result<TypedRasterQueryProcessor> {
+        let vec_of_rqps: Vec<Box<dyn RasterQueryProcessor<RasterType = f32>>> = self
+            .sources
+            .iter()
+            .map(|init_raster| {
+                let typed_raster_qp = match init_raster.query_processor() {
+                    Ok(r) => r,
+                    Err(e) => {
+                        return Err(Error::XGBoost {
```
michaelmattig, line 216: why is an XGBoost error if the input raster processor could not be created? It has nothing to do with your operaror.

Also: you can use the `map_err` function instead of a match block if you want to replace the error of a result.

operators/src/pro/ml/xgboost.rs
michaelmattig: It has nothing to do with your operator so use the source error directly.

operators/src/pro/ml/xgboost.rs
michaelmattig, line 216: I get 

>  ^^^^^ use of undeclared type `Error

here, I guess you have to import or qualify ther Error

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: the class labels are always the same for each application of the learner right? So I think they should be part of the model.json and not be supplied on each use of the operator.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,694 @@
...
+
+        for band in &self.sources {
+            let stream = band.query(query, ctx).await?;
+            band_buffer.push(stream);
+        }
+
+        let svz = StreamVectorZip::new(band_buffer);
+
+        let svz_vec: Vec<_> = svz.collect().await;
+        let stream_of_tiles_with_zipped_bands = stream::iter(svz_vec).map(Ok).boxed();
```
michaelmattig, line 407: why do you collect the stream into a vec and then create another stream out of it? Why don't you use svz directly?

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,694 @@
...
+            no_data_value,
+        }
+    }
+
+    async fn process_tile_async(
+        &self,
+        bands_of_tile: Result<Vec<RasterTile2D<P>>>,
+        pool: Arc<ThreadPool>,
+    ) -> Result<BaseTile<GridOrEmpty<GridShape<[usize; 2]>, f32>>> {
+        let bot = match bands_of_tile {
```
michaelmattig, line 263: - maybe you should do the error handling before calling this method. Then you can pass a `Vec<RasterTile2D<P>>` instead of a Result
- why is the error called `NoInputData`? Shouldn't we propagate the error that occured in the source that produced the error tile?
- collecting the input tiles into a vector should better also be done in the spawned task and not in the async code

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,694 @@
...
+                        .into_iter()
+                        .map(num_traits::AsPrimitive::as_)
+                        .collect::<Vec<f32>>();
+
+                    Ok(res)
+                })
+                .flatten()
+                .collect();
+
+            let model_content =
```
michaelmattig, line 303: try to read the data once in the `query` method once and reuse it for the tiles.

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,694 @@
...
+                    Booster::load_buffer(model_file).context(self::error::LoadBoosterFromFile)?;
+
+                // measure time for prediction
+                let predictions = bst.predict_from_dmat(
+                    &xg_matrix,
+                    &[grid_shape.number_of_elements() as u64, n_bands as u64],
+                    &mut out_dim,
+                );
+
+                match predictions {
```
michaelmattig, line 369: use `map_err` instead of the match block

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,694 @@
...
+                    n_bands,
+                    -1, // TODO: add this to settings.toml: # of threads for xgboost to use
+                    nan_val,
+                )
+                .context(self::error::CreateDMatrix)?;
+
+                let mut out_dim: u64 = 0;
+
+                let bst =
+                    Booster::load_buffer(model_file).context(self::error::LoadBoosterFromFile)?;
```
michaelmattig, line 360: isn't the error name misleading? You don't load from a file here. It's not that you couldn't load the data from the file, but that it doesn't contain a valid model or something like that.

operators/src/pro/ml/xgboost.rs
MarWeUMR: Ok, i moved the reading of the file to `_query()` in https://github.com/geo-engine/geoengine/pull/639/commits/a7a59da740eb95fab06e72c3a1a0fa5de0d0cf2c.

I expect, that the cloning of `model_content` is not ok, or is it? I did not find a way without it, that would make the code work, since sooner or later every tile consumes a copy of the model when the booster is created?

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/4d0ada181995c7c47fdc471a097241750233ffc1

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
MarWeUMR: The `model.json` is generated by XGBoost itself and does not contain information about the class labels. XGBoost always uses integer labels in form of `[0, n_classes)`. I could add the information manually to the model file as part of the routine, but that would still require a definition of the labels at some point?

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/72aa6c7ba5661768aee327f268897e18182aa408

Is there a general rule for when to use `match` and when `map_err`?

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/d38301c4f02f600b2465b594ee660a95e8956386

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/0c5afdf29e30de1f9614430b19434949b8434f3e

I struggled a lot with finding a way to move the error handling out of the function and into the future closure. I'm sure there is a better way for this?

operators/src/pro/ml/xgboost.rs
MarWeUMR: Good point.
https://github.com/geo-engine/geoengine/pull/639/commits/db3caa18b70d8dafeafe566d4b301fa9cef14786

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/92a746816eae8d0a1792b39115b1f1259118a406

operators/src/pro/ml/xgboost.rs
MarWeUMR: Should I try to be explicit in this case and use the (hopefully) correct variant, or should I go with `Into::into` where possible?
https://github.com/geo-engine/geoengine/pull/639/commits/92a746816eae8d0a1792b39115b1f1259118a406

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
michaelmattig: we could
1. hijack the model.json and add some more data, hoping that XGBoost ignores it
2. store a second json next to it with the additional info. Maybe one folder for each learned model with on model.json and metadata.json

operators/src/pro/ml/xgboost.rs
michaelmattig: can you maybe use a `Rc` (or `Arc`) instead of `clone`?

operators/src/pro/ml/xgboost.rs
michaelmattig: if you only convert error types use `map_err` and `.context`. if you do more than that use `match`

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,664 @@
...
+                predicted_grid.into(),
+                props.clone(),
+            );
+
+        Ok(rt)
+    }
+}
+
+fn process_tile(
+    // bands_of_tile: &Vec<f32>,
```
michaelmattig, line 285: remove

operators/src/pro/ml/xgboost.rs
michaelmattig: ideally, every error that occurs in your operator should be a `crate::error::Error::XGBoost`. All other errors should, at the top-level, be wrapped by that error. You could, e.g. have your methods return a `Result<(), XGBoostModuleError>` and only in the `query` method, perform the wrapping with `crate::error::Error::XGBoost`

operators/src/pro/ml/xgboost.rs
MarWeUMR: `Rc` doesn't work, because `Rc<String>` is not implementing `Send`. With `Arc` it seems to work though.

operators/src/pro/ml/xgboost.rs
MarWeUMR: Ok so I guess this part of the code can be summarized as 'learning curve'. I hope the changes I made now are making things a little bit more concise https://github.com/geo-engine/geoengine/pull/639/commits/1e11b8ab7d01829e7190ffd2475331f073d8e13c. 
I was taking a bit of inspiration by this part of the time shift operator here:
https://github.com/geo-engine/geoengine/blob/826367d205677a7d1e30cba493910bb10b0304c6/operators/src/processing/time_shift.rs#L412-L431

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/c7b4ea4f0d9ef95371fad4c8425c86320cf925e6

operators/src/pro/ml/xgboost.rs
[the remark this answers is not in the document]
MarWeUMR: My personal favourite in this case would be the second option. I'll try to come up with a solution.

operators/src/pro/ml/xgboost.rs
MarWeUMR: I think with the changes made in https://github.com/geo-engine/geoengine/commit/c7b4ea4f0d9ef95371fad4c8425c86320cf925e6 it should be the way you described it.

operators/src/pro/ml/xgboost.rs
michaelmattig: do you mean `Ok(processed_tile.await?)`? That's fine, the closure within `.then` will only be called once a new item is ready. So the `await` happens in the resulting stream and not while creating the stream.

Settings-default.toml
```diff
@@ -76,6 +76,7 @@ dataset_defs_path = "./test_data/dataset_defs"
 provider_defs_path = "./test_data/provider_defs"
 layer_defs_path = "./test_data/layer_defs"
 layer_collection_defs_path = "./test_data/layer_collection_defs"
+ml_model_defs_path = "./test_data/ml_model_defs"
```
michaelmattig, line 79: Rather put it under `pro.ml.xgboost.model_path`

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,1105 @@
...
+        .unwrap();
+
+        xg_matrix
+    }
+
+    /// This test is used to:
+    /// a) train a xgboost model
+    /// b) predict unseen data
+    /// c) check, if the predictions are correct
+    /// This test only checks, if the xgboost library works correctly
```
michaelmattig, line 521: idk if we want to test the correctness of libraries we use. Maybe move this test to your library instead?

MarWeUMR: I removed the test. My reasoning was to use this test in order to verify if the library works in the context of tile data.

Settings-default.toml
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/2e54047e11dfc53922b2a0b5103287109b43b214

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,1105 @@
...
+            ],
+            vec![
+                0, 0, 255, 0, 0, 0, 0, 255, 0, 0, 255, 255, 255, 0, 0, 0, 0, 255, 0, 0, 0, 0, 255,
+                0, 0,
+            ],
+        );
+
+        let srcs = vec![green, blue, temp];
+
+        let project_path = Path::new(env!("CARGO_MANIFEST_DIR")).parent().unwrap();
```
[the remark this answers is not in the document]
MarWeUMR: This was moved to execution context now.
https://github.com/geo-engine/geoengine/pull/639/commits/3748f16f0228197eb83b00d6b34b662d348ebd2b#diff-cc0868c1f74e67e353cab5c0229f7f9410dc513be5181228ce96c5ddcfb93d9eR150

operators/src/error.rs
```diff
@@ -314,6 +314,20 @@ pub enum Error {
...
 
     AbortTriggerAlreadyUsed,
+
+    SubPathMustNotEscapeBasePath {
+        base: PathBuf,
+        sub_path: PathBuf,
+    },
+
+    InvalidDataProviderConfig,
+    MachineLearningModelNotAvailable,
```
michaelmattig, line 324: better call it NotFound

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,916 @@
...
+            },
+            sources: MultipleRasterSources { rasters: srcs },
+        };
+
+        let mut exe_ctx = MockExecutionContext::new_with_tiling_spec(TilingSpecification::new(
+            (0., 0.).into(),
+            [5, 5].into(),
+        ));
+
+        exe_ctx.initialize_ml_model(model_path);
```
michaelmattig, line 570: please unwrap the result so we now where it failed

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,916 @@
...
```
michaelmattig, line 878: please unwrap the result so we now where it failed

operators/src/pro/ml/xgboost.rs
```diff
@@ -0,0 +1,916 @@
...
+        //    5,   5,  60,  15,  15
+        //    5,   5,  60,  15,  15
+        let temp = make_raster(vec![
+            15, 15, 60, 5, 5, 15, 15, 60, 5, 5, 60, 60, 60, 60, 60, 5, 5, 60, 15, 15, 5, 5, 60, 15,
+            15,
+        ]);
+
+        let srcs = vec![green, blue, temp];
+
+        let model_path = PathBuf::from("pro/ml/xgboost/s2_10m_de_marburg/model.json");
```
michaelmattig, line 863: please use the test_data! macro here (and in the other tests). otherwise the data may not be found, as the working dir is sometimes different, depending on how the tests are run.

operators/src/error.rs
michaelmattig, line 324: better call in ...NotAvailable

services/src/contexts/mod.rs
```diff
@@ -213,6 +214,32 @@ where
...
         op
     }
+
+    fn read_ml_model(&self, path: std::path::PathBuf) -> geoengine_operators::util::Result<String> {
+        let cfg = get_config_element::<crate::util::config::DataProvider>()
+            .map_err(|_| geoengine_operators::error::Error::InvalidDataProviderConfig)?;
+
+        let model_path = cfg.pro_ml_xgboost_model_path;
+
+        let model = std::fs::read_to_string(model_path.join(path))?;
```
michaelmattig, line 224: please use the canionicalize method to ensure the  path doesn't escape the model_path base. This should prevent bad things if `path` is something like `../../`

services/src/pro/contexts/mod.rs
```diff
@@ -120,6 +121,32 @@ where
...
         op
     }
+
+    fn read_ml_model(&self, path: std::path::PathBuf) -> geoengine_operators::util::Result<String> {
+        let cfg = get_config_element::<crate::util::config::DataProvider>()
+            .map_err(|_| geoengine_operators::error::Error::InvalidDataProviderConfig)?;
+
+        let model_path = cfg.pro_ml_xgboost_model_path;
+
+        let model = std::fs::read_to_string(model_path.join(path))?;
```
michaelmattig, line 131: please use the canionicalize method to ensure the  path doesn't escape the model_path base. This should prevent bad things if `path` is something like `../../`

services/src/pro/contexts/mod.rs
```diff
@@ -120,6 +121,32 @@ where
...
+        &mut self,
+        path: std::path::PathBuf,
+        ml_model_str: String,
+    ) -> geoengine_operators::util::Result<()> {
+        let cfg = get_config_element::<crate::util::config::DataProvider>()
+            .map_err(|_| geoengine_operators::error::Error::InvalidDataProviderConfig)?;
+
+        let model_path = cfg.pro_ml_xgboost_model_path;
+
+        std::fs::write(model_path.join(path), ml_model_str)?;
```
michaelmattig, line 146: please use the canionicalize method to ensure the  path doesn't escape the model_path base. This should prevent bad things if `path` is something like `../../`

services/src/contexts/mod.rs
```diff
@@ -213,6 +214,32 @@ where
...
+        &mut self,
+        path: std::path::PathBuf,
+        ml_model_str: String,
+    ) -> geoengine_operators::util::Result<()> {
+        let cfg = get_config_element::<crate::util::config::DataProvider>()
+            .map_err(|_| geoengine_operators::error::Error::InvalidDataProviderConfig)?;
+
+        let model_path = cfg.pro_ml_xgboost_model_path;
+
+        std::fs::write(model_path.join(path), ml_model_str)?;
```
michaelmattig, line 239: please use the canionicalize method to ensure the  path doesn't escape the model_path base. This should prevent bad things if `path` is something like `../../`

operators/src/error.rs
MarWeUMR: Changed with 99d0b9ff66a24c4e316ac5f3e0c44e051d4cb69f

services/src/contexts/mod.rs
MarWeUMR: Changed with 5fe4b7ea1b6d3e328a3e010dc88cbddbadbb3bbb

services/src/pro/contexts/mod.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/5fe4b7ea1b6d3e328a3e010dc88cbddbadbb3bbb

services/src/contexts/mod.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/5fe4b7ea1b6d3e328a3e010dc88cbddbadbb3bbb

operators/src/pro/ml/xgboost.rs
MarWeUMR: fd940cc820ae7165086bbe9e41f6af5fedf83830

operators/src/pro/ml/xgboost.rs
MarWeUMR: https://github.com/geo-engine/geoengine/commit/fd940cc820ae7165086bbe9e41f6af5fedf83830

Settings-default.toml
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/03f16457a9068a96679fa8c005c00009fe606e37

operators/src/error.rs
```diff
@@ -314,6 +314,21 @@ pub enum Error {
...
     AbortTriggerAlreadyUsed,
+
+    SubPathMustNotEscapeBasePath {
+        base: PathBuf,
+        sub_path: PathBuf,
+    },
+
+    InvalidDataProviderConfig,
+    InvalidMachineLearningConfig,
+    NotAvailable,
```
michaelmattig, line 325: I meant `MachineLearningModelNotAvailable` or something like that

operators/Cargo.toml
```diff
@@ -50,6 +50,8 @@ tracing = "0.1"
 tracing-opentelemetry = "0.18"
 typetag = "0.2"
 uuid = { version = "1.1", features = ["serde", "v4", "v5"] }
+xgboost = {git = "https://github.com/MarWeUMR/xgboost-rs", tag = "v0.1.0"}
+ndarray = "*"
```
michaelmattig, line 54: please specify the version and put it under dev-dependencies, as it is only used for tests

services/src/contexts/mod.rs
```diff
@@ -343,3 +380,85 @@ where
...
+
+    use crate::{
+        contexts::{Context, InMemoryContext, SimpleSession},
+        util::config::set_config,
+    };
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn read_model_test() {
+        // TODO: find a better solution for this
+        // necessary to avoid getting a wrong config in the other related tests
```
michaelmattig, line 400: just test both in a single test then, until we find a better solution for it

services/src/contexts/mod.rs
```diff
@@ -343,3 +380,85 @@ where
...
+
+        let model_path = PathBuf::from("xgboost/s2_10m_de_marburg/model.json");
+        let mut model = exe_ctx.read_ml_model(model_path).await.unwrap();
+
+        let actual: String = model.drain(0..277).collect();
+        let expected = "{\"learner\":{\"attributes\":{},\"feature_names\":[],\"feature_types\":[],\"gradient_booster\":{\"model\":{\"gbtree_model_param\":{\"num_parallel_tree\":\"1\",\"num_trees\":\"16\",\"size_leaf_vector\":\"0\"},\"tree_info\":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],\"trees\":[{\"base_weights\":[5.192308E-1,9.722222E-1";
+
+        set_config(
+            "machinelearning.model_defs_path",
+            "./test_data/pro/ml".to_string(),
```
michaelmattig, line 421: better save the original value before overwriting it. Otherwise there might be a conflict if it is changed in the Settings-default, but not here

services/src/contexts/mod.rs
```diff
@@ -343,3 +380,85 @@ where
...
+
+        let model_path = PathBuf::from("xgboost");
+        exe_ctx
+            .write_ml_model(model_path, String::from("model content"))
+            .await
+            .unwrap();
+
+        set_config(
+            "machinelearning.model_defs_path",
+            "./test_data/pro/ml".to_string(),
```
michaelmattig, line 454: same here

services/src/contexts/mod.rs
```diff
@@ -213,6 +216,40 @@ where
...
+        model_sub_path: std::path::PathBuf,
+        ml_model_str: String,
+    ) -> geoengine_operators::util::Result<()> {
+        let cfg = get_config_element::<crate::util::config::MachineLearning>()
+            .map_err(|_| geoengine_operators::error::Error::InvalidMachineLearningConfig)?;
+
+        let model_base_path = cfg.model_defs_path;
+
+        let model_path = canonicalize_subpath(&model_base_path, &model_sub_path)?;
+        tokio::fs::write(model_path.join("model.json"), ml_model_str).await?;
```
michaelmattig, line 249: wouldn't it make more sense to have the filename be specified in the `model_sub_path`?

services/src/pro/contexts/mod.rs
```diff
@@ -120,6 +123,40 @@ where
...
+        model_sub_path: std::path::PathBuf,
+        ml_model_str: String,
+    ) -> geoengine_operators::util::Result<()> {
+        let cfg = get_config_element::<crate::util::config::MachineLearning>()
+            .map_err(|_| geoengine_operators::error::Error::InvalidMachineLearningConfig)?;
+
+        let model_base_path = cfg.model_defs_path;
+
+        let model_path = canonicalize_subpath(&model_base_path, &model_sub_path)?;
+        tokio::fs::write(model_path.join("model.json"), ml_model_str).await?;
```
michaelmattig, line 156: same here

services/src/contexts/mod.rs
```diff
@@ -213,6 +216,40 @@ where
...
+        &mut self,
+        model_sub_path: std::path::PathBuf,
+        ml_model_str: String,
+    ) -> geoengine_operators::util::Result<()> {
+        let cfg = get_config_element::<crate::util::config::MachineLearning>()
+            .map_err(|_| geoengine_operators::error::Error::InvalidMachineLearningConfig)?;
+
+        let model_base_path = cfg.model_defs_path;
+
+        let model_path = canonicalize_subpath(&model_base_path, &model_sub_path)?;
```
michaelmattig, line 248: I think this does only work, if the path already exists. Please check this and use `path_with_base_path` instead if necessary.

services/src/pro/contexts/mod.rs
```diff
@@ -120,6 +123,40 @@ where
...
+        &mut self,
+        model_sub_path: std::path::PathBuf,
+        ml_model_str: String,
+    ) -> geoengine_operators::util::Result<()> {
+        let cfg = get_config_element::<crate::util::config::MachineLearning>()
+            .map_err(|_| geoengine_operators::error::Error::InvalidMachineLearningConfig)?;
+
+        let model_base_path = cfg.model_defs_path;
+
+        let model_path = canonicalize_subpath(&model_base_path, &model_sub_path)?;
```
michaelmattig, line 155: same here

operators/src/error.rs
MarWeUMR: Fingers crossed, that it is correct now https://github.com/geo-engine/geoengine/pull/639/commits/123c769ebd9ef44cce0d569bfaa46c1471ea7885.

operators/Cargo.toml
MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/655449c29c8578e7b27cdd6c4579ead07c50b880

services/src/contexts/mod.rs
MarWeUMR: now resolved with the `serial_test` crate https://github.com/geo-engine/geoengine/pull/639/commits/15511c1bf666395c74b1bd17f8927fcf0cf0ad1a.

services/src/contexts/mod.rs
MarWeUMR: Added a backup variable [here](https://github.com/geo-engine/geoengine/pull/639/commits/15511c1bf666395c74b1bd17f8927fcf0cf0ad1a#diff-b9fca937b91582b0887a2fea16060f5f906844f8a96587673f0f8cdf3882ef87R415-R416)

services/src/contexts/mod.rs
MarWeUMR: see above

services/src/contexts/mod.rs
MarWeUMR: Ok, now the method works like this:

1. check if path escapes somehow
2. try to get the parent dir of the specified file path -> create that directory if neccessary
3. create/overwrite the file

[here](https://github.com/geo-engine/geoengine/pull/639/commits/15511c1bf666395c74b1bd17f8927fcf0cf0ad1a#diff-b9fca937b91582b0887a2fea16060f5f906844f8a96587673f0f8cdf3882ef87R252-R263)

services/src/contexts/mod.rs
MarWeUMR: Exchanged the methods now ([see](https://github.com/geo-engine/geoengine/pull/639#discussion_r1034855518))

services/src/pro/contexts/mod.rs
MarWeUMR: see above

services/src/pro/contexts/mod.rs
MarWeUMR: see above

operators/Cargo.toml
michaelmattig: we usually don't specify the concrete patch. Just set it to "0.15"

datatypes/src/error.rs
```diff
@@ -5,8 +5,9 @@ use crate::{
     spatial_reference::SpatialReference,
 };
 use snafu::{prelude::*, AsErrorSource, ErrorCompat, IntoError};
-use std::{any::Any, convert::Infallible, sync::Arc};
+use std::{any::Any, convert::Infallible, path::PathBuf, sync::Arc};
 
+pub type Result<T, E = Error> = std::result::Result<T, E>;
```
michaelmattig, line 10: this is already defined in utils, just use that one.

operators/Cargo.toml
```diff
@@ -50,11 +50,13 @@ tracing = "0.1"
 tracing-opentelemetry = "0.18"
 typetag = "0.2"
 uuid = { version = "1.1", features = ["serde", "v4", "v5"] }
+xgboost-rs = "0.1.2"
```
michaelmattig, line 53: only specify minor version, not patch

operators/Cargo.toml
```diff
@@ -50,11 +50,13 @@ tracing = "0.1"
...
 
 [dev-dependencies]
 async-stream = "0.3"
 geo-rand = { git = "https://github.com/lelongg/geo-rand", tag = "v0.3.0" }
 rand = "0.8"
+ndarray = "0.15.6"
```
michaelmattig, line 59: only specify minor version, not patch

MarWeUMR: https://github.com/geo-engine/geoengine/pull/639/commits/e6b35477617e9dd0f4484a89bd65ec08f388f378

datatypes/src/error.rs
MarWeUMR: Had to change the return type of `canonicalize_subpath` as well, but that was the actual reason for declaring the other type, so I guess thats ok :)
https://github.com/geo-engine/geoengine/pull/639/commits/2601f19b1351e51893cd0a90d02f9e2e12e4990a

michaelmattig:
bors merge

michaelmattig:
bors retry

michaelmattig:
bors retry

PR #643, opened 2022-10-04
Workflow_monitoring

michaelmattig:

---

<TEXT>


operators/src/processing/reprojection.rs
```diff
@@ -275,13 +276,22 @@ impl RasterOperator for Reprojection {
...
+        Ok(context.initialize_operator(op.boxed(), Box::new(ReprojectionSpan {})))
+    }
+}
+
+#[derive(Clone)]
+struct ReprojectionSpan {}
+
+impl CreateSpan for ReprojectionSpan {
+    fn create_span(&self) -> Span {
+        span!(Level::TRACE, "Reprojection")
```
ChristianBeilschmidt, line 294: Maybe we should create a macro for that.

operators/src/util/raster_stream_to_png.rs
```diff
@@ -23,6 +24,10 @@ pub async fn raster_stream_to_png_bytes<T, C: QueryContext>(
 where
     T: Pixel,
 {
+    // TODO: rather use `span.in_scope` because it is safer inside async functions (cf. https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.enter)
+    let span = span!(Level::TRACE, "raster_stream_to_png_bytes");
+    let _enter = span.enter();
```
ChristianBeilschmidt, line 29: can we use in_scope here?

services/src/bin/main.rs
```diff
@@ -58,7 +96,7 @@ fn initialize_logging() -> Option<FileLogWriterHandle> {
         // Thus, we use UTC time instead.
         flexi_logger::DeferredNow::force_utc();
 
-        let (file_writer, fw_handle) = FileLogWriter::builder(file_spec)
+        let (file_writer, _fw_handle) = FileLogWriter::builder(file_spec)
```
ChristianBeilschmidt, line 99: what changed around here?

services/src/error.rs
```diff
@@ -56,6 +56,9 @@ pub enum Error {
     Proj {
         source: proj::ProjError,
     },
+    Trace {
+        source: opentelemetry::trace::TraceError,
+    },
```
ChristianBeilschmidt, line 61: Can't we wrap that into an Opentelemetry error?

services/src/error.rs
```diff
@@ -521,3 +524,9 @@ impl From<tokio::task::JoinError> for Error {
         Error::TokioJoin { source }
     }
 }
+
+impl From<opentelemetry::trace::TraceError> for Error {
+    fn from(source: opentelemetry::trace::TraceError) -> Self {
+        Error::Trace { source }
+    }
+}
```
ChristianBeilschmidt, line 532: Use `#[snafu(context(false))]` instead

services/src/pro/contexts/in_memory.rs
```diff
@@ -22,6 +22,8 @@ use snafu::ResultExt;
 use std::path::PathBuf;
 use std::sync::Arc;
```
ChristianBeilschmidt, line 24:
```suggestion
```

services/src/pro/contexts/mod.rs
```diff
@@ -24,3 +37,202 @@ pub trait ProContext: Context<Session = UserSession> {
...
+    }
+
+    fn initialize_operator(
+        &self,
+        op: Box<dyn geoengine_operators::engine::InitializedRasterOperator>,
+        span: Box<dyn geoengine_operators::engine::CreateSpan>,
+    ) -> Box<dyn geoengine_operators::engine::InitializedRasterOperator> {
+        Box::new(InitializedProcessorStatistics::statistics_with_id(
+            op,
+            "gdal_source".to_string(),
```
ChristianBeilschmidt, line 105: why is everything a gdal source?

services/src/pro/contexts/mod.rs
```diff
@@ -24,3 +37,202 @@ pub trait ProContext: Context<Session = UserSession> {
...
+    D: DatasetDb<S>,
+    L: LayerProviderDb,
+    S: Session,
+{
+    dataset_db: Arc<D>,
+    layer_provider_db: Arc<L>,
+    thread_pool: Arc<ThreadPool>,
+    session: S,
+    tiling_specification: TilingSpecification,
+}
```
ChristianBeilschmidt, line 52: Why did you need to create that?

ChristianBeilschmidt (changes requested):

Settings-default.toml
```diff
@@ -110,3 +110,7 @@ issuer = ""
 client_id = ""
 redirect_uri = ""
 scopes = [ ]
+
+[open_telemetry]
+enabled = false
+endpoint = "127.0.0.1:6831"
```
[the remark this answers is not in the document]
michaelmattig: this is the address of the JAEGER instance where Geo Engine sends the traces to. It might be even on a different machine so we need the configuration

operators/src/engine/execution_context.rs
```diff
@@ -28,6 +28,12 @@ pub trait ExecutionContext: Send
 {
     fn thread_pool(&self) -> &Arc<ThreadPool>;
     fn tiling_specification(&self) -> TilingSpecification;
+
+    fn initialize_operator(
+        &self,
+        op: Box<dyn InitializedRasterOperator>,
+        span: Box<dyn CreateSpan>,
```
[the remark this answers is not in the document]
michaelmattig: I can't make the function generic over the `CreateSpan` because then it can no longer be made into a trait object.

operators/src/pro/meta/statistics.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub fn statistics_with_id(source: S, id: String, span: Box<dyn CreateSpan>) -> Self {
+        Self { source, id, span }
+    }
+}
+
+impl InitializedRasterOperator
+    for InitializedProcessorStatistics<Box<dyn InitializedRasterOperator>>
+{
+    fn result_descriptor(&self) -> &RasterResultDescriptor {
+        debug!("[{}] | raster result descriptor", self.id);
```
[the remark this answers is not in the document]
michaelmattig: I guess we have to decide in general, what is a log message and what is a tracing event. It's possible to convert one thing into the other but they are not the same. E.g. the swagger UI URL log message should most likely not be a tracing event.

services/src/pro/contexts/mod.rs
michaelmattig: the execution context is now responsible for wrapping the operators with a statistics operator. As it should only do so in the pro version, we need a separate ExecutionContextImpl for Pro. Previously we used the same in pro and free.

services/src/error.rs
michaelmattig: why? maybe we want to see why the tracing failed.

operators/src/processing/reprojection.rs
michaelmattig: if we decide that this is the way to go for span creation we should.

operators/src/pro/adapters/stream_statistics_adapter.rs
```diff
@@ -0,0 +1,127 @@
...
+            "[{}] | poll_next_count: {}",
+            *this.id,
+            *this.poll_next_count
+        );
+
+        let v = ready!(this.stream.as_mut().poll_next(cx));
+        match v {
+            Some(_) => {
+                *this.element_count += 1;
+                debug!(
```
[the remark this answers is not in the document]
michaelmattig: it's a judgement call. I used Johannes classification of the log events here. You can suggest a different one if you don't like it.

operators/src/pro/adapters/stream_statistics_adapter.rs
```diff
@@ -0,0 +1,127 @@
...
+        let mut this = self.project();
+        *this.poll_next_count += 1;
+
+        let _enter = this.span.enter();
+
+        trace!(
+            "[{}] | poll_next_count: {}",
+            *this.id,
+            *this.poll_next_count
+        );
```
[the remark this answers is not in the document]
michaelmattig: 60deff1f01a061f9903fafef13cdf967b4f0cc08

operators/src/pro/adapters/stream_statistics_adapter.rs
```diff
@@ -0,0 +1,127 @@
...
+                debug!(
+                    "[{}] | poll_next_count: {} | element_count: {} | next",
+                    *this.id, *this.poll_next_count, *this.element_count
+                );
+            }
+            None => {
+                debug!(
+                    "[{}] | poll_next_count: {} | element_count: {} | empty",
+                    *this.id, *this.poll_next_count, *this.element_count
+                );
```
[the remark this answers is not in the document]
michaelmattig: 60deff1f01a061f9903fafef13cdf967b4f0cc08

operators/src/pro/meta/statistics.rs
```diff
@@ -0,0 +1,207 @@
...
+        match processor_result {
+            Ok(p) => {
+                let res_processor = match p {
+                    TypedRasterQueryProcessor::U8(p) => TypedRasterQueryProcessor::U8(Box::new(
+                        ProcessorStatisticsProcessor::statistics_with_id(
+                            p,
+                            self.id.clone(),
+                            self.span.clone(),
+                        ),
+                    )),
```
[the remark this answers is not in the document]
michaelmattig: I don't think so. Neither call_generic_raster_processor nor call_on_generic_raster_processor helps me here. Or am I missing something? Please make a concrete suggestion.

operators/src/util/raster_stream_to_png.rs
michaelmattig: yes but this would either mean wrapping the whole function body inside a closure or introduce another method, that's why I didn't do it yet.

services/src/bin/main.rs
```diff
@@ -1,55 +1,93 @@
...
+            tracing_opentelemetry::layer().with_tracer(tracer);
+        Some(opentelemetry)
+    } else {
+        None
+    };
+
+    let registry = registry.with(open_telemetry_layer);
+    let registry = registry.with(console_layer(&logging_config));
+    let registry = registry.with(file_layer(&logging_config));
+    registry.init();
```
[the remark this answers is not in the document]
michaelmattig: 1. what do you mean? this is already a function with the whole purpose of setting up the logging and then calling the actual start method. If I extract it into another method it would be exactly the same.
2. this is not a service but a global logging/tracing subscriber. How does this to relate to actix services and routing?

services/src/bin/main.rs
michaelmattig: nothing. I just ignore the handle at a different location in the code. We didn't use it before too

services/src/error.rs
michaelmattig: This IS the opentelemetry error.

services/src/pro/contexts/in_memory.rs
michaelmattig: ?

services/src/pro/contexts/mod.rs
michaelmattig: yeah this is wrong. I could add a `get_id` method to `CreateSpan` to set it here, or just get rid of the "id" here. It's somewhat redundant because it would always be the same for every span.

operators/src/pro/meta/statistics.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: `call_generic_raster_processor` should be it… why doesn't it work here?

services/src/error.rs
ChristianBeilschmidt: It is a trace error, right?
Aren't there other errors with respect to our spectacular remote logging?

services/src/error.rs
ChristianBeilschmidt: But you are implementing exactly what context(false) auto-implements

operators/src/engine/execution_context.rs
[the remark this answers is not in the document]
michaelmattig: I changed some stuff and it is no longer boxed 68617472345bdacfa1a856184f118722e110b5b3

operators/src/pro/meta/statistics.rs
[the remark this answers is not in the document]
michaelmattig: because we `match` on a `TypedRasterQueryProcessor` and the macro matches a `RasterDataType`

services/src/error.rs
michaelmattig: I don't know if it is `spectacular` but remote logging is called "tracing" and the `trace` module of opentelemetry is the "API for tracing applications and libraries." and thus the ` TraceError` should be exactly what we need.

services/src/error.rs
michaelmattig: ed702fe821950f56ee1a6ab0d56aacc25adbf96c

services/src/pro/contexts/mod.rs
michaelmattig: I removed the id for now because it was exactly the same as the span, thus carried no information. 68617472345bdacfa1a856184f118722e110b5b3

operators/src/pro/meta/statistics.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: We have no macro from TypedRasterQueryProcessor to TypedRasterQueryProcessor?
Maybe you can create one?

operators/src/mock/mock_dataset_data_source.rs
```diff
@@ -130,6 +131,10 @@ impl VectorOperator for MockDatasetDataSource {
         }
         .boxed())
     }
+
+    fn span(&self) -> CreateSpan {
+        || span!(Level::TRACE, MockDatasetDataSource::TYPE_NAME)
+    }
```
ChristianBeilschmidt, line 137: Is there any way we can use the `OperatorName` trait for that?

operators/src/plot/histogram.rs
```diff
@@ -141,6 +146,10 @@ impl PlotOperator for Histogram {
             }
         })
     }
+
+    fn span(&self) -> CreateSpan {
+        || span!(Level::TRACE, Histogram::TYPE_NAME)
+    }
```
ChristianBeilschmidt, line 152: Maybe we can at least use a macro for this `fn`. This way, we can easily change the behavior for all operators.

operators/src/mock/mock_dataset_data_source.rs
michaelmattig: I do use the `OperatorName` trait, as the  `TYPE_NAME` field is part of it. However, like I tried to explain in the Trello card, we can't use the trait as a generic. We can only input the `TYPE_NAME` of a concrete type. Otherwise Rust complains on not being able to caputre the outer generic of a function. This has to do with the way the `span!` macro works, as far as I undrstand.

operators/src/mock/mock_feature_collection_source.rs
```diff
@@ -83,6 +85,10 @@ where
 
 pub type MockFeatureCollectionSource<G> = SourceOperator<MockFeatureCollectionSourceParams<G>>;
 
+impl<G: Geometry + ArrowTyped> OperatorName for MockFeatureCollectionSource<G> {
+    const TYPE_NAME: &'static str = "MockFeatureCollectionSource";
+}
```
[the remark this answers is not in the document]
michaelmattig: yes, this is an *impl*ementation of that trait, see line 88

operators/src/engine/operator.rs
```diff
@@ -28,18 +29,30 @@ pub trait OperatorData {
 pub trait RasterOperator:
     CloneableRasterOperator + OperatorData + Send + Sync + std::fmt::Debug
 {
-    async fn initialize(
+    // internal logic of the operator
```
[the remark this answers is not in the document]
michaelmattig: 11c4d2df4f4ea9e34f970b7bebfb1f7dfd2b680a

operators/src/engine/mod.rs
```diff
@@ -168,25 +168,4 @@ macro_rules! call_on_bi_generic_raster_processor {
...
-        Box::new(self.clone())
-    }
-}
-
-impl Clone for Box<dyn CreateSpan> {
-    fn clone(&self) -> Box<dyn CreateSpan> {
-        self.clone_create_span()
-    }
-}
+pub type CreateSpan = fn() -> Span;
```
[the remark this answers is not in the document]
michaelmattig: 11c4d2df4f4ea9e34f970b7bebfb1f7dfd2b680a

michaelmattig:
> Please also write a changelog.

&

> I added an entry to [CHANGELOG.md](https://github.com/geo-engine/geoengine/pull/CHANGELOG.md) if knowledge of this change could be valuable to users

knowledge of monitoring is not valuable to users so I didn't add it as to not bloat the changelog

operators/src/mock/mock_dataset_data_source.rs
ChristianBeilschmidt: Is this the issue?
https://github.com/tokio-rs/tracing/issues/1047

operators/src/plot/histogram.rs
ChristianBeilschmidt: Would be nice, but we could have a normal macro instead here.

operators/src/mock/mock_dataset_data_source.rs
michaelmattig: no, the `'static` is not the problem, but the string having to be `const` in order for the macro to work

operators/src/plot/histogram.rs
michaelmattig: de3d837b15aa93ceeaa690437b8747d2d5457067

ChristianBeilschmidt (approved):

PR #650, opened 2022-10-13
Wxs_api_docs

michaelmattig:

---

<TEXT>


michaelmattig:
[WXS annotieren](https://trello.com/c/XeyWnQPb/1043-wxs-annotieren)

services/src/error.rs
```diff
@@ -393,6 +393,14 @@ pub enum Error {
...
         source: crate::pro::users::OidcError,
     },
+
+    #[snafu(display(
+        "Could not resolve a Proj string for this SpatialReference: {}",
+        spatial_ref
+    ))]
+    ProjStringUnresolvable {
+        spatial_ref: SpatialReference,
+    },
```
ChristianBeilschmidt, line 403: I'm not sure that this should be an error at the root level. Could you put it where it happens?

services/src/handlers/wfs.rs
```diff
@@ -36,133 +40,138 @@ where
...
-/// </wfs:WFS_Capabilities>
-/// ```
+/// Get WFS Capabilities
+#[utoipa::path(
+    tag = "OGC WFS",
+    get,
+    path = "/wfs/{workflow}?request=GetCapabilities",
+    responses(
+        (status = 200, description = "OK", content_type = "text/xml", body = String,
+            // TODO: add example when utoipa supports more than just json examples
```
ChristianBeilschmidt, line 66: Is there an issue to refer to?

Or maybe output it as a JSON text blob?

services/src/handlers/wfs.rs
```diff
@@ -416,8 +436,8 @@ async fn get_feature<C: Context>(
 
     // TODO: validate request further
 
-    if request.type_names.feature_type == "93d6785e-5eea-4e0e-8074-e7f78733d988" {
-        return get_feature_mock(request);
+    if request.typeNames.feature_type == "93d6785e-5eea-4e0e-8074-e7f78733d988" {
```
ChristianBeilschmidt, line 439: Please use snake_case

services/src/handlers/wfs.rs
```diff
@@ -438,7 +458,7 @@ async fn get_feature<C: Context>(
 
     // TODO: use a default spatial reference if it is not set?
     let request_spatial_ref: SpatialReference = request
-        .srs_name
+        .srsName
```
ChristianBeilschmidt, line 461: also here

services/src/handlers/wms.rs
```diff
@@ -34,96 +33,108 @@ where
...
 
-async fn wms_handler<C: Context>(
+/// Get WMS Capabilities
+#[utoipa::path(
+    tag = "OGC WMS",
+    get,
+    path = "/wms/{workflow}?request=GetCapabilities",
+    responses(
+        (status = 200, description = "OK", content_type = "text/xml", body = String,
+            // TODO: add example when utoipa supports more than just json examples
```
ChristianBeilschmidt, line 64: cf. other comment

services/src/handlers/wms.rs
```diff
@@ -531,7 +585,7 @@ mod tests {
 
         let (_, id) = register_ndvi_workflow_helper(&ctx).await;
 
-        let req = actix_web::test::TestRequest::get().uri(&format!("/wms/{id}?service=WMS&version=1.3.0&request=GetMap&layers={id}&styles=&width=335&height=168&crs=EPSG:4326&bbox=-90.0,-180.0,90.0,180.0&format=image/png&transparent=FALSE&bgcolor=0xFFFFFF&exceptions=XML&time=2014-04-01T12%3A00%3A00.000%2B00%3A00", id = id.to_string())).append_header((header::AUTHORIZATION, Bearer::new(session_id.to_string())));
+        let req = actix_web::test::TestRequest::get().uri(&format!("/wms/{id}?service=WMS&version=1.3.0&request=GetMap&layers={id}&styles=&width=335&height=168&crs=EPSG:4326&bbox=-90.0,-180.0,90.0,180.0&format=image/png&transparent=FALSE&bgcolor=0xFFFFFF&exceptions=application/json&time=2014-04-01T12%3A00%3A00.000%2B00%3A00", id = id.to_string())).append_header((header::AUTHORIZATION, Bearer::new(session_id.to_string())));
```
ChristianBeilschmidt, line 588: Does this reflect our UI?
What happens if the exceptions should be XML?

services/src/ogc/util.rs
```diff
@@ -280,9 +303,28 @@ pub fn ogc_endpoint_url(base: &Url, protocol: OgcProtocol, workflow: WorkflowId)
...
+    pub fn new(request: &'a str) -> Self {
+        Self { request }
+    }
+}
+
+impl Guard for OgcRequestGuard<'_> {
+    fn check(&self, ctx: &GuardContext<'_>) -> bool {
+        ctx.head().uri.query().map_or(false, |q| {
+            q.contains(&format!("request={}", self.request))
+                || q.contains(&format!("REQUEST={}", self.request))
```
ChristianBeilschmidt, line 320: why not case insensitive?

services/src/ogc/wcs/request.rs
```diff
@@ -221,20 +260,29 @@ mod tests {
...
                 boundingbox: WcsBoundingbox {
                     bbox: [-81., -162., 81., 162.],
-                    spatial_reference: Some(SpatialReference::epsg_4326()),
+                    spatial_reference: Some(SpatialReference::new(
+                        SpatialReferenceAuthority::Epsg,
+                        4326
+                    )),
                 },
-                gridbasecrs: SpatialReference::epsg_4326(),
+                gridbasecrs: SpatialReference::new(SpatialReferenceAuthority::Epsg, 4326),
```
ChristianBeilschmidt, line 275: maybe we should get the shorthand method again if we need it often?

services/src/ogc/wfs/request.rs
```diff
@@ -1,32 +1,42 @@
-use crate::ogc::util::{parse_bbox, parse_spatial_resolution_option, parse_time_option};
+#![allow(non_snake_case)]
+// TODO: remove this and use rename_all once utoipa supports this for `IntoParams`: https://github.com/juhaku/utoipa/issues/270
```
ChristianBeilschmidt, line 2: uh, that's ugly. Is it possible to rename the parameter names with an utoipa attribute macro?

services/src/ogc/wfs/request.rs
```diff
@@ -44,32 +59,52 @@ pub struct DescribeFeatureType {
     pub output_format: String, // TODO
```
ChristianBeilschmidt, line 59: Do you know… TODO… what?

services/src/ogc/wms/request.rs
```diff
@@ -1,60 +1,85 @@
+#![allow(non_snake_case)]
+// TODO: remove this and use rename_all once utoipa supports this for `IntoParams`: https://github.com/juhaku/utoipa/issues/270
```
ChristianBeilschmidt, line 2: cf. other comment

ChristianBeilschmidt (changes requested):

services/src/handlers/wfs.rs
michaelmattig: There is no GitHub issue. It's more of a missing feature I guess. We could output it as JSON text, but then the content type cannot be text/xml. We can either have the correct type or an example.

services/src/ogc/wfs/request.rs
michaelmattig: unfortunately this is not possible. I was also surprised that this is not just a lint you can ignore for a line. Rust is really picky here.

services/src/ogc/util.rs
michaelmattig: I can do it case insensitive here if you want but it would be inconsistent because all the other parameters are also only supported all/no caps. Also that would require a regex I guess.

services/src/error.rs
michaelmattig: why not? This is a user input and they should should be reported if they input it wrong. Do you want me to wrap it in a WCS error? "WCS Error: Could not resolve a Proj string for this SpatialReference:"? The user should know that they are in a wcs request though. It's not like that this is some deep internal error.

michaelmattig: it happens in crate::api::model::datatypes. I don't think it would be any better to have a separate ApiError I guess so I would leave it here. Do you have a suggestions where to put it?

services/src/handlers/wfs.rs
michaelmattig: The OGC API specification wants it to be camelcase and we cannot rewrite it. cf. my TODO in the ogc/wfs/request.rs

services/src/ogc/wfs/request.rs
michaelmattig: make an enum out of it or parse it or... do anything with it. Currently we don't use this at all. We don't support this API call. Initially I still mapped them as a placeholder but I can also delete them.

services/src/handlers/wfs.rs
michaelmattig: cf. other comment

services/src/handlers/wms.rs
michaelmattig: cf. other comment

services/src/ogc/wms/request.rs
michaelmattig: cf. other comment

services/src/ogc/wfs/request.rs
ChristianBeilschmidt: Could you at least put it on the struct(s)?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e6bf3bac14f843e9040bcd11c3468bf4

And maybe you could comment on the issue. It looks like everything is fine there from the answer.

services/src/ogc/wfs/request.rs
ChristianBeilschmidt: Could you add it then?

services/src/ogc/util.rs
ChristianBeilschmidt: What about lowercasing the whole string and then using contains?

services/src/handlers/wms.rs
ChristianBeilschmidt: So this did not fix it?
https://github.com/juhaku/utoipa/pull/41

services/src/handlers/wfs.rs
ChristianBeilschmidt: What about https://github.com/juhaku/utoipa/pull/41 ?

michaelmattig: I don't exactly know what this PR ist about, all I know is that the `example` parameter of the utoipa path macro only supports Json Values.

services/src/handlers/wms.rs
michaelmattig: this has nothing to do with utoipa but only with our implementation of WMS. We would have to parse the exceptions format and render them differently.

services/src/ogc/util.rs
michaelmattig: AFAIK GetCapabilities etc. have to be in camelCase.

services/src/ogc/wfs/request.rs
michaelmattig: I don't see the benefit. I'd rather delete it if it really bothers you.

services/src/ogc/wms/request.rs
michaelmattig: 9590973abc1e0abc68c781c313fbd2837a0dba46

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors merge

PR #651, opened 2022-10-17
remove potential panics from `unimplemented` blocks

ChristianBeilschmidt:

---

<TEXT>

### DIFF
--- a/datatypes/src/operations/image/colorizer.rs
+++ b/datatypes/src/operations/image/colorizer.rs
@@ -288,7 +288,17 @@ impl Colorizer {
         let breakpoints = match self {
             Self::LinearGradient { breakpoints, .. }
             | Self::LogarithmicGradient { breakpoints, .. } => breakpoints,
-            _ => unimplemented!("Must never call `color_table` for types without breakpoints"),
+
+            _ => {
+                // return empty color table for potential wrong usage
+
+                debug_assert!(
+                    false,
+                    "Must never call `color_table` for types without breakpoints"
+                );
+
+                return Vec::new();
+            }
         };
 
         let smallest_breakpoint_value = *breakpoints[0].value;
--- a/operators/src/source/gdal_source/loading_info.rs
+++ b/operators/src/source/gdal_source/loading_info.rs
@@ -412,10 +412,14 @@ impl Iterator for NetCdfCfGdalLoadingInfoPartIterator {
             return None;
         }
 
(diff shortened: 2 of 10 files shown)


ChristianBeilschmidt:
[GdalSource: Panics entfernen](https://trello.com/c/7eqRGKbL/1037-gdalsource-panics-entfernen)

datatypes/src/operations/image/colorizer.rs
```diff
@@ -288,7 +288,17 @@ impl Colorizer {
...
+
+            _ => {
+                // return empty color table for potential wrong usage
+
+                debug_assert!(
+                    false,
+                    "Must never call `color_table` for types without breakpoints"
+                );
+
+                return Vec::new();
```
michaelmattig, line 300: why don't you change the return type of this method to `Result`? Would be much cleaner.

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -412,10 +412,14 @@ impl Iterator for NetCdfCfGdalLoadingInfoPartIterator {
             return None;
         }
 
-        let steps_between = self
+        let steps_between = match self
```
michaelmattig, line 415: why use a match instead of a `?`?

services/src/lib.rs
```diff
@@ -20,7 +20,12 @@
     clippy::wildcard_imports
 )]
 // enable some restriction lints
-#![warn(clippy::print_stdout, clippy::print_stderr, clippy::dbg_macro)]
+#![warn(
+    clippy::dbg_macro,
+    clippy::print_stderr,
+    clippy::print_stdout,
+    clippy::unimplemented
```
michaelmattig, line 27: is there maybe a lint for unwrap outside of tests?

datatypes/src/operations/image/colorizer.rs
ChristianBeilschmidt: This isn't called in practice and now it just does not panic.

operators/src/source/gdal_source/loading_info.rs
ChristianBeilschmidt: You cannot use ? because it is a Result that is wrapped in an Option.

michaelmattig (approved):

michaelmattig:
bors retry

PR #655, opened 2022-10-19
detect connection close and cancel stream

michaelmattig:

---

<TEXT>


michaelmattig:
[[EXIST PII] A3.1.1 Abbrechen von Workflow-Verarbeitungen](https://trello.com/c/IVsvY8mG/1028-exist-pii-a311-abbrechen-von-workflow-verarbeitungen)

operators/src/engine/execution_context.rs
```diff
@@ -105,9 +106,12 @@ impl MockExecutionContext {
     }
 
     pub fn mock_query_context(&self, chunk_byte_size: ChunkByteSize) -> MockQueryContext {
+        let (trigger, valve) = Valve::new();
         MockQueryContext {
             chunk_byte_size,
             thread_pool: self.thread_pool.clone(),
+            valve,
+            valve_trigger: Some(trigger),
```
ChristianBeilschmidt, line 114: strange name, can you name it more accordingly?
stream cancel callback or idk

operators/src/engine/query.rs
```diff
@@ -43,37 +44,51 @@ impl TestDefault for ChunkByteSize {
 pub trait QueryContext: Send + Sync {
     fn chunk_byte_size(&self) -> ChunkByteSize;
     fn thread_pool(&self) -> &Arc<ThreadPool>;
+
+    fn valve(&self) -> &Valve;
+    fn valve_trigger(&mut self) -> Option<Trigger>;
```
ChristianBeilschmidt, line 49: especially here

operators/src/source/gdal_source/mod.rs
```diff
@@ -653,7 +655,10 @@ where
             tiling_strategy.tile_size_in_pixels,
         );
 
-        Ok(filled_stream.boxed())
+        // TODO: automatically wrap in all operators. Doing it like for the statistics opeator is not possible
+        //       because the cancellation should be per query and not per QueryProcessor. Thus the cancellation
+        //       must be part of the QueryContext and not the ExecutionContext.
+        Ok(ctx.valve().wrap(filled_stream).boxed())
```
ChristianBeilschmidt, line 661: What is the difference between this and the functionality within `tokio`?
Moreover, could we have some custom struct that wraps this so that we are not dependent on stream-cancel? It seems this isn't being updated for over a year. We should lower the dependencies since it is within our core.

operators/src/util/raster_stream_to_png.rs
```diff
@@ -25,8 +28,21 @@ where
...
+    conn_close.map(|c| {
+        crate::util::spawn(async move {
+            if c.await.is_ok() {
+                if let Some(trigger) = valve_trigger {
+                    info!("shoot trigger");
+                    trigger.cancel();
+                }
+            }
+        })
+    });
```
ChristianBeilschmidt, line 44: Can this future maybe implement Abortable and then cancel it on being aborted?

I'm unsure about the pistol and trigger terminology. Maybe we should stay with stream and cancel :wink: .

operators/src/util/raster_stream_to_png.rs
```diff
@@ -60,9 +76,10 @@ where
                 Err(error) => futures::future::err(error),
             }
         })
-        .await?;
+        .await;
 
-    Ok(output_tile.grid_array.to_png(width, height, &colorizer)?)
+    info!("output tile:  {}", output_tile.is_ok()); // TODO: remove
```
ChristianBeilschmidt, line 81: Then remove?

services/Cargo.toml
```diff
@@ -47,6 +47,7 @@ image = "0.24"
 lazy_static = "1.4"
 log = "0.4"
 mime = "0.3"
+nix = "0.25"
```
ChristianBeilschmidt, line 50: Doesn't tie this the program to linux only?

services/src/contexts/mod.rs
```diff
@@ -74,13 +75,18 @@ pub trait Context: 'static + Send + Sync + Clone {
 pub struct QueryContextImpl {
     chunk_byte_size: ChunkByteSize,
     pub thread_pool: Arc<ThreadPool>,
```
ChristianBeilschmidt, line 77: I just saw that… could you remove `pub`… we have an accessor below?

services/src/contexts/mod.rs
```diff
@@ -74,13 +75,18 @@ pub trait Context: 'static + Send + Sync + Clone {
...
+    valve: Valve,
+    valve_trigger: Option<Trigger>,
```
ChristianBeilschmidt, line 79: also here, maybe terminology

services/src/util/server.rs
```diff
@@ -294,3 +300,47 @@ pub(crate) fn log_server_info() -> Result<()> {
...
     HttpResponse::NotImplemented().finish()
 }
+
+pub struct SocketFd(pub RawFd);
+
+// attach the connections socket file descriptor to the connection data
+// TODO: the socket file descriptor might be re-used by a subsequent connection before we notice that
+//       the connections was closed. We maybe need a global list of all open connections and their socket file descriptors
+//       or some other way to invalidate the socket file descriptor in the old connection when a new one arrives.
+pub fn connection_init(connection: &dyn Any, data: &mut Extensions) {
```
ChristianBeilschmidt, line 310: we should add a test here

services/src/util/server.rs
```diff
@@ -294,3 +300,47 @@ pub(crate) fn log_server_info() -> Result<()> {
...
+            data.insert(SocketFd(fd));
+        }
+    }
+}
+
+// start a new task that monitors the request's socket file descriptor and tries to detect when the connection is closed.
+// The returned join handle can be awaited to get notified when the connection is closed.
+// TODO: further verify the heuristic used here to determine the socket status from the `recv` return value
+pub fn connection_closed(req: &HttpRequest) -> Option<JoinHandle<()>> {
+    info!("<< Monitoring req conn >>"); // TODO: remove
```
ChristianBeilschmidt, line 324: remove?

services/src/util/server.rs
```diff
@@ -294,3 +300,47 @@ pub(crate) fn log_server_info() -> Result<()> {
...
+        let fd = fd.0;
+        crate::util::spawn(async move {
+            let mut data = vec![];
+            // TODO: add a timeout to avoid infinite looping
+            loop {
+                let r = nix::sys::socket::recv(fd, data.as_mut_slice(), MsgFlags::MSG_PEEK);
+
+                match r {
+                    Ok(0) | Err(Errno::EBADF) => {
+                        info!(">> A connection closed"); // TODO: remove
```
ChristianBeilschmidt, line 335: remove?

services/src/util/server.rs
```diff
@@ -294,3 +300,47 @@ pub(crate) fn log_server_info() -> Result<()> {
...
+// start a new task that monitors the request's socket file descriptor and tries to detect when the connection is closed.
+// The returned join handle can be awaited to get notified when the connection is closed.
+// TODO: further verify the heuristic used here to determine the socket status from the `recv` return value
+pub fn connection_closed(req: &HttpRequest) -> Option<JoinHandle<()>> {
+    info!("<< Monitoring req conn >>"); // TODO: remove
+    req.conn_data::<SocketFd>().map(|fd| {
+        let fd = fd.0;
+        crate::util::spawn(async move {
+            let mut data = vec![];
+            // TODO: add a timeout to avoid infinite looping
```
ChristianBeilschmidt, line 329: can't you tie this to the actual future?

There is a join variant for futures that says: emit the first one that returns… maybe something like this.

services/src/util/server.rs
```diff
@@ -294,3 +300,47 @@ pub(crate) fn log_server_info() -> Result<()> {
...
+            loop {
+                let r = nix::sys::socket::recv(fd, data.as_mut_slice(), MsgFlags::MSG_PEEK);
```
ChristianBeilschmidt, line 331: What if the fd is invalid? Can we assure that this does not crash?

services/src/util/server.rs
```diff
@@ -294,3 +300,47 @@ pub(crate) fn log_server_info() -> Result<()> {
...
+
+// attach the connections socket file descriptor to the connection data
+// TODO: the socket file descriptor might be re-used by a subsequent connection before we notice that
+//       the connections was closed. We maybe need a global list of all open connections and their socket file descriptors
+//       or some other way to invalidate the socket file descriptor in the old connection when a new one arrives.
+pub fn connection_init(connection: &dyn Any, data: &mut Extensions) {
+    if let Some(sock) = connection.downcast_ref::<TcpStream>() {
+        let fd = sock.as_raw_fd();
+        if let Ok(fd) = NonZeroI32::try_from(fd) {
+            let fd = RawFd::from(fd);
```
ChristianBeilschmidt, line 314: Could you maybe propose this to the actix-web community and see if they agree that this is one way to go?

ChristianBeilschmidt (changes requested):

operators/src/source/gdal_source/mod.rs
michaelmattig: the difference is that you can re-use the valve. You can wrap multiple streams with it and abort them all at the same time. I think we need this.

operators/src/engine/execution_context.rs
michaelmattig: we can talk about the terminology if we decide we actually want to do it this way

operators/src/util/raster_stream_to_png.rs
michaelmattig: what future do you mean? `conn_close` shouldn't be aborted, it finishes when the connection is aborted and then cancels the query processors.

operators/src/util/raster_stream_to_png.rs
michaelmattig: This is an experimental PR and I left this outputs on purpose for you to see what happens.

services/Cargo.toml
michaelmattig: yes, windows sockets work differently. We have to think about whether we want a different solution for windows, support workflow cancellation only for linux, or tie geo engine to linux entirely. I haven't been able to get geo engine working on windows and I really tried. So we wouldn't really lose anything here.

services/src/util/server.rs
michaelmattig: what would you want to test? There is no logic happening here.

services/src/util/server.rs
michaelmattig: it's for you to comprehend the happenings

services/src/util/server.rs
michaelmattig: it's for you to comprehend the happenings

services/src/util/server.rs
michaelmattig: 👍

services/src/util/server.rs
michaelmattig: nothing can crash here, we would just get EBADF.

services/src/util/server.rs
michaelmattig: No. As you also realized, this works only under Linux, so this is not a general solution we could propose to actix.

operators/src/source/gdal_source/mod.rs
ChristianBeilschmidt: You omitted the second question, but this is super important.

services/src/util/server.rs
ChristianBeilschmidt: Anyway, we should discuss it with them to get their experise.

operators/src/engine/execution_context.rs
michaelmattig: I changed the name to `abort_registration` and `abort_trigger` db426f1844ce4883ad3021252bbf6bbfee69eeb5

operators/src/engine/query.rs
michaelmattig: I changed the name to abort_registration and abort_trigger https://github.com/geo-engine/geoengine/commit/db426f1844ce4883ad3021252bbf6bbfee69eeb5

services/Cargo.toml
michaelmattig: I made the dependency contingent on the target platform being linux

services/src/contexts/mod.rs
michaelmattig: cf my other comments about the renaming

services/src/util/server.rs
michaelmattig: I removed it somewhere here db426f1844ce4883ad3021252bbf6bbfee69eeb5

services/src/util/server.rs
michaelmattig: I discussed with the core decelopers. cf. the card in trello

operators/src/engine/query.rs
```diff
@@ -43,37 +51,96 @@ impl TestDefault for ChunkByteSize {
 pub trait QueryContext: Send + Sync {
     fn chunk_byte_size(&self) -> ChunkByteSize;
     fn thread_pool(&self) -> &Arc<ThreadPool>;
+
+    fn abort_registration(&self) -> &QueryAbortRegistration;
+    fn abort_trigger(&mut self) -> Result<QueryAbortTrigger>;
+}
+
+pub struct QueryAbortRegistration {
```
ChristianBeilschmidt, line 59: Could you briefly describe what this is?

operators/src/engine/query.rs
```diff
@@ -43,37 +51,96 @@ impl TestDefault for ChunkByteSize {
...
+
+    pub fn wrap<S: Stream>(&self, stream: S) -> QueryAbortWrapper<S> {
+        QueryAbortWrapper {
+            valved: self.valve.wrap(stream),
+        }
+    }
+}
+
+#[pin_project(project = AbortWrapperProjection)]
+pub struct QueryAbortWrapper<S> {
```
ChristianBeilschmidt, line 78: Could you briefly describe what this is?

operators/src/engine/query.rs
```diff
@@ -43,37 +51,96 @@ impl TestDefault for ChunkByteSize {
...
+impl<S> Stream for QueryAbortWrapper<S>
+where
+    S: Stream,
+{
+    type Item = S::Item;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.project().valved.poll_next(cx)
+    }
+}
```
ChristianBeilschmidt, line 92: if we would use `AbortableStream` here, we would just need to `poll_next` on poll next and to call abort on "`valved`" when abort is called, right? We would store the `AbortHandle` as a second field in `QueryAbortWrapper`.
  Shouldn't we do this then? Couldn't we omit the registration then?

services/src/util/server.rs
```diff
@@ -4,18 +4,34 @@ use crate::util::config::get_config_element;
...
+use nix::sys::socket::MsgFlags;
+use std::any::Any;
+#[cfg(target_os = "linux")]
+use std::num::NonZeroI32;
 use std::num::NonZeroUsize;
+#[cfg(target_os = "linux")]
+use std::os::unix::prelude::{AsRawFd, RawFd};
+use std::time::Duration;
+#[cfg(target_os = "linux")]
+use std::time::Instant;
```
ChristianBeilschmidt, line 29: This is ugly to `cfg` that often.

You could…

 * put both into a module and one can just import "linux stuff".
 * import the stuff inside the `fn`

services/src/util/server.rs
```diff
@@ -294,3 +310,80 @@ pub(crate) fn log_server_info() -> Result<()> {
...
+            }
+        });
+
+        // TODO: return `pending` on JoinError?
+        let handle = handle.unwrap_or_else(|_| ());
+
+        Box::pin(handle)
+    } else if let Some(timeout) = timeout {
+        Box::pin(tokio::time::sleep(timeout))
+    } else {
```
ChristianBeilschmidt, line 372: if we have a timeout specified, we only wait for the timeout?

ChristianBeilschmidt (changes requested):

operators/src/engine/query.rs
michaelmattig: The `futures::stream::Abortable` doc says

> When abort is called on the handle tied to reg or if abort has already been called, the future/stream will complete immediately without making any further progress.

so if it completes immediately we have no chance to call abort on the other streams, right? Also, we still would have to have the `abort_handles` of all the other wrapped streams. And we can't collect them in the `QueryContext` since it can't be `mut`.

services/src/util/server.rs
michaelmattig: no, only if we don't finde a file descriptor for the current request we fall back to the timeout

services/src/util/server.rs
michaelmattig: 8e9bbd185e7e2a422a5f9110eed35702276aef52

operators/src/engine/query.rs
michaelmattig: 8e9bbd185e7e2a422a5f9110eed35702276aef52

operators/src/engine/query.rs
michaelmattig: 8e9bbd185e7e2a422a5f9110eed35702276aef52

services/src/util/server.rs
ChristianBeilschmidt: Is this reasonable to fall back to a timeout?

Is is not the case that:
 * if we set a timeout, we want to use that timeout
 * also, we want to cancel on drop-out
 
?

operators/src/engine/query.rs
ChristianBeilschmidt: 1. The thing about … being mutable is strange. How do you think you can add something to the current struct? It is also interior mutability.

2. If you create a stream wrapper. What is the difference in adding 
 a. an atomic bool to the stream that says: If stopped then return X; else poll_next  – and you share this bool for all streams in the workflow?
 b. using https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.take_until ?

I'm still curious if this could be simplified. Currently, you create two Streams for each stream, right?

michaelmattig: > The thing about … being mutable is strange. How do you think you can add something to the current struct? It is also interior mutability.

No, it works with a channel or something.

> a. an atomic bool to the stream that says: If stopped then return X; else poll_next – and you share this bool for all streams in the workflow?

doesn't this require locking on each poll?

> b. using https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.take_until ?

stream-cancel uses something very similar I think

> I'm still curious if this could be simplified. Currently, you create two Streams for each stream, right?

every processor creates a stream (logic) and I wrap it (to make it abortable). So it's one stream.

services/src/util/server.rs
michaelmattig: > Is this reasonable to fall back to a timeout?

well we discussed that we want to be able to specify a maximum time for requests. This method does both, abort on connection close or when the timeout is reached

> if we set a timeout, we want to use that timeout

the ogc protocols have no side-effects. It makes no sense to keep them running to a timeout if the connection is dropped anyway.

> 
    if we set a timeout, we want to use that timeout
    also, we want to cancel on drop-out


that's what I coded I think.

ChristianBeilschmidt: For me, it looks like if you get a file descriptor, you start a future that waits for that descriptor being closed.

If you don't get a file descriptor, you start a future that waits for the timeout if this is available.

If you have neither an fd, nor a timeout, you just wait forever in a third future

ChristianBeilschmidt:
I guess you have to merge

operators/src/engine/query.rs
michaelmattig: > I guess not, since its an atomic load: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html#method.load

ok it's a lock on the hardware level then

> Ah okay, so in principle we could use a single produce, multi receiver channel where the receiver closes on the first message.

yeah, stream-cancel uses something called a `TripWire`, built upon a `tokio::sync::watch` Channel. That's basically a Future on which multiple threads can poll at the same time. So we could set up something like this ourselves but we'd have to do that and test it. Which was already done in the stream-cancel trait.

> Okay, since valved (which also creates a stream) is re-used?

no it's not reused. All the valves only wait on the same Future in the `take_until`.

services/src/util/server.rs
michaelmattig: no, in line 348 you see that I also use the timeout when the file descriptor is being monitored. If no fd is present, fall back to the timeout. If no timeout is present, produce a future that never completes, thus the request will run until the query is finished. I think this is what we want.

ChristianBeilschmidt (approved):

PR #661, opened 2022-10-26
add query extensions and quota tracking

michaelmattig:

---

I have built a QuotaManager that gets informed via a channel when a calculation was made.
the other side of the channel is packed over a QuotaTracking object into the QueryContext and brought so into the StreamStatisticsAdapter.
For this I have built an extension object into the QueryContext, which is basically a typemap.
Unfortunately, there is no other way to extend the QueryContext.


michaelmattig:
[Verbrauchte Chunks erfassen](https://trello.com/c/WbiGYE6S/1068-verbrauchte-chunks-erfassen)

operators/src/engine/query.rs
```diff
@@ -43,18 +47,46 @@ impl TestDefault for ChunkByteSize {
 pub trait QueryContext: Send + Sync {
     fn chunk_byte_size(&self) -> ChunkByteSize;
     fn thread_pool(&self) -> &Arc<ThreadPool>;
+
+    fn extensions(&self) -> &QueryContextExtensions;
```
ChristianBeilschmidt, line 51: Please add some doc-comment

operators/src/engine/query.rs
```diff
@@ -43,18 +47,46 @@ impl TestDefault for ChunkByteSize {
...
+}
+
+#[derive(Default)]
+pub struct QueryContextExtensions {
```
ChristianBeilschmidt, line 55: Please add some doc-comment

operators/src/pro/meta/quota.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// An Id for a computation used for quota tracking
+#[derive(Debug, Clone, Copy)]
+pub struct ComputationUnit {
+    pub issuer: Uuid,                // TODO: use UserId?
+    pub context: ComputationContext, // TODO: introduce the concept of workflows to the operators crate and use/add it here
+}
+
+#[derive(Clone)]
+pub struct QuotaTracking {
```
ChristianBeilschmidt, line 15: please add a short doc-comment

operators/src/pro/meta/quota.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    pub fn work_unit_done(&self) -> impl Future<Output = ()> {
+        let mut sender = self.quota_sender.clone();
+        let computation = self.computation;
+        async move {
+            sender
+                .send(computation)
+                .await
+                .expect("the quota receiver should never close the receiving end of the channel");
```
ChristianBeilschmidt, line 35: Sure, but shouldn't we rather log an error message? Why should we panic in this case?

operators/src/pro/meta/statistics.rs
```diff
@@ -138,6 +139,12 @@ where
         query: QueryRectangle<Self::SpatialBounds>,
         ctx: &'a dyn QueryContext,
     ) -> Result<BoxStream<'a, Result<Self::Output>>> {
+        let quota = ctx
+            .extensions()
+            .get::<QuotaTracking>()
+            .expect("`QuotaTracking` extension should be set during `ProContext` creation")
```
ChristianBeilschmidt, line 145: Fingers crossed. I guess we can do this :thinking: … is this "the way" to use a type map in general?

services/src/pro/contexts/postgres.rs
```diff
@@ -461,7 +475,7 @@ where
                             email character varying (256),
                             real_name character varying (256),
                             active boolean NOT NULL
-                        );
+                        );
```
ChristianBeilschmidt, line 478:
```suggestion
                        );
```

Sure, let's just add some spaces for the stats :stuck_out_tongue:

services/src/pro/handlers/users.rs
```diff
@@ -30,6 +36,8 @@ where
                 .route(web::post().to(session_project_handler::<C>)),
         )
         .service(web::resource("/session/view").route(web::post().to(session_view_handler::<C>)))
+        .service(web::resource("/quota").route(web::get().to(quota_handler::<C>)))
+        .service(web::resource("/quotas/{user}").route(web::get().to(user_quota_handler::<C>)))
```
ChristianBeilschmidt, line 40: Shouldn't it be singular? We discussed that at some point.

services/src/pro/handlers/users.rs
```diff
@@ -243,6 +251,73 @@ pub(crate) async fn session_view_handler<C: ProContext>(
...
+        ("session_token" = [])
+    )
+)]
+pub(crate) async fn quota_handler<C: ProContext>(
+    ctx: web::Data<C>,
+    session: C::Session,
+) -> Result<impl Responder> {
+    let quota_used = ctx.user_db_ref().quota_used_by_session(&session).await?;
+    Ok(web::Json(QuotaUsed { used: quota_used }))
+}
```
ChristianBeilschmidt, line 281: Coudln't we omit this API if the user knows his id?

services/src/pro/quota/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+use std::sync::Arc;
+
+use geoengine_datatypes::util::test::TestDefault;
+
+use crate::pro::users::UserId;
+
+use super::users::{UserDb, UserSession};
+
+// TODO: find a good default, choose w.r.t. resources or make configurable
+const QUOTA_CHANNEL_SIZE: usize = 1000;
```
ChristianBeilschmidt, line 15: What happens if this channel is bounded? The Futures were spawned in the Stream adapter… and… then?

I'm not sure what makes sense here.

services/src/pro/quota/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+        Self {
+            user_db,
+            quota_receiver,
+        }
+    }
+
+    pub fn run(mut self) {
+        crate::util::spawn(async move {
+            while let Some(computation) = self.quota_receiver.next().await {
+                // TODO: issue a tracing event instead?
```
ChristianBeilschmidt, line 65: Yeah, this will definitely blow our log.

services/src/pro/quota/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+                //       However, currently it is possible to reuse a context for multiple queries.
+                //       Also: the operators crate knows nothing about workflows as of yet.
+                log::info!(
+                    "Quota received. User: {}, Context: {}",
+                    computation.issuer,
+                    computation.context
+                );
+
+                let user = UserId(computation.issuer);
+                // TODO: what to do if this fails (quota can't be recorded)? Try again later?
```
ChristianBeilschmidt, line 76: I guess this is okay for now.

services/src/pro/quota/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+                let r = self.user_db.increment_quota_used(&user, 1).await;
```
ChristianBeilschmidt, line 77: It would be nice to batch this I guess. Maybe we can fetch *n* messages at once?

services/src/pro/users/postgres_userdb.rs
```diff
@@ -432,4 +432,43 @@ where
 
         Ok(())
     }
+
+    async fn increment_quota_used(&self, user: &UserId, quota_used: u64) -> Result<()> {
+        let conn = self.conn_pool.get().await?;
+        let stmt = conn
+            .prepare("UPDATE users SET quota_used = quota_used + $1 WHERE id = $2;")
+            .await?;
```
ChristianBeilschmidt, line 440: Can we keep the prepared message? It is somehow strange that we prepare it X times. Or is there a cache from the DB side?

services/src/pro/users/userdb.rs
```diff
@@ -77,4 +77,29 @@ pub trait UserDb: Send + Sync {
...
+    ///
+    async fn quota_used_by_session(&self, session: &UserSession) -> Result<u64>;
+
+    /// Gets a specific users used quota
+    ///
+    /// # Errors
+    ///
+    /// This call fails if the session is invalid
+    ///
+    async fn quota_used_by_user(&self, user: &UserId) -> Result<u64>;
```
ChristianBeilschmidt, line 104: Do we need both? Or could we derive one from another? E.g. the user id from the session

jdroenner:
can't we have a QueryContextExt or something like that? And then have Box<QueryContextExt> impl QueryContext?

operators/src/pro/meta/quota.rs
michaelmattig: It is an error that can never happen, that's why we do not need to log anything. The reason why it can never happen is given in the expect message.

operators/src/pro/meta/statistics.rs
michaelmattig: if you have any suggestions for how to pass the `Sender` to the quota channel let me know. This idea is inspired by the `Extensions` in actix.

operators/src/engine/query.rs
michaelmattig: 23e1d9ebc662381de5ae75709bbb725e618fa97b

operators/src/engine/query.rs
michaelmattig: 23e1d9ebc662381de5ae75709bbb725e618fa97b

operators/src/pro/adapters/stream_statistics_adapter.rs
```diff
@@ -82,16 +88,30 @@ where
...
+        let (tx, _rx) = channel::<ComputationUnit>(111);
+        let quota = QuotaTracking::new(
+            tx,
+            ComputationUnit {
+                issuer: Uuid::new_v4(),
+                context: ComputationContext::new(),
+            },
+        );
+        let mut v_stat_stream =
+            StreamStatisticsAdapter::new(v_stream, span!(Level::TRACE, "test"), quota);
```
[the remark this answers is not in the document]
michaelmattig: 23e1d9ebc662381de5ae75709bbb725e618fa97b

operators/src/pro/meta/quota.rs
michaelmattig: 23e1d9ebc662381de5ae75709bbb725e618fa97b

services/src/pro/contexts/postgres.rs
michaelmattig: 23e1d9ebc662381de5ae75709bbb725e618fa97b

services/src/pro/handlers/users.rs
michaelmattig: yes but it would be inconsistent with all the other APIs. We don't require the user id when getting the users project or something like that. As I see it, this API is for the user, and the other one is for an admin.

services/src/pro/quota/mod.rs
michaelmattig: right now we log all the tracing events as well. The todo was meant on a semantic level.

services/src/pro/quota/mod.rs
michaelmattig: ok

services/src/pro/quota/mod.rs
michaelmattig: we would need to group the stuff by users though. I can add a TODO for batching, but I think it's pretty dangerous because this stuff will get lost if the application crashes or exits.

michaelmattig:
> can't we have a QueryContextExt or something like that? And then have Box < QueryContextEx t> impl QueryContext?

I implemented the `QueryContextExtensions`  to allow adding data when Geo Engine Pro is used. We can't put a different `QueryContext` into the query method based on pro/free. We can't make the `QueryContext` a generic, because we use the processor as a trait object.

So your idea would be to implement  `QueryContextExt` only when pro is used? Could be possible, but we don't have a `Box<QueryContext>`, but a `&dyn QueryContext`, so I think we would first have to change that.

Still: how would you implement the `QueryContextExt` on a `Box<QueryContext>`? Where would you store the additional data?

jdroenner:
So my idea was to have QueryContext as a generic `&C` then we can set a different C in the pro and the normal version. We can require C: QueryContext always. And in the pro we provide a  QueryContextExt which implements QueryContext and has more methods to get additional stuff. This should work but i have not tested it.

michaelmattig:
> So my idea was to have QueryContext as a generic `&C` then we can set a different C in the pro and the normal version. We can require C: QueryContext always. And in the pro we provide a QueryContextExt which implements QueryContext and has more methods to get additional stuff. This should work but i have not tested it.

the QueryContext can't be generic because we use the QueryProcessors as trait objects, and you can't create objects from traits with genericfunctions.

jdroenner:
it should be fine to pass a QueryContextExt as &dyn QueryContext to an operator since it implements QueryContext

michaelmattig:
> it should be fine to pass a QueryContextExt as &dyn QueryContext to an operator since it implements QueryContext

if we pass if as &QueryContext we can't access the methods of QueryContextExt. There is no possibility to downcast references AFAIK

jdroenner:
if it does not work we can also add a as_query_context() to the QueryContextExt and call that when needed? As i said i did not test it. I just think that should work. Maybe Chris also has an idea.

operators/src/engine/execution_context.rs
```diff
@@ -131,6 +131,7 @@ impl MockExecutionContext {
         MockQueryContext {
             chunk_byte_size,
             thread_pool: self.thread_pool.clone(),
+            extensions: Default::default(),
             abort_registration,
             abort_trigger: Some(abort_trigger),
         }
```
ChristianBeilschmidt, line 137: I get the following errors all the time:

> OpenTelemetry trace error occurred. Exporter jaeger encountered the following error(s): thrift agent failed with not open

Yes, I don't have Jeager running. But there shouldn't be this error, right?

ChristianBeilschmidt, line 137: I copied the session token from the UI to swagger-UI and /session works.

But /quota gives the following error:

grafik

Issuing this directly works:

```
curl -X 'GET' \
  'http://127.0.0.1:3030/quota' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer 8ef8f6e9-f465-4eba-b743-ff732833dfe5'
``

**Review by ChristianBeilschmidt — CHANGES_REQUESTED** [review]
I issued some requests in the UI and got 

```
{
  "used": 3263
}
```

to it! Very nice :+1: .

File: operators/src/pro/meta/quota.rs (hunk 4 above)
Comment (michaelmattig) [reply operators/src/pro/meta/quota.rs:37]: clippy wants to handle all results. we cold also do `let _ = sender.send` I guess.

File: operators/src/pro/meta/statistics.rs (hunk 5 above)
Comment (michaelmattig) [reply operators/src/pro/meta/statistics.rs:145]: the thing is, it is not always there. It's only there if we are in Pro.

File: services/src/pro/handlers/users.rs (hunk 8 above)
Comment (michaelmattig) [reply services/src/pro/handlers/users.rs:281]: then I would omit the `/qoutas/{userId}` one because for me it doesn't belong in the user API but some admin API which we do not have yet.

File: services/src/pro/quota/mod.rs (hunk 9 above)
Comment (michaelmattig) [reply services/src/pro/quota/mod.rs:15]: > Do we want to block our computations when our database or our implementation of logging is not fast enough?

right now, we wouldn't. Only the task we spawn would block. Computation continues.

> Do we run into the issue of having 4000 futures that use memory and block because the channel is full?

Thinking of all the stuff that is done in order to produce a chunk, I doubt that we run into the problem of having to many quota futures.

> Is it more reasonable to limit the channel to some number or to have it unlimited if we have no strategy to cope with resolving it?

I thought an unbounded channel had some limitations? At least the std::sync::Sender was not Sync but the SyncSender was and that one was bounded. If we can interchange the tokio channel with an unbounded one I would say we should do it.

File: services/src/pro/quota/mod.rs (hunk 10 above)
Comment (michaelmattig) [reply services/src/pro/quota/mod.rs:65]: 56ad2af66b5e7c59ad180b6b15f2405e6fc39e73

File: services/src/pro/quota/mod.rs (hunk 12 above)
Comment (michaelmattig) [reply services/src/pro/quota/mod.rs:77]: > I guess we need to measure some concurrent workflows, because there is a blocking if this is not fast enough.

it can't ever block because quota management and workflow processing is independent of each other. The query processor does not wait for the quota to be recorderd.

File: services/src/pro/users/postgres_userdb.rs (hunk 13 above)
Comment (michaelmattig) [reply services/src/pro/users/postgres_userdb.rs:440]: I don't understand the problem. What would the alternative here, string concatenation?

> Prepared statements can only be used with the connection that created them.

We are using a connection pool so if we wanted to cache them we would need one cache per connection. I don't think this is a) feasible, b) a problem and c) part of this user story.

File: services/src/pro/users/userdb.rs (hunk 14 above)
Comment (michaelmattig) [reply services/src/pro/users/userdb.rs:104]: suggest a different name

File: operators/src/pro/meta/quota.rs [hunk 17]
```diff
@@ -5,12 +5,14 @@ use uuid::Uuid;
...
 /// An Id for a computation used for quota tracking
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct ComputationUnit {
     pub issuer: Uuid,                // TODO: use UserId?
     pub context: ComputationContext, // TODO: introduce the concept of workflows to the operators crate and use/add it here
 }
 
+/// This type holds a `Sender` to a channel that is used to track the computation units.
+/// It is passed to the `StreamStatisticsAdapter` via the `QueryContext`.
```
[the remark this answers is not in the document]
michaelmattig: does it work with external types? Please test it

services/test.http
```diff
@@ -296,5 +296,5 @@ curl 'http://localhost:3030/workflow' -X POST -H 'User-Agent: Mozilla/5.0 (Windo
...
--H 'Authorization: Bearer {{anonymousSession.response.body.$.id}}' -H 'Connection: keep-alive' -H 'Referer: http://localhost:3030/' -H 'Cookie: default-theme=ngax' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-origin' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache'
\ No newline at end of file
+curl 'http://localhost:4200/api/wfs/27ea8d72-ba91-523b-88f6-92f2a76bb58b?service=WFS&version=2.0.0&request=GetFeature&outputFormat=application/json&typeNames=27ea8d72-ba91-523b-88f6-92f2a76bb58b&bbox=-20037508.342789244,-9509989.31112849,20037508.342789244,17571955.5584226&time=2014-04-01T12:00:00.000Z/2014-04-01T12:00:00.000Z&srsName=EPSG:3857&queryResolution=39135.75848201024' -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:106.0) Gecko/20100101 Firefox/106.0' -H 'Accept: application/json, text/plain, */*' -H 'Accept-Language: de,en-US;q=0.7,en;q=0.3' -H 'Accept-Encoding: gzip, deflate, br'
```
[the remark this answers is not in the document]
michaelmattig: I use it for debugging and testing. I initially didn't commit my changes but as I often change between different branches I didn't want to lose my test queries. I can revert the changes to the file if you want.

operators/src/engine/execution_context.rs
michaelmattig: You should disable the opentelemetry tracing in the Settings.toml if you don't want to use it.

michaelmattig: is there anything printed in the log of geo engine?

operators/src/pro/meta/statistics.rs
ChristianBeilschmidt: Sure, but you are in a ProContext right? It could have a getter for that.

services/src/pro/handlers/users.rs
ChristianBeilschmidt: But then we lose functionality, right?

services/src/pro/quota/mod.rs
ChristianBeilschmidt: 1. You are right that the future just spawns into the wild!

2. 
https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html

It has 
```
[Functions](https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html#functions)
[channel](https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html)
Creates a bounded mpsc channel for communicating between asynchronous tasks with backpressure.
[unbounded_channel](https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.unbounded_channel.html)
Creates an unbounded mpsc channel for communicating between asynchronous tasks without backpressure.
```

I'm not entirely against a bounded channel, I'm just curious what the implications might be.

services/src/pro/users/userdb.rs
ChristianBeilschmidt: Since `UserSession` contains the field `id: UserId`, we can omit `quota_used_by_session` and call it whenever we have either a `UserId` or a `UserSession`, right?

operators/src/pro/meta/quota.rs
michaelmattig: It is an error that can never happen, that's why we do not need to log anything.

operators/src/pro/meta/statistics.rs
michaelmattig: We have the same `&dyn QueryContext` in free and in pro. We can't make that generic because we use the query processor as a trait object and trait objects can't have generic functions.

services/src/pro/handlers/users.rs
michaelmattig: yes. I'd vote for keeping both methods and move the `/qoutas/{userId} ` into an admin API once we have one. We can also have a vote in the dev team which method to keep.

services/src/pro/quota/mod.rs
michaelmattig: I think there are no implications because updating the quota in the database will always be faster than creating new tiles. So the channel should never fill up.

operators/src/pro/meta/quota.rs
ChristianBeilschmidt: In the style of defensive programm: then just do nothing?

let _ = …

services/src/pro/quota/mod.rs
ChristianBeilschmidt: Okay, so… make it unbounded then?

Btw: I'm not 100% sure that creating a tile is always slower then accessing an external database, starting a transaction, updating a record and submitting the transaction.

services/src/pro/handlers/users.rs
ChristianBeilschmidt: But your admin API is hypothetical. There could also be different approaches. For instance, what if rights management allows project leaders to query the quotas of other team members? Should this be part of the admin API then?

operators/src/engine/execution_context.rs
ChristianBeilschmidt: grafik

The backend seems to get the request
> 2022-11-04T10:20:21.380003Z  INFO actix_web::middleware::logger: 127.0.0.1 "POST /anonymous HTTP/1.1" 200 319 "http://localhost:3030/" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:106.0) Gecko/20100101 Firefox/106.0" 0.000434

ChristianBeilschmidt: https://trello.com/c/OesNSvEg/1080-cors-f%C3%BCr-swagger-ui-in-der-entwicklung-fixen

services/src/pro/users/userdb.rs
ChristianBeilschmidt: Well I don't know. If your are so hesitant in keeping stuff that we not necessarily need now, because of some admin API that might or might not come, then keep it.

But then at least name it appropriately and update the docs.

operators/src/pro/meta/statistics.rs
ChristianBeilschmidt: You have to explain that to me why this is not possible. We also have other methods only in the ProContext, right?

operators/src/pro/meta/quota.rs
michaelmattig: what's better

```
let _ = bla(); // ignore result because it can never fail
```

or 

```
bla().expect("never fails because...")
```

operators/src/pro/meta/statistics.rs
michaelmattig: If the extensions are optional I would need to unwrap two times. One for the extensions and one for the thing I actually need.

What do you mean by parametrizing the QueryContext?

services/src/pro/handlers/users.rs
michaelmattig: Shall I create a separate QuotaDb maybe for a) incrementing quota and b) accessing arbitrary user's quota? Then accessing your own quota goes through the UserDb and the rest to QuotaDb and we have no doubled functions.

services/src/pro/users/userdb.rs
michaelmattig: cf my other comment regardign a QuotaDb

operators/src/engine/execution_context.rs
michaelmattig: ok yes it may be a problem then with 127.0.0.1 vs localhost.

services/src/pro/quota/mod.rs
michaelmattig: > Btw: I'm not 100% sure that creating a tile is always slower then accessing an external database, starting a transaction, updating a record and submitting the transaction.

it doesn't have to be always slower, just on average, so that the bottleneck is actually creating the tiles and not updating the quota in the database. I changed it to an unbounded channel now, so now the bottleneck is our ram which may be filled up with pending quota messages that are not yet written to the database.

3715bcf54d31857d07e77e3b748076cf72108b6f

operators/src/pro/meta/statistics.rs
ChristianBeilschmidt: I thought you would just return an error here instead of panicking and you are done.

services/src/pro/handlers/users.rs
ChristianBeilschmidt: Nah, don't let us introduce some new type of DB where we do not know whether we need it.

It makes no difference now if we have this method in here on in another DB.
So just imagine you are in the QuotaDb ;-).

I made my point regarding redundancy. It is not wrong, so just decide what you want to go with.

operators/src/pro/meta/statistics.rs
michaelmattig: it's an error that can never happen, so why return it?

services/src/pro/handlers/users.rs
michaelmattig: I'd go with the "redundancy" because to me it's two entirely different things, accessing your own quota and accessing other people's quota.

operators/src/pro/meta/statistics.rs
ChristianBeilschmidt: Does having a QueryContext as a generic associated type help here?

services/src/pro/users/userdb.rs
ChristianBeilschmidt: Okay, then just rename it

michaelmattig: rename it to what?

ChristianBeilschmidt: I proposed my idea about this 5 days ago.

michaelmattig: f8a8b3d6f76bad8778c460f2c38f42646f6cb1b8

ChristianBeilschmidt (dismissed):

ChristianBeilschmidt:
Please add a changelog entry (cf. checkbox above).

For me, it is okay then.
I don't know if @jdroenner needs his thumbs up because you starter a discussion here?

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #664, opened 2022-10-31
warn not fail on gdal pixel size missmatch

jdroenner:

---

changed the assert in gdal source where dataset pixel size and gdal pixelsize are compared into a warning. Somehow this is not always equal even if the number in the dataset is the number from gdalinfo.


jdroenner:
[GdalSource pixel size führt zu fehlern](https://trello.com/c/O2rRZork/1074-gdalsource-pixel-size-f%C3%BChrt-zu-fehlern)

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #674, opened 2022-11-11
Fix WCS output

michaelmattig:

---

<TEXT>


michaelmattig:
[WCS Tiff hat Linien drin](https://trello.com/c/wZkQgMDv/1047-wcs-tiff-hat-linien-drin)

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -150,15 +155,15 @@ struct GdalDatasetWriter<P: Pixel + GdalType> {
...
-    y_pixel_size: f64,
     _type: std::marker::PhantomData<P>,
     use_big_tiff: bool,
+    global_offset: GridIdx2D,
+    global_limit: GridIdx2D,
 }
 
 impl<P: Pixel + GdalType> GdalDatasetWriter<P> {
     fn new(
+        tiling_specification: TilingSpecification,
```
jdroenner, line 166: to keep the write more easy to reuse please add the offset and limit as prams to the `new` method and add a `new_with_tiling_spec` or something where they are calculated (and then call `new`).

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -150,15 +155,15 @@ struct GdalDatasetWriter<P: Pixel + GdalType> {
     gdal_tiff_options: GdalGeoTiffOptions,
     gdal_tiff_metadata: GdalGeoTiffDatasetMetadata,
     output_bounds: SpatialPartition2D,
-    output_geo_transform: GeoTransform,
-    x_pixel_size: f64,
-    y_pixel_size: f64,
     _type: std::marker::PhantomData<P>,
     use_big_tiff: bool,
+    global_offset: GridIdx2D,
```
jdroenner, line 160: maybe call this `window_start` and the other `window_end`? Also do you think we could already put them into a window struct like it is in the gdal source? Then it is more easy when we switch all things to pixels boxes.

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -183,6 +188,18 @@ impl<P: Pixel + GdalType> GdalDatasetWriter<P> {
         );
         let output_bounds = query_rect.spatial_bounds;
 
+        let tile_info = tiling_specification
```
jdroenner, line 191: do you only need the geotransform?  you can get that directly from the TilingStrategy. You could also add a method to the TilingSpec and reuse that also in createing the TilingStrategy.

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -248,43 +264,62 @@ impl<P: Pixel + GdalType> GdalDatasetWriter<P> {
 
         let tile_bounds = tile_info.spatial_partition();
 
-        let (upper_left, grid_array) = if self.output_bounds.contains(&tile_bounds) {
+        // compute the upper left pixel index in the output raster and extract the input data
+        let (output_ul_idx, grid_array) = if self.output_bounds.contains(&tile_bounds) {
```
jdroenner, line 268: we could also do this using pixels right? We know the tile grid bounds from the TilingStrategy (+query rect) and each tile has a GridIdx of the tiling grid.

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -248,43 +264,62 @@ impl<P: Pixel + GdalType> GdalDatasetWriter<P> {
...
+            // TODO: compute the intersection on the `SpatialPartition2D`s once the float precision issue is fixed
+
+            let tile_offset = tile_info.global_upper_left_pixel_idx();
+            let tile_limit = tile_offset
+                + GridIdx2D::from([
+                    tile_info.tile_size_in_pixels.axis_size_y() as isize,
+                    tile_info.tile_size_in_pixels.axis_size_x() as isize,
+                ]);
+
+            if tile_limit.inner()[0] < self.global_offset.inner()[0]
```
jdroenner, line 285: you can desstructure it with `GridIdx([idx_y, idx_x]) = ...` to make it more easy to unterstand.

jdroenner (changes requested):

operators/src/util/raster_stream_to_geotiff.rs
michaelmattig: I calculate the `contains` on the pixels now b0d17f67c2c79b8581d6ed3c9d6c4b2cd66fa10e

operators/src/util/raster_stream_to_geotiff.rs
michaelmattig: I destructured all the things b0d17f67c2c79b8581d6ed3c9d6c4b2cd66fa10e

operators/src/util/raster_stream_to_geotiff.rs
jdroenner: I was thinking about a struct `RasterWindow { ul: GridIdx2D, lr: GridIdx2D}`

jdroenner:
bors merge

jdroenner (approved):

jdroenner:
bors merge

PR #675, opened 2022-11-14
Raster Workflow to Dataset

glombiewski:
- [x] I added an entry to [`CHANGELOG.md`](CHANGELOG.md) if knowledge of this change could be valuable to users.

---

- Support for multiple time steps in raster workflow export
- Support for Task API


glombiewski:
[[RESPECT] 9.6 Persistieren von Workflow-Berechnungen (a)](https://trello.com/c/3x4agKAG/904-respect-96-persistieren-von-workflow-berechnungen-a)

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -48,11 +53,18 @@ where
...
-    .await?;
-
-    let bytes = gdal::vsi::get_vsi_mem_file_bytes_owned(file_path)?;
+    .await?
+    .into_iter()
+    .map(|x| {
+        gdal::vsi::get_vsi_mem_file_bytes_owned(
+            x.params
+                .expect("params is always set in raster_stream_to_geotiff")
+                .file_path,
```
ChristianBeilschmidt, line 62: Why don't you use `file_path` from line 44 anymore?

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -118,9 +127,47 @@ where
...
+                    dataset_writer.time_intervals.push(current_interval);
+                    let mut next_file_path: String =
+                        dataset_writer.placeholder_path.to_string_lossy().into();
+
+                    //TODO: Consider making time_string format and place_holder configurable
+                    let time_string = current_interval
+                        .start()
+                        .as_date_time()
+                        .ok_or(Error::TimeInstanceNotDisplayable)?
+                        .format(&DateTimeParseFormat::custom("%Y-%m-%d".to_string()));
```
ChristianBeilschmidt, line 151: I guess without configuration is fine.
But ISO timestamps with date and time would cover most use cases, right?

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -227,20 +339,27 @@ impl<P: Pixel + GdalType> GdalDatasetWriter<P> {
...
+            self.rasterband_index,
+            self.use_big_tiff,
+            self.output_geo_transform,
+        )?);
+
+        Ok(())
+    }
+
+    fn reset_dataset(&mut self) -> Option<Dataset> {
+        std::mem::replace(&mut self.dataset, None)
```
ChristianBeilschmidt, line 362: Can't you replace that with `self.dataset.take()`?

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -344,22 +469,35 @@ impl<P: Pixel + GdalType> GdalDatasetWriter<P> {
...
+        let input_dataset = self.reset_dataset().expect("dataset should be initialized");
+        let output_file_path = self
+            .result
+            .last()
+            .expect("finish should only be called after a temporal slice finished")
+            .params
+            .as_ref()
+            .expect("params should be set for a finished temporal slice")
+            .file_path
+            .clone();
```
ChristianBeilschmidt, line 482: Can we do anything to not have to unwrap 3x in this statement?

This somehow sounds like a state machine, because you rely on state transitions :sweat_smile: .
Maybe this can be done by inserting 2 instead of one type? :thinking:

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -798,11 +944,88 @@ mod tests {
...
+        // for i in 0..3 {
+        //     let file_name = format!(
+        //         "../test_data/raster/cloud_optimized_geotiff_timestep_{}_from_stream_compressed.tiff",
+        //         i
+        //     );
+        //     geoengine_datatypes::util::test::save_test_bytes(
+        //         &bytes.pop().expect("bytes should have length 3"),
+        //         file_name.as_str(),
+        //     );
+        // }
```
ChristianBeilschmidt, line 1016:
```suggestion
```

services/src/handlers/wcs.rs
```diff
@@ -440,7 +440,11 @@ async fn wcs_get_coverage_handler<C: Context>(
         .await)?
     .map_err(error::Error::from)?;
 
-    Ok(HttpResponse::Ok().content_type("image/tiff").body(bytes))
+    assert_eq!(bytes.len(), 1);
```
ChristianBeilschmidt, line 443: I think we should not crash the program here. Maybe just return an error instead.

Alternatively, replace it with a debug_assert :shrug:

services/src/handlers/workflows.rs
```diff
@@ -63,6 +66,10 @@ where
     .service(
         web::resource("datasetFromWorkflow/{id}")
             .route(web::post().to(dataset_from_workflow_handler::<C>)),
+    )
+    .service(
+        web::resource("datasetFromWorkflowTask/{id}")
+            .route(web::post().to(dataset_from_workflow_handler_task::<C>)),
```
ChristianBeilschmidt, line 72: Is it useful to have both methods?

Actually, we should only use a task for that, right?

services/src/handlers/workflows.rs
```diff
@@ -523,9 +477,167 @@ async fn create_dataset<C: Context>(
...
+example = json!({"task_id": "7f8a4cfe-76ab-4972-b347-b197e5ef0f3c"})
+)
+),
+params(
+("id" = WorkflowId, description = "Workflow id")
+),
+security(
+("session_token" = [])
+)
+)]
```
ChristianBeilschmidt, line 610:
```suggestion
#[utoipa::path(
    tag = "Workflows",
    post,
    path = "/datasetFromWorkflowTask/{id}",
    request_body = RasterDatasetFromWorkflow,
    responses(
        (status = 200, description = "Id of created task", body = TaskResponse,
        example = json!({"task_id": "7f8a4cfe-76ab-4972-b347-b197e5ef0f3c"})
        )
    ),
    params(
        ("id" = WorkflowId, description = "Workflow id")
    ),
    security(
        ("session_token" = [])
    )
)]
```

There seems to be no auto formatted, but maybe we can do it accordingly.

ChristianBeilschmidt (changes requested):

ChristianBeilschmidt:
I guess you could also add a changelog entry

operators/src/util/raster_stream_to_geotiff.rs
glombiewski: line 44 is a prefix (the folder where the files go into). Each timerange is a separate file, so it has its own file name.

operators/src/util/raster_stream_to_geotiff.rs
glombiewski: b543a653315680ecef95c34cea64c343fe2f1fc6

Yes.

operators/src/util/raster_stream_to_geotiff.rs
glombiewski: I believe in other tests those kinds of lines are kept (e.g., cloud_optimized_geotiff_from_stream), because you might need to regenerate those files if the gdal version changes.

services/src/handlers/wcs.rs
glombiewski: I added a convenience function for single timesteps and a InvalidNumberOfTimeSteps error if it is not a single timestep. Removes redundant code in tests and allows a map to service error here, so I feel such a convenience function is justified. 8b7769d2a5ca85f0dd06a8268905e3a3f94c19cf

services/src/handlers/workflows.rs
glombiewski: I removed it. f0fbc323499f3052b7b1663076599f7df685e88e

operators/src/util/raster_stream_to_geotiff.rs
glombiewski: I did some refactoring for this to reduce lines of code for the method. 358486707b7e9cd6b2d3017c9e19b4c69df97de1
I added the time, but : is problematic in some OS, isn't it? So I used "-" for times as well. But you probably have more experience with weird timestamps in filenames and what to do there. 42b7121d68de34d829c79c9d4d91d0f82f52a17c

operators/src/util/raster_stream_to_geotiff.rs
glombiewski: - Clippy complained about too many lines. Since I had to do large scale refactoring anyway, I went the extra mile with write_no_data_value_grid and write_masked_data_grid. 358486707b7e9cd6b2d3017c9e19b4c69df97de1
- I removed the error::Error::EmptyInput again, since you said there is no way for raster queries to not have tiles.

Please note that 358486707b7e9cd6b2d3017c9e19b4c69df97de1 caused merge conflicts with the tile specification thing, so that needed to be untangled. 1190fa5b60e875a869de8b6b77ae2ac7b2b57493 a012c83feed0560f3f9eaf979b3e9f867bfc5932 5992963681eab30920b7d725d9646aed2f3a8453 95e78dd4f4931722d10ab8ebe521e50f2afd4823

operators/src/util/raster_stream_to_geotiff.rs
```diff
@@ -53,15 +55,57 @@ where
...
+        tile_limit,
+        conn_closed,
+        tiling_specification,
+    )
+    .await?
+    .into_iter()
+    .map(|x| {
+        gdal::vsi::get_vsi_mem_file_bytes_owned(
+            x.params
+                .expect("params is always set in raster_stream_to_geotiff")
```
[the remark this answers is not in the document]
glombiewski: be26049d69aa5e5b30df269083c554ad1763b180

I don't believe this can occur at the moment, but for now I decided to use an empty vector to represent a gap. There are other possibilities (skip gaps, create an empty tiff and read that in, ...), but I did not get a definitive answer in Slack, so I took the empty vector route.

services/src/handlers/workflows.rs
```diff
@@ -379,70 +382,99 @@ pub struct RasterDatasetFromWorkflowResult {
...
+    session: <C as Context>::Session,
+) -> Result<DatasetId> {
+    let dataset_id = DatasetId::new();
+    //TODO: Recognize MetaDataDefinition::GdalMetaDataRegular
+    let meta_data = if slice_info.len() == 1 {
+        let loading_info_slice = slice_info.pop().expect("slice_info has len one");
+        let time = Some(loading_info_slice.time);
+        let params = loading_info_slice
+            .params
+            .expect("params is always set in raster_stream_to_geotiff"); //TODO: What is the use-case for None in the GdalMetaDataList?
```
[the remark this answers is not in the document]
glombiewski: be26049d69aa5e5b30df269083c554ad1763b180

I changed the expect message, but the expect should be valid for this case, because a static dataset should not consist of just a gap.

ChristianBeilschmidt:
bors merge

PR #683, opened 2022-11-17
Add rasterization operator

dbrandenstein:
- [x] I added an entry to [`CHANGELOG.md`](CHANGELOG.md) if knowledge of this change could be valuable to users.

---

Implement a rasterization operator which rasterizes points into a grid or density representation.


dbrandenstein:
[[GFBio] Rasterization für Punktdaten (Heatmap)](https://trello.com/c/ajTGM5DC/994-gfbio-rasterization-f%C3%BCr-punktdaten-heatmap)

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+
+                    let grid_geo_transform = GeoTransform::new(
+                        grid_spatial_bounds.upper_left(),
+                        grid_resolution.x,
+                        -grid_resolution.y,
+                    );
+
+                    let mut grid_data;
+
+                    if let MultiPoint(processor) = &self.input {
```
ChristianBeilschmidt, line 271: This confused me at first.
Couldn't you let the `if let` block return the grid_data?
The else block returns anyway.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                ),
+            },
+        };
+
+        let tiling_strategy = self
+            .tiling_specification
+            .strategy(query.spatial_resolution.x, -query.spatial_resolution.y);
+        let tile_shape = tiling_strategy.tile_size_in_pixels;
+
+        let tiles = ctx.thread_pool().install(|| {
```
ChristianBeilschmidt, line 245: Do you do anything parallel in here?

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                tiling_specification: self.tiling_specification,
+                radius: self.radius,
+                stddev: self.stddev,
+            }
+            .boxed(),
+        ))
+    }
+}
+
+pub struct GridRasterizationQueryProcessor {
```
ChristianBeilschmidt, line 206: Maybe you can write a short doc-comment what this does.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                                                coord
+                                                    - Coordinate2D::new(self.radius, -self.radius),
+                                                coord
+                                                    + Coordinate2D::new(self.radius, -self.radius),
+                                            )
+                                            .expect("Bounds should be valid since radius was checked to be strictly positive")
+                                            .intersection(&tile_info.spatial_partition());
+
+                                            let grid_bounds = match spatial_partition {
+                                                None => { return None; },
```
ChristianBeilschmidt, line 392: Maybe add a comment what happens here and why you can return `None`

ChristianBeilschmidt (changes requested):

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                            .into(),
+                    ),
+                ))
+            });
+
+        Ok(tiles.boxed())
+    }
+}
+
+fn gaussian(x: f64, stddev: f64) -> f64 {
```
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/e5da1695b25feb3f71272a1a504d7ba438abd053

operators/src/processing/rasterization/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/e5da1695b25feb3f71272a1a504d7ba438abd053

This should not strictly be necessary since I only query for points with distance of at most radius from the tile borders so the intersection should never be empty. However, I added this check to be safe when a vector source returns points further away (there is a comment in the OGR source that the spatial filter can be inaccurate, for example). In case the intersection is actually empty nothing needs to be computed so no output values are produced (that's why I return None in the filter_map call).

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                                        .coordinates()
+                                        .par_iter()
+                                        .filter_map(|&coord| {
+                                            let spatial_partition = SpatialPartition2D::new(
+                                                coord
+                                                    - Coordinate2D::new(self.radius, -self.radius),
+                                                coord
+                                                    + Coordinate2D::new(self.radius, -self.radius),
+                                            )
+                                            .expect("Bounds should be valid since radius was checked to be strictly positive")
```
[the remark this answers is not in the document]
dbrandenstein: Now handled in https://github.com/geo-engine/geoengine/pull/683/commits/fc14bc58725a86f03929166554ac05f9ccdc2905 by clipping to the f64 bounds.

I noticed this can also be a problem in SpatialPartition2D's snap_to_grid, for example. If an over-/underflow occurs there it will panic. Maybe this should also be fixed. I don't think quietly clipping to the f64 bounds would be the right solution there though.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+
+                if let MultiPoint(processor) = &self.input {
+                    let mut chunks = processor.query(vector_query, ctx).await?;
+
+                    tile_data = vec![0.; tile_size_x * tile_size_y];
+
+                    while let Some(chunk) = chunks.next().await {
+                        let chunk = chunk?;
+                        ctx.thread_pool().install(|| {
+                            rayon::scope(|s| {
```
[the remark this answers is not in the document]
dbrandenstein: The scopeless rayon spawn requires static lifetime of the chunks to be processed. The scope ensures nothing escapes the function body.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+#[serde(tag = "type")]
+pub enum GridOrDensity {
+    Grid(GridParams),
+    Density(DensityParams),
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
+pub struct DensityParams {
+    radius: f64,
+    stddev: f64,
```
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/e5da1695b25feb3f71272a1a504d7ba438abd053

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+#[serde(rename_all = "lowercase")]
+#[serde(tag = "type")]
+pub enum GridOrDensity {
+    Grid(GridParams),
+    Density(DensityParams),
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
+pub struct DensityParams {
+    radius: f64,
```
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/e5da1695b25feb3f71272a1a504d7ba438abd053

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+    stddev: f64,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GridParams {
+    spatial_resolution: SpatialResolution,
```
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/e5da1695b25feb3f71272a1a504d7ba438abd053

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+
+/// An operator that rasterizes vector data
+pub type Rasterization = Operator<RasterizationParams, SingleVectorSource>;
+
+impl OperatorName for Rasterization {
+    const TYPE_NAME: &'static str = "Rasterization";
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/a2ffdcdbdb57954a96a6425984753b80f1ab6ae2

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+pub enum GridSizeMode {
+    Fixed,
+    Relative,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/a2ffdcdbdb57954a96a6425984753b80f1ab6ae2

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                                                            x as usize + y as usize * tile_size_x,
+                                                            gaussian(distance, self.stddev),
+                                                        ))
+                                                    } else {
+                                                        None
+                                                    }
+                                                })
+                                                .collect::<Vec<_>>())
+                                        })
+                                        .for_each_with(send, |send, val| send.send(val).expect("Receiver should be allocated for scope duration"));
```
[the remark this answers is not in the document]
dbrandenstein: No, but `send` returns a result which must be used.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+
+        let tiling_specification = context.tiling_specification();
+
+        let out_desc = RasterResultDescriptor {
+            spatial_reference: in_desc.spatial_reference,
+            data_type: RasterDataType::F64,
+            measurement: Measurement::default(),
+            bbox: None,
+            time: in_desc.time,
+            resolution: None,
```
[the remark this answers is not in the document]
dbrandenstein: The output resolution is always equal to the query solution. But it is not known at this point.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+        let vector_source = self.sources.vector.initialize(context).await?;
+        let in_desc = vector_source.result_descriptor();
+
+        let tiling_specification = context.tiling_specification();
+
+        let out_desc = RasterResultDescriptor {
+            spatial_reference: in_desc.spatial_reference,
+            data_type: RasterDataType::F64,
+            measurement: Measurement::default(),
+            bbox: None,
```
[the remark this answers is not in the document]
dbrandenstein: I guess no? A safe cut-off point would be the query rectangle + one tile height/width in each direction. But because this depends on the concrete query rectangle and query resolution it is not known at this point.
Maybe another option would be to return the whole area of use of the respective projection (if given in the source descriptor)?

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+            }
+            .boxed()),
+            GridOrDensity::Density(params) => InitializedDensityRasterization::new(
+                vector_source,
+                out_desc,
+                tiling_specification,
+                params.radius,
+                params.stddev,
+            )
+            .map(InitializedRasterOperator::boxed),
```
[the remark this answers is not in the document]
dbrandenstein: Well, I could. But I wanted to do the validation of `radius` in `new` and Clippy dislikes returning something else than `Self/Result<Self,_>` in a function called `new` (in this case `Result<dyn InitializedRasterOperator, error::Error>`), so I opted for this solution. I could also ignore the Clippy warning and use `boxed()` in the `new` function or call it `create_boxed` or something instead if you prefer using `boxed()`.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+        radius: f64,
+        stddev: f64,
+    ) -> Result<Self, error::Error> {
+        ensure!(
+            radius > 0.,
+            error::InvalidOperatorSpec {
+                reason: "The radius for density rasterization must be strictly positive."
+                    .to_string()
+            }
+        );
```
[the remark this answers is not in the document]
dbrandenstein: You are right. https://github.com/geo-engine/geoengine/pull/683/commits/9b45219070a09e93db08075638a78c0aceec3946

operators/src/processing/rasterization/mod.rs
dbrandenstein: What exactly is there to say about it? It's more or less copied from other operators, where it is not commented as well.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+    type RasterType = f64;
+
+    async fn raster_query<'a>(
+        &'a self,
+        query: RasterQueryRectangle,
+        ctx: &'a dyn QueryContext,
+    ) -> util::Result<BoxStream<'a, util::Result<RasterTile2D<Self::RasterType>>>> {
+        let grid_resolution = match self.grid_size_mode {
+            GridSizeMode::Fixed => SpatialResolution {
+                x: f64::max(self.spatial_resolution.x, query.spatial_resolution.x),
```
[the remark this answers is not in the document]
dbrandenstein: I decided that it makes sense to limit the resolution to the query resolution in case the grid resolution is finer (smaller), because one would not notice the difference anyways. It avoids the computation of an unnecessarily fine grid which is then downsampled anyways.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                y: f64::max(self.spatial_resolution.y, query.spatial_resolution.y),
+            },
+            GridSizeMode::Relative => SpatialResolution {
+                x: f64::max(
+                    self.spatial_resolution.x * query.spatial_resolution.x,
```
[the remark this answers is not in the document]
dbrandenstein: It's not squared, it's the query resolution scaled by the grid resolution parameter. Also explained in https://github.com/geo-engine/geoengine/pull/683/commits/e5da1695b25feb3f71272a1a504d7ba438abd053.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                    query.spatial_resolution.x,
+                ),
+                y: f64::max(
+                    self.spatial_resolution.y * query.spatial_resolution.y,
+                    query.spatial_resolution.y,
+                ),
+            },
+        };
```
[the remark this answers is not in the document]
dbrandenstein: Yes, the limit is the query resolution and with that implicitly the tile size.

operators/src/processing/rasterization/mod.rs
dbrandenstein: Good catch, not anymore. These simple additions didn't really seem to benefit from it so I removed the parallel computations. https://github.com/geo-engine/geoengine/pull/683/commits/892b2e8fe4f0d3e3d20dad595b96005e9af51462

operators/src/processing/rasterization/mod.rs
dbrandenstein: Yes, you're right. If you want I can check whether it's MultiPoint data in the beginning and then just return a stream of empty tiles. Shouldn't really have a performance impact but probably is clearer.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,980 @@
...
+                        tile_bounds.lower_left() - Coordinate2D::new(self.radius, self.radius),
+                        tile_bounds.upper_right() + Coordinate2D::new(self.radius, self.radius),
+                    )?,
+                    time_interval: query.time_interval,
+                    spatial_resolution: query.spatial_resolution,
+                };
+
+                let tile_geo_transform = tile_info.tile_geo_transform();
+
+                let mut tile_data;
```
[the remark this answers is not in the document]
dbrandenstein: I described the method in general here. Lmk if it needs more detail. https://github.com/geo-engine/geoengine/pull/683/commits/cb3367da16b7a5ee740fbdd6c257b12767b1e446

operators/src/processing/rasterization/mod.rs
dbrandenstein: You probably meant the GridRasterization `raster_query` method and not the implementation of `InitializedRasterOperator`. It looked like your comment was referring to the latter because the conversation view gave the whole code block as context.
I described the method in general here. Lmk if it needs more detail. https://github.com/geo-engine/geoengine/pull/683/commits/cb3367da16b7a5ee740fbdd6c257b12767b1e446

operators/src/processing/rasterization/mod.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Sounds reasonable. Could you add this as a comment?

operators/src/processing/rasterization/mod.rs
ChristianBeilschmidt: What you should do, in any case, is wrap a computation by a `spawn_blocking`.
The reason is that async stuff relies on collaborative scheduling, which means returning after a very short time to let others also use the thread pool. This means everything inside async and between calls of `.await`, which releases scope, should be kind of quick. Thus, computations should be inside such blocks, which you need to await and do the computation inside another "thread".

operators/src/processing/rasterization/mod.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: cf. above, it's about initializing tile_data

operators/src/processing/rasterization/mod.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Then just say `let _ = send.send(…)`. Or you could do a `debug_assert!(send.send(…).is_some())`. Just to reduce potential panics in the code.

operators/src/processing/rasterization/mod.rs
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/27b4d08ab72076434fed4c6edfe2b5d3b40ac068

operators/src/processing/rasterization/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/c66df5ccc70d40de1fb304c2dba4cef57c5a3025
I also changed from returning an empty grid to a zeroed grid when the vector data type is not `MultiPoint`.
It makes more sense in this case: If there exist no points, the count is zero. Instead of the grid doesn't exist.

operators/src/processing/rasterization/mod.rs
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/57162ad1b04bcc26c500ceeb6ed41da48e68cd60

operators/src/processing/rasterization/mod.rs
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/c66df5ccc70d40de1fb304c2dba4cef57c5a3025

operators/src/processing/rasterization/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/fc5eea38ef9908e97d8958a7ef60d36d06618563
I now use the tokio mpsc channel and `spawn_blocking` and got rid of the rayon scope in the process.
I think the loop in line 301 should be fast enough to stay as is. If you disagree I can change that as well.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,1082 @@
...
+                let mut grid_data = vec![0.; grid_size_x * grid_size_y];
+                while let Some(chunk) = chunks.next().await {
+                    for &coord in chunk?.coordinates() {
+                        if !grid_spatial_bounds.contains_coordinate(&coord) {
+                            continue;
+                        }
+                        let [y, x] = grid_geo_transform.coordinate_to_grid_idx_2d(coord).0;
+                        grid_data[x as usize + y as usize * grid_size_x] += 1.;
+                    }
+                }
```
ChristianBeilschmidt, line 308: Looping over all coordinates also looks like blocking to me, so this should be inside spawn_blocking as well.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,1082 @@
...
+                let tile_geo_transform = tile_info.tile_geo_transform();
+
+                let mut chunks = points_processor.query(vector_query, ctx).await?;
+
+                let mut tile_data = vec![0.; tile_size_x * tile_size_y];
+
+                while let Some(chunk) = chunks.next().await {
+                    let chunk = chunk?;
+                    let (send, mut receive) = tokio::sync::mpsc::unbounded_channel();
+                    ctx.thread_pool().install(|| {
```
ChristianBeilschmidt, line 403: Please use `spawn_blocking_with_thread_pool` here since this is a blocking computation.

ChristianBeilschmidt (changes requested):

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,1082 @@
...
+
+                Ok(RasterTile2D::new_with_tile_info(
+                    query.time_interval,
+                    tile_info,
+                    GridOrEmpty::Grid(tile_grid.into()),
+                ))
+            }),
+    )
+    .boxed()
+}
```
ChristianBeilschmidt, line 506: Wouldn't it be better to output EmptyTiles instead of outputting tiles with zeroes?

ChristianBeilschmidt (changes requested):

operators/src/processing/rasterization/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/07e582ba672fe3453ad1d4670cabce9756c23a3e

operators/src/processing/rasterization/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/1bb2b3ec81799658c4f72d696aec5079bf1ac9e9

operators/src/processing/rasterization/mod.rs
dbrandenstein: It did before, but since the operator counts points, outputting a zero in case no points are present instead of "nothing" seemed more logical to me. Empty tiles are also not treated like zeroed tiles by the colorizer iirc. I think it is confusing to get a different result when a vector source does not contain any points (because it has a different type) versus when it does, but not in the queried region (or is actually empty, but still of the MultiPoint type).

operators/src/processing/rasterization/mod.rs
```diff
@@ -446,17 +454,16 @@ impl RasterQueryProcessor for DensityRasterizationQueryProcessor {
                                             .collect::<Vec<_>>(),
                                     )
                                 })
-                                .for_each_with(send, |send, val| {
-                                    debug_assert!(send.send(val).is_ok());
+                                .for_each(|res| {
+                                    let mut tile_data = shared_tile_data.lock().unwrap();
```
ChristianBeilschmidt, line 458: This somehow feels wrong, since you basically sequentialize the access to the tile.

What about something like this?
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=20989f8bc076d3683b4b4a670b74977d

Generally, we would need a benchmark for that to really assess it.

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,1089 @@
...
+#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "type")]
+pub enum GridOrDensity {
+    /// A grid which summarizes points in cells (2D histogram)
+    Grid(GridParams),
+    /// A heatmap calculated from a gaussian density function
+    Density(DensityParams),
+}
+
```
ChristianBeilschmidt, line 66: It is allowed to have a spatial resolution of (0,0). Maybe we should prohibit that?

ChristianBeilschmidt (changes requested):

dbrandenstein:
> grafik
> 
> It looks like there is a fixed "aura" around each point. Is this behavior due to the cutoff-point?

Yes. The radius should optimally be chosen by the user such that its introduced inaccuracies are not (or just slightly) visible in the resulting raster. I explained this a bit more in the docs.

operators/src/processing/rasterization/mod.rs
dbrandenstein: This is one of the reasons I limited the grid resolution to the query resolution (= its size to the tile size). I assumed that there is already some kind of check in place which prevents the user from ever querying a (0,0) resolution, which would also lead to all kinds of problems elsewhere (e.g., tiling strategy).
If you want me to prohibit it here I see no disadvantage, though. The one use case I could see is if someone wants the grid resolution to become "infinitely" smaller when zooming in. However, the same can be achieved with the relative grid size mode and a multiplier of 1.

operators/src/processing/rasterization/mod.rs
dbrandenstein: Simplified in https://github.com/geo-engine/geoengine/pull/683/commits/ac01de8afac02cff861242474bd23b6c894569b9

operators/src/processing/rasterization/mod.rs
```diff
@@ -0,0 +1,1057 @@
...
+    /// Determines how to interpret the grid resolution
+    grid_size_mode: GridSizeMode,
+}
+
+/// The parameter spec for `Rasterization`
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RasterizationParams {
+    grid_or_density: GridOrDensity,
+}
```
ChristianBeilschmidt, line 91: Sry, I just saw that.
You could get rid of this struct an used `GridOrDensity` directly as params.

`pub type Rasterization = Operator<GridOrDensity, SingleVectorSource>;`

ChristianBeilschmidt (changes requested):

operators/src/processing/rasterization/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/683/commits/e76f07d3aa946539f37361e631cfc725ac0bc3b2

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #684, opened 2022-11-22
add over/under color feature to colorizer

MarWeUMR:
- [ x] I added an entry to [`CHANGELOG.md`](CHANGELOG.md) if knowledge of this change could be valuable to users.

---

There are two new properties in the colorizer:

- `over_color`
- `under_color`

These colors can be used as an alternative to the `default_color`. They are applied, if a given value is outside of the margin defined by the breakpoints of the colorizer. The implementation is loosely based on https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Colormap.html#matplotlib.colors.Colormap


MarWeUMR:
[Overflow-/Underflow-Farben](https://trello.com/c/n9Wl1oB2/1050-overflow-underflow-farben)

datatypes/src/operations/image/colorizer.rs
```diff
@@ -18,19 +18,25 @@ pub enum Colorizer {
     LinearGradient {
         breakpoints: Breakpoints,
         no_data_color: RgbaColor,
-        default_color: RgbaColor,
+        default_color: RgbaColor, // TODO: deprecated, this will be removed. Please use over/under values.
+        over_color: Option<RgbaColor>,
+        under_color: Option<RgbaColor>,
```
ChristianBeilschmidt, line 23: It should be the other way around.
Maybe you do something like an "either"… either `default_color` or `over_color`.

But you serialize only into `over_color` and `under_color`, not `default_color` anymore.

And maybe you can deserialize this either type directly into this pair of colors :thinking: .

datatypes/src/operations/image/colorizer.rs
```diff
@@ -211,6 +229,24 @@ impl Colorizer {
...
+    }
+
+    pub fn under_color(&self) -> Option<RgbaColor> {
+        match self {
+            Colorizer::LinearGradient { under_color, .. }
+            | Colorizer::LogarithmicGradient { under_color, .. }
+            | Colorizer::Palette { under_color, .. } => *under_color,
+            Colorizer::Rgba => None,
+        }
+    }
```
ChristianBeilschmidt, line 248: Should be un-optional as well.

datatypes/src/operations/image/colorizer.rs
```diff
@@ -728,7 +833,49 @@ mod tests {
...
+            serde_json::json!({
+                "type": "palette",
+                "colors": {
+                    "1": [255, 255, 255, 255],
+                    "2": [0, 0, 0, 255]
+                },
+                "noDataColor": [0, 0, 0, 0],
+                "defaultColor": [0, 0, 0, 0],
+                "overColor": [255, 0, 255, 255],
+                "underColor": [255, 0, 255, 255]
```
ChristianBeilschmidt, line 878: Somehow, only one should be set. This should not contradict.

operators/src/util/raster_stream_to_png.rs
```diff
@@ -90,6 +90,8 @@ pub fn default_colorizer_gradient<T: Pixel>() -> Result<Colorizer> {
         ],
         RgbaColor::transparent(),
         RgbaColor::pink(),
+        None,
+        None,
```
ChristianBeilschmidt, line 94: I would say… black for under and white for over instead of pink.

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -214,6 +214,8 @@ impl SentinelS2L2aCogsDataProvider {
                                     ],
                                     RgbaColor::transparent(),
                                     RgbaColor::transparent(),
+                                    None,
+                                    None,
```
ChristianBeilschmidt, line 218: Please also set under/over instead of default here

ChristianBeilschmidt (changes requested):

datatypes/src/operations/image/colorizer.rs
MarWeUMR: Ok, I really hope that I'm not misunderstanding something here, but this is the new implementation:
https://github.com/geo-engine/geoengine/pull/684/commits/1667bc7a42482d61c00da0265b0de578c593e20f#diff-36b1b82e93e26eceb563391b9c7a3994a3254b9416494a2d64d3b3bfd5da45b1R23

datatypes/src/operations/image/colorizer.rs
MarWeUMR: [1667bc7](https://github.com/geo-engine/geoengine/pull/684/commits/1667bc7a42482d61c00da0265b0de578c593e20f#diff-36b1b82e93e26eceb563391b9c7a3994a3254b9416494a2d64d3b3bfd5da45b1R315) Line: 315

datatypes/src/operations/image/colorizer.rs
MarWeUMR: This is now being tested with two distinct operations.

operators/src/util/raster_stream_to_png.rs
MarWeUMR: There are a few more places like this. I hope I've caught all of them now.

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
MarWeUMR: see above.

datatypes/src/operations/image/colorizer.rs
```diff
@@ -3,44 +3,140 @@ use crate::operations::image::RgbaTransmutable;
...
+                under_color: under_color
+                    .ok_or_else(|| serde::de::Error::custom("Missing underColor field"))?,
+            }),
+            //TODO: what about the `Rgba` variant?
+            _ => Err(serde::de::Error::custom(
+                "Encountered an unknown colorizer type.",
+            )),
+        }
+    }
+}
```
ChristianBeilschmidt, line 131: Now, we cannot deserialize the Rgba variant anymore.

You could make use of serde's `#[serde(from = "FromType")]`, cf. https://serde.rs/container-attrs.html#from.

Then, you could have all fields on the deserializer structs, but optionally.
Then, you only succeed if you have either over/under OR default.
For Rgba, you just pass the enum variant through.

Then, we can say that this Deserializer struct will be removed in the future, removing the default_color option.
And this seems less brittle but more verbose than the thing above.

datatypes/src/operations/image/colorizer.rs
```diff
@@ -211,6 +312,24 @@ impl Colorizer {
         }
     }
 
+    pub fn over_color(&self) -> RgbaColor {
+        match self {
+            Colorizer::LinearGradient { over_color, .. }
+            | Colorizer::LogarithmicGradient { over_color, .. }
+            | Colorizer::Palette { over_color, .. } => *over_color,
+            Colorizer::Rgba => None,
```
ChristianBeilschmidt, line 320: Why is it possible to return `None` if the return type is not `Option`? :thinking: 

If this is a problem, just output transparent for Rgba.

datatypes/src/operations/image/colorizer.rs
```diff
@@ -211,6 +312,24 @@ impl Colorizer {
...
+        }
+    }
+
+    pub fn under_color(&self) -> RgbaColor {
+        match self {
+            Colorizer::LinearGradient { under_color, .. }
+            | Colorizer::LogarithmicGradient { under_color, .. }
+            | Colorizer::Palette { under_color, .. } => *under_color,
+            Colorizer::Rgba => None,
+        }
```
ChristianBeilschmidt, line 330: same here

datatypes/src/operations/image/colorizer.rs
```diff
@@ -3,44 +3,140 @@ use crate::operations::image::RgbaTransmutable;
...
+        over_color: RgbaColor,
+        under_color: RgbaColor,
     },
     #[serde(rename_all = "camelCase")]
     Palette {
         colors: Palette,
         no_data_color: RgbaColor,
-        default_color: RgbaColor,
+        over_color: RgbaColor,
+        under_color: RgbaColor,
```
ChristianBeilschmidt, line 38: For palettes, it does not make sense to have an over or under color, since this means exact values.
We should keep the default color here.

datatypes/src/operations/image/colorizer.rs
```diff
@@ -458,28 +585,33 @@ impl<'c> ColorMapper<'c> {
...
+                } else if value > *max_value {
+                    *over_color
                 } else {
                     let color_table_factor = (color_table.len() - 1) as f64;
                     let table_entry = f64::round(
                         color_table_factor * ((value - *min_value) / (*max_value - *min_value)),
                     ) as usize;
-                    *color_table.get(table_entry).unwrap_or(default_color)
+                    // TODO:                                    vvv ----this was default color, what should it be now?
+                    *color_table.get(table_entry).unwrap_or(no_data_color)
```
ChristianBeilschmidt, line 604: cf. comment above

datatypes/src/operations/image/colorizer.rs
```diff
@@ -458,28 +585,33 @@ impl<'c> ColorMapper<'c> {
...
             ColorMapper::ColorMap {
                 color_map,
                 no_data_color,
-                default_color,
+                ..
             } => {
                 if let Ok(value) = NotNan::<f64>::new(value.as_()) {
-                    *color_map.0.get(&value).unwrap_or(default_color)
+                    // TODO:                               vvv -----this was default color, what should it be now?
+                    *color_map.0.get(&value).unwrap_or(no_data_color)
```
ChristianBeilschmidt, line 614: cf. comment above

datatypes/src/operations/image/colorizer.rs
```diff
@@ -669,6 +802,34 @@ mod tests {
...
+    #[test]
+    fn logarithmic_color_table_with_extrema() {
+        let colorizer = Colorizer::logarithmic_gradient(
+            vec![
+                (5.0, RgbaColor::black()).try_into().unwrap(),
+                (15.0, RgbaColor::white()).try_into().unwrap(),
+            ],
+            RgbaColor::transparent(),
+            RgbaColor::pink(),
+            RgbaColor::pink(),
```
ChristianBeilschmidt, line 814: Please use different colors here, so that you can test that you are not mistakenly switching colors, etc.

datatypes/src/operations/image/colorizer.rs
```diff
@@ -669,6 +802,34 @@ mod tests {
...
+
+        // TODO: should it be like this here?
+        // or should the colorizer take the boundary points (5 & 15) and use the
+        // remaining slots to generate the table? In this case 5: black, 10: grey 15: white
+        // instead of evenly spacing it over the values including extrema?
+        assert_eq!(color_table[0], RgbaColor::pink()); // undercolor
+        assert_eq!(color_table[1], RgbaColor::new(78, 78, 78, 255)); // at 7
+        assert_eq!(color_table[2], RgbaColor::new(161, 161, 161, 255)); // at 10
+        assert_eq!(color_table[3], RgbaColor::new(222, 222, 222, 255)); // at 13
+        assert_eq!(color_table[4], RgbaColor::pink()); // overcolor
```
ChristianBeilschmidt, line 830: If you specify a color for the value `5.0`, you want to have this color back and not another color.
So, the first should be black here and the last one white.

ChristianBeilschmidt (changes requested):

datatypes/src/operations/image/colorizer.rs
MarWeUMR: I feel like I'm not understanding how breakpoints/colortables are supposed to work.
If the colorizer is specified with breakpoints `[5.,15.]` and we now generate a colortable of 5 values from `[4.,...,16.]`, shouldn't the first and last values be the over/under color in this case?

datatypes/src/operations/image/colorizer.rs
MarWeUMR: But in the case of a color table, its used only with linear or log colorizer, isn't it? And there we don't have the default color anymore now. Should it be transparent in this case?

datatypes/src/operations/image/colorizer.rs
MarWeUMR: Ok I've implemented it now according to your example from slack.
https://github.com/geo-engine/geoengine/pull/684/commits/d6c04ea287ef8ed52afd716ccdb7ca14e51aacf5

datatypes/src/operations/image/colorizer.rs
MarWeUMR: That was totally my fault, sorry

datatypes/src/operations/image/colorizer.rs
MarWeUMR: Palette now uses default color again. https://github.com/geo-engine/geoengine/pull/684/commits/f0a0bd67359b995ca311acbb57ac04b7707288b8

datatypes/src/operations/image/colorizer.rs
MarWeUMR: Changed back with https://github.com/geo-engine/geoengine/pull/684/commits/f0a0bd67359b995ca311acbb57ac04b7707288b8

datatypes/src/operations/image/colorizer.rs
MarWeUMR: https://github.com/geo-engine/geoengine/pull/684/commits/cd8b2dc6a2dfab354845303f76b731d2185fce6c

datatypes/src/operations/image/colorizer.rs
```diff
@@ -9,6 +9,19 @@ use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::str::FromStr;
 
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[serde(untagged)]
+pub enum ColorFields {
```
ChristianBeilschmidt, line 15: Maybe just call it `DefaultColors`, since the no data color and breakpoints are also color fields.

datatypes/src/operations/image/colorizer.rs
```diff
@@ -458,21 +509,31 @@ impl<'c> ColorMapper<'c> {
...
                 let value: f64 = value.as_();
                 if f64::is_nan(value) {
                     *no_data_color
-                } else if value < *min_value || value > *max_value {
-                    *default_color
+                } else if value < *min_value {
+                    match color_fields {
+                        ColorFields::DefaultColor { default_color } => *default_color,
+                        ColorFields::OverUnder { under_color, .. } => *under_color,
+                    }
```
ChristianBeilschmidt, line 521: If you would implement over_color and under_color on this enum, you could call it here and also use it in `over_color` of the colorizer.

datatypes/src/operations/image/colorizer.rs
ChristianBeilschmidt: This case should not happen in general.
I guess, the most reasonable thing would be to check if `table_entry` would be < 0 and then use the under_color.

If the entry is above the largest index, we should use the over color.

You can check if the first case can ever happen. If not, we should use the over color.

datatypes/src/operations/image/colorizer.rs
ChristianBeilschmidt: Ah, I thought it was between 5 and 15. In this case, you are right.

You should set the over and under color to different values and then check that the implementation works.

ChristianBeilschmidt (changes requested):
I think it is almost there.

datatypes/src/operations/image/colorizer.rs
MarWeUMR: Ok, upon further investigation, I think we could go with an `except(...)` call here (or add a new error?), because I don't see any possibility for this call to fail.

```
let color_table_factor = (color_table.len() - 1) as f64;
let table_entry = f64::round(
    color_table_factor * ((value - *min_value) / (*max_value - *min_value))
) as usize;
```
There is no way for this to be outside of the index set of the color table, unless `value` is outside of the min/max range. But in that case, the other conditional branches would have caught it.

datatypes/src/operations/image/colorizer.rs
MarWeUMR: Changed with https://github.com/geo-engine/geoengine/pull/684/commits/cacee41b1ac1cdda64e8404e7a84816f18cc8868

datatypes/src/operations/image/colorizer.rs
MarWeUMR: Changed with https://github.com/geo-engine/geoengine/pull/684/commits/86ec9cd2db7b3f5fdc9ce59ee03ac2a652c072b7

datatypes/src/operations/image/colorizer.rs
MarWeUMR: Changed with https://github.com/geo-engine/geoengine/commit/cacee41b1ac1cdda64e8404e7a84816f18cc8868

datatypes/src/operations/image/colorizer.rs
ChristianBeilschmidt: In theory, yes, but the round computation is done within this branch. This does not seem super robust.

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #691, opened 2022-12-05
add handler for adding public system datasets

michaelmattig:

---

<TEXT>


michaelmattig:
[[RESPECT] 3.6.2 Admin-Schnittstelle für Rasterdaten: Datasets](https://trello.com/c/Gl7O8c9F/907-respect-362-admin-schnittstelle-f%C3%BCr-rasterdaten-datasets)

Settings-default.toml
```diff
@@ -158,3 +158,6 @@ scopes = [ ]
 [open_telemetry]
 enabled = false
 endpoint = "127.0.0.1:6831"
+
+[data]
+volumes = { f6aa9f3d-a211-43e8-9b91-b23ab9632791 = "test_data/"}
```
ChristianBeilschmidt, line 163:
```suggestion
volumes = { f6aa9f3d-a211-43e8-9b91-b23ab9632791 = "test_data/"}

```

services/src/datasets/upload.rs
```diff
@@ -3,15 +3,37 @@ use std::path::{Path, PathBuf};
...
 identifier!(FileId);
+identifier!(VolumeId);
+
+pub trait AdjustFilePath {
+    fn adjust_file_path(&self, file_path: &Path) -> Result<PathBuf>;
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, ToSchema)]
+pub struct Volume {
+    pub id: VolumeId,
```
ChristianBeilschmidt, line 25: Can't we give volumes a name like in Docker, etc.?

We always have these cryptic ids that no one likes to work with.

Settings-default.toml
ChristianBeilschmidt, line 163: Do we want to support multiple volumes for data?

services/src/handlers/datasets.rs
```diff
@@ -51,12 +57,45 @@ where
         web::scope("/dataset")
             .service(web::resource("/suggest").route(web::get().to(suggest_meta_data_handler::<C>)))
             .service(web::resource("/auto").route(web::post().to(auto_create_dataset_handler::<C>)))
+            .service(web::resource("/volumes").route(web::get().to(list_volumes_handler)))
+            .service(
+                web::resource("/public").route(web::post().to(create_system_dataset_handler::<C>)),
```
ChristianBeilschmidt, line 62: This sounds like a temporary solution to do it under `/public`.

Wouldn't it be better to treat user and public datasets the same and just give them other access rights?

services/src/handlers/datasets.rs
```diff
@@ -144,13 +183,69 @@ async fn get_dataset_handler<C: Context>(
     Ok(web::Json(dataset))
 }
 
+/// Creates a new dataset using files available on the volumes.
+/// Requires an admin session.
+#[utoipa::path(
+    tag = "Datasets",
+    post,
+    // TODO: "public" doesn't really make sense, because there are no private datasets. But this is more consistent with pro version
```
ChristianBeilschmidt, line 191: Ah okay, you also saw it yourself that this handler can only be temporary.

services/src/pro/handlers/datasets.rs
```diff
@@ -0,0 +1,426 @@
...
+            .service(web::resource("/auto").route(web::post().to(auto_create_dataset_handler::<C>)))
+            .service(web::resource("/volumes").route(web::get().to(list_volumes_handler)))
+            .service(
+                web::resource("/public").route(web::post().to(create_system_dataset_handler::<C>)),
+            )
+            .service(web::resource("/{dataset}").route(web::get().to(get_dataset_handler::<C>)))
+            .service(web::resource("").route(web::post().to(create_user_dataset_handler::<C>))), // must come last to not match other routes
+    )
+    .service(web::resource("/datasets").route(web::get().to(list_datasets_handler::<C>)));
+}
```
ChristianBeilschmidt, line 42: Why do you define this twice?

ChristianBeilschmidt (changes requested):

services/src/datasets/upload.rs
michaelmattig: My thinking was that admins most likely will look for the path when selecting a volume. In docker the path is the name of the volume right? We could omit the id and only have the path, but then we would have to put some limitations on the name, e.g. length if we want to use it in URLs, permissions, etc. in the long run.

Settings-default.toml
michaelmattig: I think so. Even on our dev machine we have a data_ssd for quick access and a data_raid for larger files.

services/src/handlers/datasets.rs
michaelmattig: unfortunately it doesn't work to have the same route, and use a different handler depending on the kind of session token. I tried a single handler with Either<Admin, User> and Either<SystemData, UserData> but it wouldn't match. I also tried two services one with admin and one with regular session, but it would always match the first one and throw an error if the session is not an admin one e.g.

Do you maybe have an idea of how to do this?

services/src/pro/handlers/datasets.rs
michaelmattig: because I need a different handler for public datasets in the pro version that explicitly sets the permissions. The free handler doesn't know about permissions.

services/src/handlers/datasets.rs
```diff
@@ -144,13 +183,69 @@ async fn get_dataset_handler<C: Context>(
...
+        )
+    ),
+    security(
+        ("session_token" = [])
+    )
+)]
+async fn create_system_dataset_handler<C: Context>(
+    _session: AdminSession,
+    ctx: web::Data<C>,
+    create: web::Json<CreateSystemDataset>,
```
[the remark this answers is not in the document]
michaelmattig: do you mean the CreateSystemDataset? We use structs from different files all the time, like AdminSession and Context

services/src/pro/handlers/users.rs
```diff
@@ -444,6 +445,7 @@ mod tests {
     ) -> ServiceResponse
     where
         C::ProjectDB: ProProjectDb,
+        C::DatasetDB: UpdateDatasetPermissions,
```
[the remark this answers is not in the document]
michaelmattig: ideally yes, but the ProContext is just a sub trait of the Context trait. And a Context just requires the Context::DatasetDB to be a DatabaseDb and nothing more.

services/src/datasets/upload.rs
ChristianBeilschmidt: Just allow characters, digits and hyphens or so.

In Docker, they have names:
https://docs.docker.com/storage/volumes/

So its not the path, it is a simple name like "my-vol"

services/src/handlers/datasets.rs
ChristianBeilschmidt: I guess `Either` is only about being able to deserialize stuff.
So, are both things differentiable by their inner schema?

services/src/handlers/datasets.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Yeah, but `AdminSession` you use everywhere, and this thing is only for this handler.

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Require a DatasetDb to have `UpdateDatasetPermissions`.

services/src/datasets/upload.rs
michaelmattig: f9ce82dda228b2a2ab5852b8cab78f7748c19cd9

services/src/handlers/datasets.rs
michaelmattig: what is an inner schema? The CreateUserDataset and CreateSystemDataset have differently named fields. Session and AdminSession are identical though.

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
michaelmattig: Ok, but how?

services/src/handlers/datasets.rs
ChristianBeilschmidt: But `AdminSession`  has a custom deserializer.

So, is there an error why this does not match?

services/src/handlers/datasets.rs
```diff
@@ -144,13 +183,69 @@ async fn get_dataset_handler<C: Context>(
...
+    let volumes = get_config_element::<Data>()?.volumes;
+    let volume_path = volumes
+        .get(&create.volume)
+        .ok_or(error::Error::UnknownVolumeId)?;
+    let volume = Volume {
+        id: create.volume,
+        path: volume_path.clone(),
+    };
+
+    adjust_meta_data_path(&mut definition.meta_data, &volume)?;
```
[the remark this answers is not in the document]
ChristianBeilschmidt: Can we prohibit both?

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: I would say just add the trait to the super trait (Context).

services/src/handlers/datasets.rs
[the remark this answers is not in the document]
michaelmattig: we do, with `canonicalize_subpath` and `path_with_base_path`

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
michaelmattig: the super trait is for non-pro as well, so I can't add it there, because there are no users or permissions in non-pro

services/src/handlers/datasets.rs
michaelmattig: I tried

```
pub async fn create_user_dataset_handler<C: Context>(
    session: Either<C::Session, AdminSession>,
    ctx: web::Data<C>,
    create: Either<web::Json<CreateUserDataset>, web::Json<CreateSystemDataset>>,
) -> Result<web::Json<IdResponse<DatasetId>>> {
    let (session, create) = match (session, create) {
        (Either::Left(session), Either::Left(create)) => ...
        (Either::Right(session), Either::Right(create)) => ...
```` 

but then I get

> [services/src/handlers/datasets.rs:1143] res.response().body() = BoxBody(
    Bytes(
        b"{\"error\":\"BodyDeserializeError\",\"message\":\"EOF while parsing a value at line 1 column 0\"}",
    ),
)

so actix can't seem to handle the `Either`in the body? I don't know.

File: services/src/handlers/datasets.rs (hunk 8 above)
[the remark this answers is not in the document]
Comment (ChristianBeilschmidt) [reply services/src/handlers/datasets.rs:227]: But we write it down as a full path in the dataset. Can't we have it relative to a volume?

File: services/src/pro/handlers/users.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (ChristianBeilschmidt) [reply services/src/pro/handlers/users.rs:448]: There is a ProContext trait for that I recall

File: services/src/handlers/datasets.rs (hunk 3 above)
Comment (ChristianBeilschmidt) [reply services/src/handlers/datasets.rs:62]: Does Either work for web::Json at all?

File: services/src/pro/handlers/users.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (michaelmattig) [reply services/src/pro/handlers/users.rs:449]: Yes, but the DatasetDb ist defined in the Context trait, which is a super trait of the ProContext

File: services/src/handlers/datasets.rs (hunk 3 above)
Comment (ChristianBeilschmidt) [reply services/src/handlers/datasets.rs:62]: Yeah, that also sounds reasonable. I would say you can use the second definition if the rest is the same, right?

If you would use `#[serde(untagged)]` with the first struct, would that be the same effect than `Either`?

File: services/src/pro/handlers/users.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (ChristianBeilschmidt) [reply services/src/pro/handlers/users.rs:449]: Can't we set the database type field from the ProContext rather than the context itself :thinking: .

File: services/src/handlers/datasets.rs (hunk 3 above)
Comment (michaelmattig) [reply services/src/handlers/datasets.rs:62]: I tried it, but it also doesn't work. The documentation of `Either` says:

> /// It is important to note that this extractor, by necessity, buffers the entire request payload
/// as part of its implementation. Though, it does respect any `PayloadConfig` maximum size limits.

so I guess one cannot combine an `Either` extractor with other extractors. If I require a regular session, everything works with my new struct with the enum inside, but when I add the `Either` to match on both session types, it breaks. In the other handler that use `Either` we do not use an extractor from the payload, but only from the path and query.

File: services/src/pro/handlers/users.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (michaelmattig) [reply services/src/pro/handlers/users.rs:449]: We could add another associated type to the `ProContext` for a ProDatasetDB with a ProDatasetDb bound and then a new accessor method `pro_dataset_db()` which would be called instead of the regular `dataset_db()`. Other than that I don't see how to solve it and if you have an idea please don't leave me guessing ;)

File: services/src/handlers/datasets.rs (hunk 3 above)
Comment (ChristianBeilschmidt) [reply services/src/handlers/datasets.rs:62]: Maybe you can write another extractor that retrieves a UUID.
Then, it checks whether this UUID is the admin UUID and returns the admin thing-e, or otherwise a normal Session to check.

File: services/src/pro/handlers/users.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (ChristianBeilschmidt) [reply services/src/pro/handlers/users.rs:449]: Yeah, that is what I thought.

File: services/src/handlers/datasets.rs (hunk 3 above)
Comment (michaelmattig) [reply services/src/handlers/datasets.rs]: 68ba46b28acc48bfeb8fad8ef79447178e18c62b

File: services/src/handlers/datasets.rs (hunk 4 above)
Comment (michaelmattig) [reply services/src/handlers/datasets.rs]: 68ba46b28acc48bfeb8fad8ef79447178e18c62b

File: services/src/handlers/datasets.rs (hunk 8 above)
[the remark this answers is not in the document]
Comment (michaelmattig) [reply services/src/handlers/datasets.rs:311]: it's not a bad idea to do this. This would also allow us to move whole volumes to a different path. We would have to change all of the MetaData definitions though and make the `path` two-parts: 1. volume/upload ref 2. path relative to volume/upload. And then when computing the resultof  `loading_info` turn it into an absolute path that gdal then uses. We also would have to handle the case of external resources differently I guess, because how do URLs fit into this? and the virtual filesystems of gdal that come before the file path? Then we would also need to adjust all of the existing definition files.

I would propose to do this in a separate task.

File: services/src/pro/handlers/users.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (michaelmattig) [reply services/src/pro/handlers/users.rs]: b8f303224a732d82a5e1ba65ce2081429843c069

File: services/src/pro/server.rs [hunk 9]
```diff
@@ -34,6 +35,7 @@ async fn start<C>(
 where
     C: ProContext,
     C::ProjectDB: ProProjectDb,
+    C::DatasetDB: UpdateDatasetPermissions,
```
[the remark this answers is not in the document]
michaelmattig: b8f303224a732d82a5e1ba65ce2081429843c069

services/src/handlers/datasets.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: I guess you are right.

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors retry

ChristianBeilschmidt:
bors retry

ChristianBeilschmidt:
bors retry

ChristianBeilschmidt:
bors retry

ChristianBeilschmidt:
bors retry

PR #703, opened 2023-01-04
Save Layer as a new Dataset

glombiewski:

---

- Moved some functionality of saving a workflow to a dataset to datasets module
- Changed the result descriptor of datasets originating from workflows to match their query rectangle
- Added a layer_to_dataset function to layers handler for saving a layer via the workflow functionality


glombiewski:
[[RESPECT] 3.4 Overviews für Rasterdaten erzeugen](https://trello.com/c/OfeNYbKS/908-respect-34-overviews-f%C3%BCr-rasterdaten-erzeugen)

services/src/datasets/create_from_workflow.rs
```diff
@@ -0,0 +1,268 @@
...
+                no_data_value: Default::default(), // TODO: decide how to handle the no data here
+                spatial_reference: request_spatial_ref,
+            },
+            GdalGeoTiffOptions {
+                compression_num_threads: get_config_element::<crate::util::config::Gdal>()?.compression_num_threads,
+                as_cog: self.info.as_cog,
+                force_big_tiff: false,
+            },
+            tile_limit,
+            Box::pin(futures::future::pending()), // datasets shall continue to be built in the background and not cancelled
```
jdroenner, line 110: can this cause never ending tasks?

services/src/datasets/create_from_workflow.rs
```diff
@@ -0,0 +1,268 @@
...
+            name: info.name,
+            description: info.description.unwrap_or_default(),
+            source_operator: "GdalSource".to_owned(),
+            symbology: None,  // TODO add symbology?
+            provenance: None, // TODO add provenance that references the workflow
+        },
+        meta_data,
+    };
+
+    // TODO: build pyramides, prefereably in the background
```
jdroenner, line 259: pyramides are build when data is stored as COG. I guess we don't need to do that as an additional task.

services/src/handlers/layers.rs
```diff
@@ -535,6 +542,100 @@ async fn layer_to_workflow_id_handler<C: Context>(
...
+        }
+    };
+
+    let session = C::Session::from(session.clone());
+
+    let execution_context = ctx.execution_context(session.clone())?;
+
+    let raster_operator = layer
+        .workflow
+        .clone()
```
jdroenner, line 592: do you really need to clone that workflow?

services/src/handlers/layers.rs
```diff
@@ -535,6 +542,100 @@ async fn layer_to_workflow_id_handler<C: Context>(
...
+        .get_raster()?
+        .initialize(&execution_context)
+        .await?;
+
+    let result_descriptor = raster_operator.result_descriptor();
+
+    let qr = QueryRectangle {
+        spatial_bounds: result_descriptor
+            .bbox
+            .ok_or(Error::InvalidLayerResultDescriptor {
```
jdroenner, line 603: we should discuss this. For me this is reasonable but the ResultDescriptor is not invalid since that fields are options.

jdroenner (changes requested):

services/src/datasets/create_from_workflow.rs
glombiewski: I did not write this code - I assume this was considered when Michael wrote the abortable query execution. If it wasn't, I can think about it.

services/src/datasets/create_from_workflow.rs
```diff
@@ -0,0 +1,268 @@
...
+
+impl<C: Context> RasterDatasetFromWorkflowTask<C> {
+    async fn process(&self) -> error::Result<RasterDatasetFromWorkflowResult> {
+        let operator = self.workflow.operator.clone();
+
+        let operator = operator.get_raster().context(crate::error::Operator)?;
+
+        let execution_context = self.ctx.execution_context(self.session.clone())?;
+        let initialized = operator
+            .clone()
```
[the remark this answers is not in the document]
glombiewski: No need to clone. 6fbb94da3d6e988441baf76f04cd58d715265832

services/src/datasets/create_from_workflow.rs
```diff
@@ -0,0 +1,268 @@
...
+            &self.file_path,
+            p,
+            query_rect,
+            query_ctx,
+            GdalGeoTiffDatasetMetadata {
+                no_data_value: Default::default(), // TODO: decide how to handle the no data here
+                spatial_reference: request_spatial_ref,
+            },
+            GdalGeoTiffOptions {
+                compression_num_threads: get_config_element::<crate::util::config::Gdal>()?.compression_num_threads,
```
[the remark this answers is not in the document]
glombiewski: The create_system_dataset handler uses get_config_element in the handler. So I moved that call to the workflows and layers handlers. 6fbb94da3d6e988441baf76f04cd58d715265832

services/src/handlers/layers.rs
glombiewski: I guess we can rename it to LayerResultDescriptorMissingFields or instead of an error have some sensible default values. I would need guidance for the latter.

services/src/datasets/create_from_workflow.rs
glombiewski: d9b7cbc73daa7c5b2990340dd37619fc3debdcac

services/src/handlers/layers.rs
glombiewski: I guess an alternative would be to initialize the raster operator and then use the initialized operator for the task processing. Would be a bit more refactoring and more boxes, but if that is better than cloning, I could do that.

services/src/handlers/layers.rs
jdroenner: lets call it "LayerResultDescriptorMissingFields" and then we need a new task / card where we change that a raster result descriptor always has a bound. Currently we know bounds in all cases i know off. If we really don't know the bounds we can use the bounds of the SRS as fallback.

glombiewski: 974d065ddd83e975f5edb25342c127f44007ba1a

jdroenner (approved):

jdroenner:
bors merge

jdroenner:
bors retry

jdroenner:
bors retry

jdroenner:
bors retry

jdroenner:
bors retry

jdroenner:
bors merge

jdroenner:
bors retry

jdroenner:
bors retry

jdroenner:
bors retry

jdroenner (approved):

jdroenner:
bors merge

PR #704, opened 2023-01-05
fix wms error format

michaelmattig:

---

<TEXT>


michaelmattig:
[Fix: cartopy WMS](https://trello.com/c/pImtg5PS/1163-fix-cartopy-wms)

services/src/handlers/wms.rs
```diff
@@ -186,9 +184,7 @@ where
             </GetMap>
         </Request>
         <Exception>
-            <Format>XML</Format>
-            <Format>INIMAGE</Format>
-            <Format>BLANK</Format>
+            <Format>application/json</Format>
```
jdroenner, line 187: is this a valid option?

michaelmattig: actually no, the standard says

- XML (mandatory)
- INIMAGE (optional)
- BLANK (optional)

default value is also XML.

We could have our frontend specify our "vendor specific" application/json and implement XML additionally. Or, like before, just ignore the parameter and always output json regardless. But I guess one of the two ist has to be in order for the examples to work again.

jdroenner: i wonder what happens if we don't declare on of the official options. XML is mandatory...
We should add json but maybe add some minimal form of xml? Do you know what the minimal requirement for XML would be?

michaelmattig: I added XML exception format: c270e3e3bc4419071d65f1e39fb0d416b9546c01

jdroenner (approved):

jdroenner:
cargo merge

jdroenner:
bors merge

PR #717, opened 2023-01-18
always read gdal offset scale params first

jdroenner:

---

<TEXT>


jdroenner:
[Read GDAL dedicated scale and offset values](https://trello.com/c/d4iqIfAX/1220-read-gdal-dedicated-scale-and-offset-values)

operators/src/processing/meteosat/mod.rs
```diff
@@ -101,27 +101,19 @@ mod test_util {
...
-                .properties_map
-                .insert(new_satellite_key(), RasterPropertiesEntry::Number(v.as_()));
+            props.insert_property(new_satellite_key(), RasterPropertiesEntry::Number(v.as_()));
         }
 
         if let Some(v) = slope {
-            props
-                .properties_map
-                .insert(new_slope_key(), RasterPropertiesEntry::Number(v));
+            props.insert_property(new_slope_key(), RasterPropertiesEntry::Number(v));
```
michaelmattig, line 112: shouldn't this set the `props.scale`?

operators/src/processing/meteosat/mod.rs
```diff
@@ -101,27 +101,19 @@ mod test_util {
...
         }
 
         if let Some(v) = offset {
-            props
-                .properties_map
-                .insert(new_offset_key(), RasterPropertiesEntry::Number(v));
+            props.insert_property(new_offset_key(), RasterPropertiesEntry::Number(v));
```
michaelmattig, line 116: shouldn't this set the `props.offset`?

operators/src/processing/raster_scaling.rs
```diff
@@ -185,8 +190,18 @@ where
         tile: RasterTile2D<P>,
         _pool: Arc<ThreadPool>,
     ) -> Result<RasterTile2D<P>> {
-        let offset = Self::prop_value(&self.offset, &tile.properties)?;
-        let slope = Self::prop_value(&self.slope, &tile.properties)?;
+        // either use the provided metadata/constant or the default values from the properties
```
michaelmattig, line 193: you could do this once in the initialization of the query processor, instead of for each tile, right?

operators/src/processing/meteosat/mod.rs
jdroenner: well thats an interesting question. You argued that we should separate the "real" GDAL 'offset' and 'scale' values from values that might be stored as properties. Lat time i checked the MSG driver exposes the "calibration_offset" and "calibration_slope" values as properties and not via the dedicated 'offset' 'scale' values. So i think we should keep it like this since this is what will be used with MSG data.

operators/src/processing/meteosat/mod.rs
jdroenner: see above

operators/src/processing/raster_scaling.rs
jdroenner: i guess this is a bit problematic. At least the 'Scale' and 'Offset' values might be different for different files. So we can't do this on a global time-series level. Also the data is parsed from the real GDAL datasets and that is done when the data is read. We might add information about the metadata keys to expect to the result descriptor but thats something different.

michaelmattig (approved):

michaelmattig:
bors merge

PR #721, opened 2023-01-20
check quota and allow updating it

michaelmattig:

---

<TEXT>


michaelmattig:
[Credit-System: User Credits Einsatz](https://trello.com/c/b95KAN0M/1178-credit-system-user-credits-einsatz)

services/src/handlers/plots.rs
```diff
@@ -179,7 +181,10 @@ async fn get_plot_handler<C: Context>(
         });
     };
 
-    let processor = initialized.query_processor().context(error::Operator)?;
+    let processor = execution_context
+        .create_plot_query_processor(initialized)
+        .await
+        .context(error::Operator)?;
```
ChristianBeilschmidt, line 187: 1. What happens if we call it like before?
2. Why is it wrapped here, but all other methods just take a reference to the execution context?

services/src/pro/util/tests.rs
```diff
@@ -296,3 +310,87 @@ pub(in crate::pro) mod mock_oidc {
         result
     }
 }
+
+#[allow(clippy::missing_panics_doc)]
+pub async fn register_ndvi_workflow_helper(ctx: &ProInMemoryContext) -> (Workflow, WorkflowId) {
```
ChristianBeilschmidt, line 315: Just as a thought:
Should we only compile this if(test)?

Maybe clippy would not bother then about the missing panics doc.
And we could not use it apart from test.

services/src/pro/users/postgres_userdb.rs
```diff
@@ -66,17 +66,23 @@ where
 
         let stmt = tx
             .prepare(
-                "INSERT INTO users (id, email, password_hash, real_name, active) VALUES ($1, $2, $3, $4, $5);",
+                "INSERT INTO users (id, email, password_hash, real_name, quota_available, active) VALUES ($1, $2, $3, $4, $5, $6);",
```
ChristianBeilschmidt, line 69: If there is a quote no matter what is specified if it is used, shouldn't we set it explicitly to 0 in the settings. Currently, it is commented out.

services/src/pro/users/hashmap_userdb.rs
```diff
@@ -56,6 +58,14 @@ impl UserDb for HashMapUserDb {
 
         self.users.write().await.insert(id.to_string(), user);
 
+        let mut quota_available = self.quota_available.write().await;
+        quota_available.insert(
+            id,
+            crate::util::config::get_config_element::<crate::pro::util::config::User>()?
+                .default_available_quota
+                .unwrap_or(0),
```
ChristianBeilschmidt, line 66: maybe already deserliaze it as 0 if unset

services/src/pro/contexts/mod.rs
```diff
@@ -137,6 +183,30 @@ where
...
+        operator.query_processor()
+    }
+
+    async fn create_plot_query_processor(
+        &self,
+        operator: Box<dyn InitializedPlotOperator>,
+    ) -> geoengine_operators::util::Result<TypedPlotQueryProcessor> {
+        check_quota(&*self.user_db, &self.session).await?;
+        operator.query_processor()
+    }
```
ChristianBeilschmidt, line 208: Is this the right place?
Shouldn't it be checked when we call `query`?

ChristianBeilschmidt (changes requested):

services/src/handlers/plots.rs
michaelmattig: 1. If you create the processor from the InitializedOperator, like before, then no quota check is performed. This here is basically the point where, in Pro, the quota is checked, and the execution prevented if it is less than zero. In free, no check is performed, because there is no quota.

2. I did it this way to implement a different behavior, depending on the context being Pro or Free. That#s why `create_plot_query_processor` is a trait method with different implementations depending on the type of context.

I am sure there are plenty other ways to implement the quota check. E.g. duplicating the WMS handler code in Pro, and including a check there. But I found my solution to be the easiest, least intrusive and most effective.

services/src/pro/util/tests.rs
michaelmattig: Stackoverflow says that the Rust compiler doesn't include unused methods in the binary. So I don't think we need to do this. But we could.

services/src/pro/users/postgres_userdb.rs
michaelmattig: Right now, if no quota is set, it would just default to zero. If we make it non-optional and no quota is set, geo engine will crash at runtime instead, when the first workflow is run. If the latter is our preferred behavior, we can make it non-optional. I thought taking 0 made sense.

services/src/pro/users/hashmap_userdb.rs
michaelmattig: would also be possible but leaving it this way is more flexible right? Because we still know whether it was set to 0 explicitly or left blank.

services/src/pro/contexts/mod.rs
michaelmattig: During initialization of the query processor we already do some work, like computing ResultDescriptors, initializing models and idk what. So why would we want to do that work and only abort later when finally query is called?

services/src/pro/util/tests.rs
ChristianBeilschmidt: No, this is clear to me. It's just that clippy says, okay, this is a regular method, so why do you just panic?
And it is, we could use it, but we use it only in test. Wondering if we could just enclose it in #[cfg(test)] ?

services/src/pro/contexts/mod.rs
ChristianBeilschmidt: Wouldn't you want to run a query as long as you have resources and not to be prevented to run the first query although you would have had resources at least for that?

services/src/pro/users/hashmap_userdb.rs
ChristianBeilschmidt: But this leaves the default value to each implementation, right? It is sufficient, but I just wanted to see if it is actually meant to be.

services/src/pro/util/tests.rs
michaelmattig: how about we make the whole file #cfg[test]?

services/src/pro/contexts/mod.rs
michaelmattig: that's how it currently works: If you have at least 1 credit, you can create the query processor and query it as often as you want.

If we want to check every query-call, we have to do things differently. Like introducing a wrapper operator that checks this in the query-method before delegating it to the real workflow.

services/src/pro/users/postgres_userdb.rs
```diff
@@ -436,7 +451,13 @@ where
     async fn increment_quota_used(&self, user: &UserId, quota_used: u64) -> Result<()> {
         let conn = self.conn_pool.get().await?;
         let stmt = conn
-            .prepare("UPDATE users SET quota_used = quota_used + $1 WHERE id = $2;")
+            .prepare(
+                "
+            UPDATE users SET 
+                quota_available = quota_available - $1,
```
ChristianBeilschmidt, line 457: add a test where the quota_available gets negative after a query

services/src/handlers/wms.rs
```diff
@@ -316,7 +316,10 @@ async fn wms_map_handler<C: Context>(
             Box::new(irp)
         };
 
-        let processor = initialized.query_processor().context(error::Operator)?;
+        let processor = execution_context
```
ChristianBeilschmidt, line 319: save_dataset and maybe others (?) are missing

ChristianBeilschmidt (changes requested):

Settings-test.toml
```diff
@@ -28,3 +28,7 @@ client_id = ""
 client_secret = ""
 redirect_uri = ""
 scopes = [ ]
+
+[user]
+quota_check = true
+default_available_quota = 0
```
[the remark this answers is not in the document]
michaelmattig: 27586db60cfc6f20d0938060f84902b2f3d96ff1

services/src/pro/util/tests.rs
michaelmattig: 27586db60cfc6f20d0938060f84902b2f3d96ff1

services/src/handlers/wms.rs
michaelmattig: is now automatically checked because every operator is wrapped: 10bc0646cd8e6938210e1ff24a11e9124fd983af

services/src/pro/users/postgres_userdb.rs
michaelmattig: bf8442899e6d3725431a335c3cf664c008a11173

services/src/pro/contexts/mod.rs
michaelmattig: ok so I built it into the query-method now. We ultimately need it there, because otherwise one could create a processor with 1 credit and use it forever because the quota is not checked again.

So right now, every processor in the chain will check the quota. We discussed caching it instead. The cache would be in the query context. Right now, this is not possible because the query context can be re-used for subsequent queries.

I would propose to make the QueryContext non-reusable. So it should be consumed by the query-method. Then we can cache the quota there, and, we can also have something like a unique query id stored in the context that we can use for stuff like logging and quota tracking.

In order for the QueryContext to be consumed, we have to change the parameter in the query method from a fat pointer to a box though. And: for operators with more than one child, we would need to clone the QueryContext, in oder to put it into each child's query method.

operators/src/error.rs
```diff
@@ -349,6 +353,13 @@ pub enum Error {
         expected: usize,
         found: usize,
     },
+
+    #[cfg(feature = "pro")]
+    QuotaExhausted,
```
ChristianBeilschmidt, line 358: I guess we said we did not want to have `cfg = pro` all around the code. Could we do something about this?

operators/src/pro/meta/quota.rs
```diff
@@ -34,3 +36,10 @@ impl QuotaTracking {
         let _ = self.quota_sender.send(self.computation); // ignore the Result because the quota receiver should never close the receiving end of the channel
     }
 }
+
+#[async_trait]
+pub trait QuotaCheck {
+    async fn check_quota(&self) -> Result<bool>;
```
ChristianBeilschmidt, line 42: It should be more verbose what it is doing. For instance, `has_quota_available` would be an option.

Maybe you can add some doc-comments as well.

services/src/pro/contexts/mod.rs
```diff
@@ -137,6 +184,30 @@ where
         op
     }
 
+    async fn create_raster_query_processor(
+        &self,
+        operator: Box<dyn InitializedRasterOperator>,
+    ) -> geoengine_operators::util::Result<TypedRasterQueryProcessor> {
+        check_quota(&*self.user_db, &self.session).await?;
```
ChristianBeilschmidt, line 191: It is checked here and injected into the StatisticsProcessor? Why is this necessary?

services/src/pro/contexts/mod.rs
```diff
@@ -312,6 +393,28 @@ where
...
+            .user_db
+            .quota_available(&self.session)
+            .await
+            .map_err(
+                |e| geoengine_operators::error::Error::QueryingProcessorFailed {
+                    source: Box::new(e),
+                },
+            )?
+            > 0)
+    }
```
ChristianBeilschmidt, line 415: Do we need to check (or start doing this…) performance degredations?
This way, we could decide to work on this now or… not.

services/src/pro/handlers/drone_mapping.rs
```diff
@@ -422,6 +423,11 @@ mod tests {
         let ctx = ProInMemoryContext::test_default();
         let session = create_session_helper(&ctx).await;
 
+        ctx.user_db_ref()
+            .update_quota_available_by_user(&session.user.id, 9999)
+            .await
+            .unwrap();
```
ChristianBeilschmidt, line 429: Should we really put this into this test? It is not really part of what we want to test.

ChristianBeilschmidt (changes requested):

services/src/pro/contexts/mod.rs
ChristianBeilschmidt: I guess it is already designed this way, so we could do that.

services/src/pro/handlers/drone_mapping.rs
michaelmattig: by default, in all tests, the available quota is zero. We could change it and have by default 9999 and only set it to zero in the tests about quotas

services/src/pro/contexts/mod.rs
```diff
@@ -54,55 +57,99 @@ pub trait ProContext: Context<Session = UserSession> {
...
+        geoengine_operators::error::Error::CreatingProcessorFailed {
+            source: Box::new(e),
+        }
+    })?;
+
+    if quota_available <= 0 {
+        return Err(geoengine_operators::error::Error::CreatingProcessorFailed {
+            source: Box::new(crate::error::Error::QuotaExhausted),
+        });
+    }
```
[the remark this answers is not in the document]
michaelmattig: no, if zero or less is available, it is exhausted

services/src/pro/contexts/mod.rs
michaelmattig: the problem here is, if the QueryContext is cloneable, there is no benefit in it being consumed by the query method. You can just clone it before you put it in. So we would somehow have to invalidate it after being put into a query-method, but put it in again, for all the child processors.

I guess then the best idea would be to avoid wrapping every single processor with the quota check, but only wrap the top-level one. But how can the execution context distinguish the top-level operator from the other ones? Maybe we have to do the top-level wrapping in a different place.

One idea would be to do it in the `Workflow`. Right now, we access the top-level operator by `workflow.operator` directly. We could make `Workflow` into a trait with a method `operator()` which, in free just outputs the operator but in pro it wraps it with a quota checker.

services/src/pro/contexts/mod.rs
michaelmattig: it was leftover from the previous attempt. I removed it

services/src/pro/contexts/mod.rs
michaelmattig: I performed a benchmark (see my Trello comment)

>     Starting benchmark, quota check enabled: false
    Run 0 time 249.679598ms
    Run 1 time 242.757376ms
    Run 2 time 243.797694ms
    Run 3 time 247.024981ms

>    Starting benchmark, quota check enabled: true
    Run 0 time 255.951031ms
    Run 1 time 251.298686ms
    Run 2 time 251.200896ms
    Run 3 time 245.754991ms

services/src/pro/handlers/users.rs
```diff
@@ -276,18 +282,21 @@ pub(crate) async fn quota_handler<C: ProContext>(
     ctx: web::Data<C>,
     session: C::Session,
 ) -> Result<impl Responder> {
-    let quota_used = ctx.user_db_ref().quota_used(&session).await?;
-    Ok(web::Json(QuotaUsed { used: quota_used }))
+    let available = ctx.user_db_ref().quota_available(&session).await?;
+    let used = ctx.user_db_ref().quota_used(&session).await?;
+
+    Ok(web::Json(Quota { available, used }))
```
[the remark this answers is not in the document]
michaelmattig: for this to work, Quota should have to implement Responder, which it doesn't

>   .route(web::get().to(get_user_quota_handler::<C>))
    |                                   -- ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `actix_web::Responder` is not implemented for `pro::handlers::users::Quota`
    |                                   |
    |                                   required by a bound introduced by this call

services/src/pro/handlers/users.rs
```diff
@@ -309,13 +318,52 @@ pub(crate) async fn user_quota_handler<C: ProContext>(
...
+                source: Box::new(error::Error::OperationRequiresAdminPrivilige),
             });
         }
     }
 
-    let quota_used = ctx.user_db_ref().quota_used_by_user(&user).await?;
-    Ok(web::Json(QuotaUsed { used: quota_used }))
+    let available = ctx.user_db_ref().quota_available_by_user(&user).await?;
+    let used = ctx.user_db_ref().quota_used_by_user(&user).await?;
+    Ok(web::Json(Quota { available, used }))
```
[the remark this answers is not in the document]
michaelmattig: for this to work, Quota should have to implement Responder, which it doesn't

>   .route(web::get().to(get_user_quota_handler::<C>))
    |                                   -- ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `actix_web::Responder` is not implemented for `pro::handlers::users::Quota`
    |                                   |
    |                                   required by a bound introduced by this call

services/src/pro/handlers/users.rs
```diff
@@ -309,13 +318,52 @@ pub(crate) async fn user_quota_handler<C: ProContext>(
...
+) -> Result<impl Responder> {
+    let user = user.into_inner();
+
+    let update = update.into_inner();
+
+    ctx.user_db_ref()
+        .update_quota_available_by_user(&user, update.available)
+        .await?;
+
+    Ok(actix_web::HttpResponse::Ok())
```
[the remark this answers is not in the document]
michaelmattig: for this to work, Quota should have to implement Responder, which it doesn't

>   .route(web::get().to(get_user_quota_handler::<C>))
    |                                   -- ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `actix_web::Responder` is not implemented for `pro::handlers::users::Quota`
    |                                   |
    |                                   required by a bound introduced by this call

operators/src/error.rs
michaelmattig: I removed the feature flag. It doesn't hurt to have the enum variant in non-pro and we do it this way all the time, e.g. with permission exceptions in the services crate. 3c7babbdf87ddc47574187df72dcdb57a808070b

operators/src/pro/meta/statistics.rs
```diff
@@ -155,7 +164,7 @@ where
         match stream_result {
             Ok(stream) => {
                 tracing::debug!(event = "query ok");
-                Ok(StreamStatisticsAdapter::new(stream, span.clone(), quota).boxed())
+                Ok(StreamStatisticsAdapter::new(stream, span.clone(), quota_tracker).boxed())
```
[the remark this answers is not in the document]
michaelmattig: I renamed the operator and processor to a generic name. 3c7babbdf87ddc47574187df72dcdb57a808070b

services/src/pro/contexts/mod.rs
```diff
@@ -54,55 +57,99 @@ pub trait ProContext: Context<Session = UserSession> {
...
             layer_provider_db,
+            user_db,
             thread_pool,
             session,
             tiling_specification,
         }
     }
 }
 
+async fn check_quota<U: UserDb>(
```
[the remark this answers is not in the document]
michaelmattig: the method was redundant and I removed it.

operators/src/error.rs
ChristianBeilschmidt: But now there is a QuotaError also for non-pro, right?

operators/src/pro/meta/quota.rs
ChristianBeilschmidt: I don't see in in this change, maybe you added the wrong one?

services/src/pro/handlers/drone_mapping.rs
ChristianBeilschmidt: That sounds about right. I don't know about the 9999 part. Is this the maximum quote we can think of? :smile:

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Okay, but shouldn't we implement that then? Using it as some generic responder doesn't make this better, right?

services/src/pro/contexts/mod.rs
ChristianBeilschmidt: So we leave it there for now?

operators/src/error.rs
michaelmattig: Please make a suggestion then.

operators/src/pro/meta/quota.rs
michaelmattig: a845e97cc0954aa90fb5c2acd53d1cda62c98a5f

services/src/pro/contexts/mod.rs
michaelmattig: I'd say so

services/src/pro/handlers/drone_mapping.rs
michaelmattig: it's big enough for the tests and short enough to be read

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
michaelmattig: > Okay, but shouldn't we implement that then? Using it as some generic responder doesn't make this better, right?

I wouldn't want to implement this trait for every single struct we return. What's the problem with web::Json all of a sudden? We use it like this all the time.

operators/src/error.rs
ChristianBeilschmidt: I don't understand this. I always advocate not mixing the pro/non-pro code.

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: This is why I suggested using `web::Json<Quota>`.

services/src/pro/handlers/drone_mapping.rs
ChristianBeilschmidt: If it should never fail, maybe int max?

operators/src/error.rs
michaelmattig: > I don't understand this. I always advocate not mixing the pro/non-pro code.

I'm all for that

> This is the only #[cfg(feature = "pro")] that was there.

true. That's why I removed it. I just meant that there are other errors, that are only ever used when a feature is activated (e.g. pro) that are not conditionally compiled.

> What about putting the QuotaError inside the QueryingProcessorFailed or something similar?

that's what I meant here: " I have no other solution here, other than using a generic Box error.". So a new variant `QueryingProcessorFailed { source: Box<dyn Error> }` to hide the real error from pro.

Or do you have a better idea?

services/src/pro/handlers/drone_mapping.rs
michaelmattig: that's a bulky number, but if you insist I can change it. which int would you like? 32 or 64 bit? signed or unsigned?

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
michaelmattig: 58cd9244c5445fc15d50c454f0863b11d7d0d7ed

services/src/pro/handlers/users.rs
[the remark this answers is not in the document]
michaelmattig: 58cd9244c5445fc15d50c454f0863b11d7d0d7ed

operators/src/error.rs
ChristianBeilschmidt: No, I guess we have to do it this way.

We just have to get sure that we still get a reasonable error out.

michaelmattig: 89751bd532aead2d2f0d99f81480fed7038de92d

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #732, opened 2023-01-31
custom retry mechanism for gdal source

michaelmattig:

---

<TEXT>


michaelmattig:
[Resilienter S2 Provider: Checkpoint / Retry](https://trello.com/c/EXRu68RP/1143-resilienter-s2-provider-checkpoint-retry)

michaelmattig:
[Resilienter S2 Provider: Checkpoint / Retry](https://trello.com/c/EXRu68RP/1143-resilienter-s2-provider-checkpoint-retry)

datatypes/src/util/mod.rs
```diff
@@ -5,6 +5,7 @@ pub mod helpers;
 pub mod identifiers;
 pub mod ranges;
 mod result;
+pub mod retry;
```
ChristianBeilschmidt, line 8: Shoudln't this rather be part of operators?

(Then, we could remove futures and tokio from datatypes again.)

operators/src/source/gdal_source/mod.rs
```diff
@@ -498,14 +549,22 @@ impl GdalRasterLoader {
...
+        if let Err(error) = &dataset_result {
+            let is_file_not_found = matches!(
+                error,
+                Error::Gdal {
+                    source: GdalError::NullPointer {
+                        method_name,
+                        msg
+                    },
+                } if *method_name == "GDALOpenEx" && (*msg == "HTTP response code: 404" || msg.ends_with("No such file or directory"))
+            );
```
ChristianBeilschmidt, line 561: I guess it is wise to write tests for that behavior.

operators/src/source/gdal_source/mod.rs
```diff
@@ -420,11 +434,48 @@ impl GdalRasterLoader {
...
+        retry(
+            dataset_params
+                .retry
+                .map(|r| r.max_retries)
+                .unwrap_or_default(),
+            GDAL_RETRY_INITIAL_BACKOFF_MS,
+            GDAL_RETRY_EXPONENTIAL_BACKOFF_FACTOR,
+            move || {
+                let ds = dataset_params.clone();
+                let file_path = ds.file_path.clone();
```
ChristianBeilschmidt, line 448: Is cloning necessary? Maybe because it is "reused"? If not, we could remove that.

operators/src/source/gdal_source/mod.rs
```diff
@@ -420,11 +434,48 @@ impl GdalRasterLoader {
...
+                            if is_vsi_curl {
+                                // clear the VSICurl cache, to force GDAL to try to re-download the file
+                                // otherwise it will assume any observed error will happen again
+                                unsafe {
+                                    if let Some(Some(c_string)) =
+                                        file_path.to_str().map(|s| CString::new(s).ok())
+                                    {
+                                        VSICurlPartialClearCache(c_string.as_ptr());
+                                    }
+                                }
```
ChristianBeilschmidt, line 469: 1. This should maybe be part of rust-gdal.
2. Can we test this?

operators/src/source/gdal_source/mod.rs
```diff
@@ -420,11 +434,48 @@ impl GdalRasterLoader {
         tile_information: TileInformation,
         tile_time: TimeInterval,
     ) -> Result<RasterTile2D<T>> {
-        crate::util::spawn_blocking(move || {
-            Self::load_tile_data(&dataset_params, tile_information, tile_time)
-        })
+        let is_vsi_curl = dataset_params.file_path.starts_with("/vsicurl/");
```
ChristianBeilschmidt, line 437: What about `/vsicurl_streaming/` or `/vsizip//vsicurl/` ?

ChristianBeilschmidt (changes requested):

operators/src/source/gdal_source/mod.rs
michaelmattig: 1. it could be, but I didn't want to force Johannes to make a release to save half a line in Geo Engine
2. it's tested implicitly in the sentinel2 provider. The fact that it runs through, shows, that gdal retries. I don't really think it makes sense to write a test that ensures that it doesn't work without this line.

operators/src/source/gdal_source/mod.rs
michaelmattig: I could do `file_path.contains("/vsicurl")` but then it would detect a local folder name vsicurl also as being a remote file. So should I add your two examples explicitly or do we need a gdal file path parser that extracts the virtual file systems from a path?

operators/src/source/gdal_source/mod.rs
michaelmattig: I did not find a way to avoid cloning here. Let me know if you have any ideas.

datatypes/src/util/mod.rs
michaelmattig: 92f3e595cc959be5f89d1faaf008eb4179ef9852

operators/src/source/gdal_source/mod.rs
michaelmattig: 3b007f8b0f53c02c061129542fef22082e452973

operators/src/source/gdal_source/mod.rs
ChristianBeilschmidt: But shouldn't this offer a good test? You basically inserted this because there are certain errors, where GDAL would return the same result even if the source changed. And this line avoids that, right?

michaelmattig: b8829790a8bbe5ed8debae4a36e920ab5ec93cea

operators/src/source/gdal_source/mod.rs
michaelmattig: a80d47091639d8e20e4c5ef664b1a0c6bd751082

ChristianBeilschmidt (approved):

operators/src/source/gdal_source/mod.rs
```diff
@@ -47,14 +50,19 @@ use serde::{Deserialize, Serialize};
...
 use std::marker::PhantomData;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::time::Instant;
 use tracing::{span, Level};
 
 mod error;
 mod loading_info;
 
+static GDAL_RETRY_INITIAL_BACKOFF_MS: u64 = 1000;
```
ChristianBeilschmidt, line 62: Is there a reason why to start with one second?

ChristianBeilschmidt (changes requested):

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #734, opened 2023-02-08
Gbif provider fix: Column names not in quotation marks

dbrandenstein:

---

Column names were not put in quotation marks everywhere, so queries with column name "order" failed.


ChristianBeilschmidt:
bors merge

PR #753, opened 2023-03-01
Connection to the Aruna Object Storage

glombiewski:

---

- Changed the dependency from ScienceObjectDb to Aruna Object Storage
- Changed the expected object mapping to have a dedicated meta object within the Aruna Object Storage that is part of the same object group as the data object
- Removed wiremock-grpc test dependency, due to incompatibility with handling of multiple services
- Provided custom mocking for grpc server


services/src/datasets/external/nfdi/mod.rs
```diff
@@ -504,25 +579,34 @@ impl MetaDataProvider<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectan
...
+            .get_single_file_object(&aruna_dataset_ids)
+            .await
+            .map_err(|e| geoengine_operators::error::Error::DatasetMetaData {
+                source: Box::new(e),
+            })?;
+
+        match &meta_data.data_type {
             DataType::SingleRasterFile(info) => {
-                let result_descriptor = Self::create_raster_result_descriptor(md.crs.into(), info);
+                let result_descriptor =
```
jdroenner, line 603: you can set the ResultDescriptor resolution and bbox from the rasters GeoTransform and with/height (both are in the RasterInfo).
if aruna can only store single raster file datasets, the `time_interval` in the `RasterInfo` is the temporal bound?

services/src/datasets/external/nfdi/mod.rs
```diff
@@ -458,25 +523,35 @@ impl MetaDataProvider<OgrSourceDataset, VectorResultDescriptor, VectorQueryRecta
...
+            .get_single_file_object(&aruna_dataset_ids)
+            .await
+            .map_err(|e| geoengine_operators::error::Error::DatasetMetaData {
+                source: Box::new(e),
+            })?;
+
+        match meta_data.data_type {
             DataType::SingleVectorFile(info) => {
-                let result_descriptor = Self::create_vector_result_descriptor(md.crs.into(), &info);
+                let result_descriptor =
```
jdroenner, line 548: please set the temporal bound of the ResultDescriptor if i is set in the VectorInfo. Can we add (optional) spatial bounds to VectorInfo to set the spatial bounds of the ResultDescriptor?

jdroenner (changes requested):

services/src/datasets/external/nfdi/mod.rs
glombiewski: 92da6f78a46dc3666f36e9a15225addf43a6c333

On a closer look, RasterInfo is kinda redundant with RasterResultDescriptor itself. It would probably be wise to consolidate them somehow, but given the deadline, I won't.

services/src/datasets/external/nfdi/mod.rs
glombiewski: 92da6f78a46dc3666f36e9a15225addf43a6c333

I added both time and spatial bounds to VectorInfo. Time was also necessary, because temporalExtend are the fields used for time, not the actual time itself. Similar to RasterInfo, VectorInfo is almost the same as a VectorResultDescriptor now, but I won't do that now.

jdroenner (approved):

jdroenner:
bors merge

PR #759, opened 2023-03-10
Cargo Sparse Checkout

ChristianBeilschmidt:

---

Cargo allows for sparse checkout of dependencies. This should make the build process faster, since only what is necessary is fetched from the registry. This will become the default in some versions.


ChristianBeilschmidt:
It seems it would require some extra nightly flags for now.

PR #765, opened 2023-03-23
move user input validation into deserialization

michaelmattig:

---

<TEXT>


services/src/api/model/services.rs
```diff
@@ -70,6 +69,7 @@ pub struct MetaDataSuggestion {
 
 #[derive(Debug, Serialize, Deserialize, Clone, ToSchema)]
 #[serde(rename_all = "camelCase")]
+// TODO: validate user input
```
ChristianBeilschmidt, line 72: I'm not sure if the TODO makes sense. You could only check for certain types of input errors, right? So you could check if the name is too long but not if it violates a duplicate contraint.

services/src/projects/project.rs
```diff
@@ -459,24 +460,15 @@ impl Default for ProjectFilter {
     }
 }))]
 pub struct CreateProject {
+    #[validate(length(min = 1))]
+    #[validate(length(max = 256))]
     pub name: String,
     pub description: String,
```
ChristianBeilschmidt, line 466: You previously checked for the description of not being empty, but you do not check this right now.

services/src/util/extractors.rs
```diff
@@ -0,0 +1,165 @@
...
+                                message: validation_errors_to_string(&validation_errors),
+                            }
+                            .into()
+                        })
+                }
+                Err(e) => Err(handle_json_payload_error(e)),
+            })
+            .boxed_local()
+    }
+}
```
ChristianBeilschmidt, line 68: What is the idea behind not wrapping `web::Json`?

You could say 

```rust
pub struct ValidatedJson<T>(web::Json<T>);
```

Then, you could just forward `FromRequest` and then call validated, right?

services/test.http
```diff
@@ -1,9 +1,9 @@
-POST http://localhost:3030/user
+POST http://localhost:3030/api/user
 Content-Type: application/json
 
 {
-  "email": "foo@example.com",
-  "password": "secret123",
+  "email": "fooxample.com",
+  "password": "se3",
```
ChristianBeilschmidt, line 6: Did you intend to commit this?

ChristianBeilschmidt (changes requested):

services/src/api/model/services.rs
michaelmattig: Dataset names do not have to be unique. It doesn't really make sense to limit this globally, and as users can share datasets, there will be name clashes anyway.

Also we could limit the number of breakpoints in a symbology here etc. So I think it would make sense to validate here. As the PR is about getting the Task list working again, I didn't want to start completing our user input validation here though.

services/src/util/extractors.rs
michaelmattig: the idea of forwarding to web::Json FromRequest is nice. However, the method returns a future, but we cannot await this future here, because we are not in an async function.

Furthermore, the `pub struct ValidatedJson<T>(pub T);` just wraps the `T` value for returning it to the users and it is constructed in the method. There is no benefit in wrapping a web::Json here, because the user wants the `T` value and the parsing and validation happens beforehand.

services/src/util/extractors.rs
```diff
@@ -0,0 +1,165 @@
...
+
+/// A Query extractor that validates the content after deserialization
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+pub struct ValidatedQuery<T>(pub T);
```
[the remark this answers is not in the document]
michaelmattig: cf my other comment

services/src/projects/project.rs
michaelmattig: 5cb7a0f1d944f31e6f219bd04f3ecf3115f4280a

services/test.http
michaelmattig: 5cb7a0f1d944f31e6f219bd04f3ecf3115f4280a

services/src/util/extractors.rs
ChristianBeilschmidt: You also return a `Future`, so you can just await, map, etc. I do not see this point.

The nice thing would just be to get the `actix` implementation and all updates automatically.

The compiler should optimize the nested newtypes.

services/src/util/extractors.rs
```diff
@@ -0,0 +1,165 @@
...
+                    .into()
+                })
+            })
+            .map_or_else(futures::future::err, |value| {
+                futures::future::ok(ValidatedQuery(value))
+            })
+    }
+}
+
+// TODO: Validated Extractors for Path and Form
```
ChristianBeilschmidt, line 148: Don't we use `Path` a lot?

ChristianBeilschmidt (changes requested):

services/src/util/extractors.rs
michaelmattig: Yes, but we only use it for IDs (UUIDs), I think. There is not really much to validate here apart from the format, which is already checked.

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #769, opened 2023-03-27
temp aggregate should be valid for subquery window

jdroenner:

---

<TEXT>

### DIFF
--- a/operators/src/adapters/sparse_tiles_fill_adapter.rs
+++ b/operators/src/adapters/sparse_tiles_fill_adapter.rs
@@ -340,8 +340,9 @@ where
                         }
                         // 4. The received tile has a TimeInterval that starts after the current TimeInterval and is not directly connected to the current TimeInterval.
                         else {
+                            debug_assert!(this.sc.current_time.end() <= tile.time.start());
                             // if the current_idx is the first in a new grid run then it is the first one with a new TimeInterval.
-                            // We need to generate a fill TimeInterval since current and tile TimeInterval are not connedted.
+                            // We need to generate a fill TimeInterval since current and tile TimeInterval are not connected.
                             if this.sc.current_idx_is_first_in_grid_run() {
                                 this.sc.current_time = TimeInterval::new(
                                     this.sc.current_time.end(),
--- a/operators/src/processing/temporal_raster_aggregation/subquery.rs
+++ b/operators/src/processing/temporal_raster_aggregation/subquery.rs
@@ -22,7 +22,7 @@ pub async fn subquery_all_tiles_fold_fn<P: Pixel, F: TemporalRasterPixelAggregat
 ) -> Result<TileAccumulator<P, F>> {
     crate::util::spawn_blocking_with_thread_pool(accu.pool.clone(), || {
         let mut accu = accu;
-        accu.add_tile(tile)?;
+        accu.add_tile(tile);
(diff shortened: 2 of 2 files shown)


jdroenner:
[Temporal Aggregation erzeugt Tiles, die nicht das gesamte Agg Window abdecken.](https://trello.com/c/NuKYr9J0/1276-temporal-aggregation-erzeugt-tiles-die-nicht-das-gesamte-agg-window-abdecken)

operators/src/adapters/sparse_tiles_fill_adapter.rs
```diff
@@ -340,8 +340,9 @@ where
                         }
                         // 4. The received tile has a TimeInterval that starts after the current TimeInterval and is not directly connected to the current TimeInterval.
                         else {
+                            debug_assert!(this.sc.current_time.end() <= tile.time.start());
```
michaelmattig, line 343: what if the tile is the first in the grid run and there is actually a gap after the previous one? Then this would not hold. Actually the comment in 4 states exactly this.

operators/src/processing/temporal_raster_aggregation/subquery.rs
```diff
@@ -44,16 +44,24 @@ where
     P: Pixel,
     F: TemporalRasterPixelAggregator<P> + 'static,
 {
-    pub fn add_tile(&mut self, in_tile: RasterTile2D<P>) -> Result<()> {
-        self.time = self.time.union(&in_tile.time)?;
+    pub fn add_tile(&mut self, in_tile: RasterTile2D<P>) {
+        // we do no time modification here since the result represents the aggreate over query time
```
michaelmattig, line 48: this comment doesn't really make sense without knowing that we erroneously did it before right? Maybe instead describe what the method does here? 

Perhaps something like this:

"Add a tile to the accumulator, which represents the aggregate over a time interval that contains the in_tile. TODO: for tiles which are only partially contained in the aggregate time interval, investigate whether the pixels have to be scaled (e.g. if the pixel is a count, assume uniform distribution and divide it by the fraction of time that is contained in the aggregate)".

operators/src/adapters/sparse_tiles_fill_adapter.rs
jdroenner: yes it does. `sc` is the state container so the time in the container aka the latest time so far must always be smaller then the time of the newly arrived `tile`.
"4." states that "The received tile has a TimeInterval that starts after the current TimeInterval" so it must hold.

operators/src/processing/temporal_raster_aggregation/subquery.rs
jdroenner: [eebc485](https://github.com/geo-engine/geoengine/pull/769/commits/eebc485ccf9b5342e0401e74ca701799763a3a39)

operators/src/adapters/sparse_tiles_fill_adapter.rs
jdroenner: yes it does. sc is the state container so the time in the container aka the latest time so far must always be smaller then the time of the newly arrived tile.
"4." states that "The received tile has a TimeInterval that starts after the current TimeInterval" so it must hold.

michaelmattig:
bors merge

PR #770, opened 2023-03-29
Fix_postgres

michaelmattig:

---

<TEXT>


michaelmattig:
[Warnings beim Start mit Postgres-Backend (Permissions)](https://trello.com/c/k9G68Qmx/1279-warnings-beim-start-mit-postgres-backend-permissions)

services/Cargo.toml
```diff
@@ -97,13 +97,14 @@ zip = "0.6"
 nix = "0.26"
 
 [dev-dependencies]
+assert_cmd = "2.0"
 httptest = "0.15.2"
+hyper = "0.14.24"
+prost = "0.11.6"
```
ChristianBeilschmidt, line 103:
```suggestion
hyper = "0.14"
prost = "0.11"
```

Could you omit the last digit? It doesn't matter anyway in this config.

services/tests/startup.rs
```diff
@@ -0,0 +1,139 @@
...
+    .await
+    .unwrap();
+
+    // The connection object performs the actual communication with the database,
+    // so spawn it off to run on its own.
+    tokio::spawn(async move {
+        if let Err(e) = connection.await {
+            eprintln!("connection error: {e}");
+        }
+    });
```
ChristianBeilschmidt, line 76: Unfortunately, I still don't get why you are doing this.

services/tests/startup.rs
```diff
@@ -0,0 +1,139 @@
...
+
+            // println!("{buf}");
+
+            assert!(!buf.contains("WARN"));
+
+            if buf.contains("Tokio runtime found") {
+                startup_succesful = true;
+                break;
+            }
+        }
```
ChristianBeilschmidt, line 44: Can't you read after you've shutdown the server?

ChristianBeilschmidt: Or… use async stderr from tokio:
https://docs.rs/tokio/1.27.0/tokio/io/fn.stderr.html

ChristianBeilschmidt (changes requested):
There is an error in the tests.

services/tests/startup.rs
michaelmattig: this is from the example on how to connect to postgres https://docs.rs/tokio-postgres/latest/tokio_postgres/

services/tests/startup.rs
michaelmattig: no. How should I know when I can shutdown the server without reading the logs?

why should I use the stderr from tokio? I think it's ok to block here.

ChristianBeilschmidt: You are already in an async function, so just await?

services/tests/startup.rs
ChristianBeilschmidt: Ok, but you don't need to wait for it at the end?

michaelmattig: if I don't await the connection, the test never completes, so unfortunately it seems we need this code. 

awaiting it at the end doesn't work.

services/tests/startup.rs
michaelmattig: there is no read_line method in AsyncReadExt, so I would prefer to leave the code like this and not implement it on my own.

services/tests/startup.rs
ChristianBeilschmidt: No, I meant awaiting the join handle of the spawn, not the inner await of the connection.

services/tests/startup.rs
ChristianBeilschmidt: And what about warnings that occur after the statement `Tokio runtime found`?
Now, you quit reading the 99 lines.

services/tests/startup.rs
michaelmattig: if I await it, the test never stops, because the connection never closes.

I could do

```
drop(client);
conn_handle.await.unwrap();
```

then it stops correctly. But I don't really see the point in doing this.

services/tests/startup.rs
michaelmattig: This test is only for checking warning during start up. "Tokio runtime found" is the last message during a successful startup. I haven't found a different way to detect this.

I quit after 99 lines because I don't want the test to run indefinitely and we never output even close to that number of messages during a successful stratup.

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #788, opened 2023-05-17
Tile cache lru replacement

michaelmattig:

---

<TEXT>


michaelmattig:
[Replacement von Tiles im Cache](https://trello.com/c/scVG55h1/1332-replacement-von-tiles-im-cache)

Settings-default.toml
```diff
@@ -178,3 +178,7 @@ volumes = { test_data = "test_data/"}
 
 [cache]
 enabled = false
+# storage limit for collecting query results before insertion into the cache
+landing_zone_byte_size = 100_000_000 # 100 MB
```
ChristianBeilschmidt, line 182: Maybe just define a Cache size and take 10% for the "landing zone"?

operators/src/engine/operator.rs
```diff
@@ -175,6 +175,16 @@ pub trait InitializedVectorOperator: Send + Sync {
...
 pub struct CanonicOperatorName(serde_json::Value);
 
+impl CanonicOperatorName {
+    pub fn new(value: serde_json::Value) -> Self {
+        CanonicOperatorName(value)
+    }
+
+    pub fn byte_size(&self) -> usize {
+        std::mem::size_of_val(self)
+    }
```
ChristianBeilschmidt, line 185: maybe we should check that this works?

operators/src/pro/cache/error.rs
```diff
@@ -0,0 +1,22 @@
...
+    QueryNotFoundInLandingZone,
+    InvalidRasterDataTypeForInsertion,
+}
+
+impl From<CacheError> for Error {
+    fn from(e: CacheError) -> Self {
+        Error::QueryingProcessorFailed {
+            source: Box::new(e),
+        }
+    }
```
ChristianBeilschmidt, line 21: Are these errors where the engine fails?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -2,52 +2,114 @@ use std::{collections::HashMap, pin::Pin, sync::Arc};
...
+    backend: RwLock<TileCacheBackend>,
+}
+
+impl TileCache {
+    pub fn new(cache_byte_size_total: usize, landing_zone_byte_size_total: usize) -> Result<Self> {
+        // TODO: ensure cache > landing zone
+        ensure!(
+            cache_byte_size_total >= landing_zone_byte_size_total,
+            CacheMustBeLargerOrEqualThanLandingZone
+        );
```
ChristianBeilschmidt, line 53: You did check that here, don't you?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -2,52 +2,114 @@ use std::{collections::HashMap, pin::Pin, sync::Arc};
...
+
+        Ok(Self {
+            backend: RwLock::new(TileCacheBackend {
+                operator_caches: Default::default(),
+                lru: LruCache::unbounded(), // we need no cap because we evict manually
+                cache_byte_size_total,
+                cache_byte_size_used: 0, // TODO: include overhead of HashMap?
```
ChristianBeilschmidt, line 60: We could leave it as is if this is contant.

operators/src/pro/cache/tile_cache.rs
```diff
@@ -2,52 +2,114 @@ use std::{collections::HashMap, pin::Pin, sync::Arc};
...
+                landing_zone_byte_size_total,
+                landing_zone_byte_size_used: 0, // TODO: include overhead of LruCache?
```
ChristianBeilschmidt, line 62: Same here

operators/src/pro/cache/tile_cache.rs
```diff
@@ -77,14 +143,38 @@ pub enum CachedTiles {
     F64(Arc<Vec<RasterTile2D<f64>>>),
 }
 
+impl CachedTiles {
+    fn byte_size(&self) -> usize {
+        // TODO: include size of the Vec itself and the Arc
```
ChristianBeilschmidt, line 148: it's just `std::mem::size_of::<CacheEntry>()`?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -307,38 +444,189 @@ impl TileCache {
...
+        while backend.cache_byte_size_used + entry_size > backend.cache_byte_size_total {
+            // this should always work, because otherwise it would mean the cache is not empty but the lru is.
+            // the landing zone is smaller than the cache size and the entry must fit into the landing zone.
+            if let Some((id, op)) = backend.lru.pop_lru() {
+                if let Some(evict_cache) = backend.operator_caches.get_mut(&op) {
+                    if let Some(entry) = evict_cache.entries.remove(&id) {
+                        backend.cache_byte_size_used -= entry.byte_size();
+                    }
+                }
+            }
```
ChristianBeilschmidt, line 533: This is super unreadable. Could you remove indentation layers, here?

ChristianBeilschmidt (changes requested):

operators/src/engine/operator.rs
michaelmattig: actually: id didn't. And moreover it appears that json::Value can be vastly bigger in memory than a json string. So I changed it to use the byte representation like you suggested

06ca2acb65400e096f0091afc467fccf0bcf5633

Settings-default.toml
michaelmattig: should this be fixed or configurable?

If it's fixed it may be a problem for very large cache size and only few insertions where much of the cache size will be wasted.

operators/src/pro/cache/tile_cache.rs
michaelmattig: you mean `CachedTiles`, right?

93cd66214d664374034b452067c33d5923694167

operators/src/pro/cache/tile_cache.rs
michaelmattig: I removed two indentations layers: 2317aaf5501492506fcd12b5ba899a4520a3a9a5

operators/src/pro/cache/error.rs
michaelmattig: you are right, these are internal errors that should not interfere with what the user gets to see. However, the functions should not panic, and have to return something if something goes wrong. So shall I change the return type of the TileCache functions to Options instead of Results then?

Settings-default.toml
ChristianBeilschmidt: Then, I would suggest setting a percentage for this, with 10% being the default.

operators/src/pro/cache/error.rs
ChristianBeilschmidt: I guess it is good for operations and debugging, that there are distinct errors.
Could we then just not convert it into Errors that we hand out? This way, we ensure that they stay internally.

michaelmattig: 2bfd416ea160135be0a8c7d68b6d550e40eaa7a3

ChristianBeilschmidt (approved):

PR #792, opened 2023-05-24
add bbox test to pip filter

jdroenner:

---

The PIP tester now generates the BBOX of each polygon at start. When testing, if a point i s inside a (multi)- polygon it first checks if it is inside (one of) the BBOX. In my test case this  results in a speed-up of 60x.


jdroenner:
[BBOX test for PIP tester](https://trello.com/c/lXRHezjE/1339-bbox-test-for-pip-tester)

operators/src/processing/point_in_polygon/tester.rs
```diff
@@ -90,6 +99,39 @@ impl<'a> PointInPolygonTester<'a> {
...
+    fn precalculate_polygon_bounds(
+        feature_offsets: &'a [i32],
+        polygon_offsets: &'a [i32],
+        ring_offsets: &'a [i32],
+        coordinates: &'a [Coordinate2D],
+    ) -> Vec<Vec<BoundingBox2D>> {
+        let mut multi_poly_bounds = Vec::with_capacity(feature_offsets.len() - 1);
+
+        for (feature_start_index, feature_end_index) in
+            two_tuple_windows(feature_offsets.iter().map(|&c| c as usize))
```
michaelmattig, line 111: why do you not use the iterators on the MultiPolygonCollection? Like `polygons.geometries().for_each()` or something like that. you should be able to get the coordinates this way too.

operators/src/processing/point_in_polygon/tester.rs
```diff
@@ -90,6 +99,39 @@ impl<'a> PointInPolygonTester<'a> {
         }
     }
 
+    fn precalculate_polygon_bounds(
```
michaelmattig, line 102: this should be easy to parallelize, right?

operators/src/processing/point_in_polygon/tester.rs
```diff
@@ -130,6 +172,14 @@ impl<'a> PointInPolygonTester<'a> {
         let polygon_offsets = self.polygon_offsets;
         let ring_offsets = self.ring_offsets;
 
+        let polygon_bounding_box = &self.polygon_bounds[feature_index];
+        if !polygon_bounding_box
+            .iter()
+            .any(|b| b.contains_coordinate(&coordinate))
```
michaelmattig, line 178: I think it would make sense to precompute the multi polygon bounds as well so you need only one check here

operators/src/processing/point_in_polygon/tester.rs
jdroenner: i thought that it would be a good idea to reuse the slices already created. However, i rebuild it using the iterators. [81ff742](https://github.com/geo-engine/geoengine/pull/792/commits/81ff742edf53e417c4c48d9c70684e9ba3889e51)

operators/src/processing/point_in_polygon/tester.rs
jdroenner: yes but lets do parallelization as an extra task. It will change a lot of other things, too.

operators/src/processing/point_in_polygon/tester.rs
jdroenner: i added that:  [81ff742](https://github.com/geo-engine/geoengine/pull/792/commits/81ff742edf53e417c4c48d9c70684e9ba3889e51)

jdroenner:
speedup in my test case is now ~ 100x

operators/src/processing/raster_vector_join/util.rs
```diff
@@ -137,6 +143,10 @@ impl CoveredPixels<MultiPoint> for MultiPointCoveredPixels {
     fn collection(self) -> FeatureCollection<MultiPoint> {
         self.collection
     }
+
+    fn covered_spatial_bounds(&self) -> Option<BoundingBox2D> {
+        self.collection.bbox()
```
michaelmattig, line 148: I am not sure if this is correct. This is the bbox of the whole collection, but we are always looking at single tiles, which may not cover the whole collection.

operators/src/processing/raster_vector_join/aggregated.rs
```diff
@@ -70,9 +70,14 @@ where
 
         let collection = covered_pixels.collection_ref();
 
+        let covered_bounds = covered_pixels
+            .covered_spatial_bounds()
+            .and_then(|b| b.intersection(&query.spatial_bounds))
+            .unwrap_or(query.spatial_bounds); // TODO: if the intersection is empty, we should skip the query
```
michaelmattig, line 76: there are two cases where you unwrap a None:
1. the collection has no features
2. the query does not intersect the covered pixels

in the first case we could skip, but we don't need the intersection for this.
the second case would be a bug or not?

jdroenner: i removed that

datatypes/src/primitives/bounding_box.rs
```diff
@@ -343,8 +343,17 @@ impl BoundingBox2D {
...
+            .lower_left_coordinate
+            .min_elements(other.lower_left_coordinate);
+        self.upper_right_coordinate = self
+            .upper_right_coordinate
+            .max_elements(other.upper_right_coordinate);
+    }
+
     #[must_use]
-    pub fn extend(&mut self, other: &Self) -> Self {
+    pub fn union(&self, other: &Self) -> Self {
```
michaelmattig, line 356: is union really a better word here? If two bboxes are completely disjoint, is it still a union if we create a new one that contains both?

michaelmattig (approved):

PR #796, opened 2023-06-02
dataset names everywhere in dataset api

ChristianBeilschmidt:

---

<TEXT>

### DIFF
--- a/services/src/handlers/datasets.rs
+++ b/services/src/handlers/datasets.rs
@@ -143,7 +143,7 @@ pub async fn list_datasets_handler<C: ApplicationContext>(
     Ok(web::Json(list))
 }
 
-/// Retrieves details about a dataset using the internal id.
+/// Retrieves details about a dataset using the internal name.
 #[utoipa::path(
     tag = "Datasets",
     get,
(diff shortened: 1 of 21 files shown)


ChristianBeilschmidt:
[Stable Names for datasets](https://trello.com/c/zJVzRWzc/852-stable-names-for-datasets)

services/src/handlers/datasets.rs
```diff
@@ -169,29 +169,34 @@ pub async fn list_datasets_handler<C: ApplicationContext>(
...
-        ("dataset" = DatasetId, description = "Dataset id")
+        ("dataset" = DatasetName, description = "Dataset Name")
     ),
     security(
         ("session_token" = [])
     )
 )]
 pub async fn get_dataset_handler<C: ApplicationContext>(
-    dataset: web::Path<DatasetId>,
+    dataset: web::Path<DatasetName>,
```
michaelmattig, line 185: I can't comment the right line, but the doc comment still says "internal id" which is no longer true

ChristianBeilschmidt: [927b174](https://github.com/geo-engine/geoengine/pull/796/commits/927b174498f2a3bb9df7199d06f082ada6cdb218)

michaelmattig (approved):

PR #802, opened 2023-06-15
cache expiration

michaelmattig:

---

<TEXT>


michaelmattig:
[TTL für Tile-Cache-EInträge](https://trello.com/c/R5Ux2Bgn/1333-ttl-f%C3%BCr-tile-cache-eintr%C3%A4ge)

operators/src/pro/cache/tile_cache.rs
```diff
@@ -100,9 +100,42 @@ pub struct OperatorTileCache {
...
+            if matches && r.is_expired() {
+                expired_entries.push(**id);
+
+                return false;
+            }
+
+            matches
+        });
+
+        let match_found = match_found.map(|(id, _)| *id);
```
ChristianBeilschmidt, line 129: just do the map there?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -165,6 +202,21 @@ impl CachedTiles {
...
+            CachedTiles::U8(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::U16(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::U32(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::U64(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::I8(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::I16(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::I32(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::I64(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::F32(v) => v.iter().any(|t| t.cache_hint.is_expired()),
+            CachedTiles::F64(v) => v.iter().any(|t| t.cache_hint.is_expired()),
```
ChristianBeilschmidt, line 217: is this expensive?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -389,15 +441,37 @@ impl TileCache {
...
-            let typed_stream = entry.tile_stream(query);
+        // remove expired entries from lru index
+        for entry in expired_entries {
+            backend.lru.pop_entry(&entry);
+        }
 
-            (*id, typed_stream)
-        };
+        // unpack match if there is one
+        let (entry_id, typed_stream) = result?;
```
ChristianBeilschmidt, line 474: When you do `?`, you could fail to promote. Is that ok?

operators/src/processing/interpolation/mod.rs
```diff
@@ -224,7 +225,7 @@ where
             ctx,
             sub_query,
         )
-        .filter_and_fill())
+        .filter_and_fill(CacheExpiration::NoCache)) // Filler tiles may be produced by gaps in the input data and we cannot be sure if the gaps are permanent or not. Thus, we do not cache them.
```
ChristianBeilschmidt, line 228: Don't filler tiles have a useful cache time?

operators/src/processing/map_query.rs
```diff
@@ -44,7 +45,13 @@ where
             log::debug!("Query was rewritten to empty query. Returning empty / filled stream.");
             let s = futures::stream::empty();
 
-            Ok(SparseTilesFillAdapter::new_like_subquery(s, query, self.additional_data).boxed())
+            Ok(SparseTilesFillAdapter::new_like_subquery(
+                s,
+                query,
+                self.additional_data,
+                CacheExpiration::NoCache, // TODO: can we do better?
```
ChristianBeilschmidt, line 52: Use the input?

operators/src/processing/neighborhood_aggregate/mod.rs
```diff
@@ -254,14 +255,16 @@ where
...
-            sub_query,
+        Ok(
+            RasterSubQueryAdapter::<'a, P, _, _>::new(
+                &self.source,
+                query,
+                self.tiling_specification,
+                ctx,
+                sub_query,
+            )
+            .filter_and_fill(CacheExpiration::NoCache), // Filler tiles may be produced by gaps in the input data and we cannot be sure if the gaps are permanent or not. Thus, we do not cache them.
```
ChristianBeilschmidt, line 266: cf. above

operators/src/processing/reprojection.rs
```diff
@@ -555,7 +555,7 @@ where
                 ctx,
                 sub_query_spec,
             )
-            .filter_and_fill())
+            .filter_and_fill(CacheExpiration::NoCache)) // TODO: if we were sure that gaps only occur where the projection is not defined in the input, we could cache this forever
```
ChristianBeilschmidt, line 558: hm, we should think about something here

services/src/datasets/external/aruna/mod.rs
```diff
@@ -486,6 +488,7 @@ impl ArunaDataProvider {
                 allow_alphaband_as_mask: true,
                 retry: None,
             }),
+            cache_ttl: CacheTtl::Seconds(15 * 60), // the download-link validity
```
ChristianBeilschmidt, line 491: could we do a constant here?

services/src/datasets/external/gbif.rs
```diff
@@ -2069,6 +2073,7 @@ mod tests {
                     .iter()
                     .cloned()
                     .collect(),
+                    CacheHint::default(), // TODO: make configurable in data provider(?)
```
ChristianBeilschmidt, line 2076: sounds reasonable

services/src/datasets/external/gfbio_abcd.rs
```diff
@@ -855,6 +859,7 @@ mod tests {
                 .iter()
                 .cloned()
                 .collect(),
+                CacheHint::default(), // TODO: make configurable in data provider(?)
```
ChristianBeilschmidt, line 862: +1

services/src/datasets/external/nature40.rs
```diff
@@ -464,6 +465,7 @@ impl MetaDataProvider<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectan
                 Some(self.auth().to_vec()),
             )?,
             result_descriptor: raster_descriptor_from_dataset(&dataset, band_index as isize)?,
+            cache_ttl: CacheTtl::Seconds(15 * 60),
```
ChristianBeilschmidt, line 468: constant

services/src/datasets/external/nature40.rs
```diff
@@ -959,7 +962,8 @@ mod tests {
                         gdal_config_options: None,
                         allow_alphaband_as_mask: true,
                         retry: None,
-                    })
+                    }),
+                    cache_ttl: CacheTtl::Seconds(15 * 60),
```
ChristianBeilschmidt, line 966: cf. above

services/src/datasets/external/netcdfcf/mod.rs
```diff
@@ -514,6 +515,7 @@ impl NetCdfCfDataProvider {
                 end, // TODO: Use this or time dimension size (number of steps)?
                 step,
                 band_offset: dataset_id.entity * dimensions_time,
+                cache_ttl: CacheTtl::default(),
```
ChristianBeilschmidt, line 518: could also be configurable

ChristianBeilschmidt (changes requested):

datatypes/src/primitives/ttl.rs
```diff
@@ -0,0 +1,258 @@
+use serde::de::Error;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use super::DateTime;
+
+/// Config parameter to indicate how long a value may be cached
+#[derive(Default, Debug, Clone, Copy, PartialEq)]
+pub enum CacheTtl {
+    #[default] // TODO: make the default globally configurable in Settings.toml?
```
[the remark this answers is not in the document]
michaelmattig: I think NoCache is a reasonable default most of the time. But maybe very high-traffic instances want to e.g. cache everything for a minute or something like that regardless of the expiration date of the incorporated data.

datatypes/src/primitives/ttl.rs
```diff
@@ -0,0 +1,258 @@
...
+    NoCache,
+    Seconds(u32),
+    Unlimited,
```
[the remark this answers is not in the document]
michaelmattig: We could do that as a simplification, yes.

In reality, however, I think, there are three cases: 1. Something should not be cached, 2 something can be cached for some time, 3. something is guaranteed to never change and can be cached forever. That's why I built it this way. 

There are cases for point 3, e.g. in the reprojection operator. If the area of a tile in the source projection is not defined, we know this will always be no data.

If we, however, do not really exploit this information, then we may use a simpler model of reality. 

Is there a benefit in using the simpler model?

datatypes/src/raster/operations/convert_data_type.rs
```diff
@@ -55,6 +55,7 @@ where
             global_geo_transform: self.global_geo_transform,
             properties: self.properties,
             tile_position: self.tile_position,
+            cache_hint: self.cache_hint, // TODO: update created date?
```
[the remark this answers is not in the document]
michaelmattig: But we do not really do updates, but rather create new tiles. We could maybe just do CacheHint::from(self.cache_hint.expiration())` here, because that uses the current system date time.

datatypes/src/primitives/ttl.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+}
+
+/// Type for storing the expiration data to avoid recomputing it from creation date time and ttl
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub enum CacheExpiration {
+    NoCache,   // TODO: rename: Always?
+    Unlimited, // TODO: rename: Never?
+    Expires(DateTime),
+}
```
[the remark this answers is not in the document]
michaelmattig: you mean CacheExpirtion would just be a `type CacheExpiration = Option<DateTime>`?

operators/src/adapters/raster_subquery/raster_subquery_adapter.rs
```diff
@@ -543,8 +553,12 @@ pub fn identity_accu<T: Pixel>(
 ) -> impl Future<Output = Result<RasterTileAccu2D<T>>> {
     crate::util::spawn_blocking(move || {
         let output_raster = EmptyGrid2D::new(tile_info.tile_size_in_pixels).into();
-        let output_tile =
-            RasterTile2D::new_with_tile_info(query_rect.time_interval, tile_info, output_raster);
+        let output_tile = RasterTile2D::new_with_tile_info(
+            query_rect.time_interval,
+            tile_info,
+            output_raster,
+            CacheHint::unlimited(),
```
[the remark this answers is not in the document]
michaelmattig: `CacheHint::unlimited()` is the initial state in every accumulator. It is then merged with the cache hint of all the tiles that go into the accumulator. If all are also unlimited cacheable, it stays unlimited. If any is non cacheable or do expire, the final cache hint will reflect this accordingly.

The only requirement here is that at least one tile is fed into the accumulator. This should always be the case for non-empty query rectangles, because at the very least we get empty tiles.

operators/src/pro/cache/tile_cache.rs
```diff
@@ -100,9 +100,42 @@ pub struct OperatorTileCache {
...
+            }
+
+            matches
+        });
+
+        let match_found = match_found.map(|(id, _)| *id);
+
+        for id in &expired_entries {
+            self.entries.remove(id);
+        }
```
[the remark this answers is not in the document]
michaelmattig: it's not possible to mutable borrow the `entries` while iterating over them.

operators/src/pro/cache/tile_cache.rs
```diff
@@ -100,9 +100,42 @@ pub struct OperatorTileCache {
...
-        self.entries.iter().find(|(_id, r)| r.matches(query))
+    /// Find a match for the query in the cache.
+    ///
+    /// This method will also remove expired entries it encounters.
+    ///
+    /// It returns the expired entries, s.t. the caller can remove them from the lru index structure.
+    fn find_match(&mut self, query: &RasterQueryRectangle) -> FindMatchResult {
+        let mut expired_entries = vec![];
+
+        let match_found = self.entries.iter().find(|(id, r)| {
```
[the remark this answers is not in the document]
michaelmattig: currently yes. In the future we could use a spatio-temporal index structure to find matches more efficiently. But right now we don't have that.

operators/src/pro/cache/tile_cache.rs
michaelmattig: where?

operators/src/pro/cache/tile_cache.rs
michaelmattig: it only fails here if no match was found, so there is nothing to promote

operators/src/processing/interpolation/mod.rs
michaelmattig: c.f. my comment in the trello card.

services/src/datasets/external/gbif.rs
michaelmattig: ok, but let's first agree on how we model ttl (seconds or enum)

datatypes/src/primitives/ttl.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Ok, so either write this also inside the TODO or remove it?

datatypes/src/primitives/ttl.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Yeah, so maybe we could go with the HTTP model where unlimited=1 year.

Simpler is simpler and in the end, what is unlimited anyway.

datatypes/src/primitives/ttl.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Just a datetime, no option. For NoCache, just set it to now :-)

datatypes/src/raster/operations/convert_data_type.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Sure, we can do that.

datatypes/src/raster/raster_tile.rs
```diff
@@ -44,6 +45,24 @@ pub struct BaseTile<G> {
...
+    G: PartialEq,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.time == other.time
+            && self.tile_position == other.tile_position
+            && self.global_geo_transform == other.global_geo_transform
+            && self.grid_array == other.grid_array
+            && self.properties == other.properties
+        // deliberately omitted: self.cache_hint == other.cache_hint
+    }
```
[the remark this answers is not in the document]
ChristianBeilschmidt: 1. We should implement stuff for production, not test.
So we rather impl a new comparison method that we can use there.

2. Now we should think how we will use it in production? Can the IDE find uses of that?
If so, we should make it so that it does the right thing there.

operators/src/pro/cache/tile_cache.rs
ChristianBeilschmidt: 126

operators/src/processing/map_query.rs
ChristianBeilschmidt: If we map query, we will get tiles back from which we can use the expiration?

operators/src/processing/interpolation/mod.rs
ChristianBeilschmidt: I think you are right that operators should not output gaps.
We could add a parameter to the filler that would input the expiration time. Then, at least each operator can decide on that.

jdroenner: Please think of the implications. This will make all results where a tile is filled non-cachable. I bet that almost all queries will contain some filled tiles. There are operators (GdalSource, all based on SubQueryAdapter) that just don't output empty tiles and  rely on the filler to produce them. We probably need to pass something like a strategy to the filler to decide what the TTL of a tile created by it should be.

datatypes/src/raster/raster_tile.rs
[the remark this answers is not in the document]
michaelmattig: > Now we should think how we will use it in production? Can the IDE find uses of that?

Unfortunately it can't. But I guess even if it wasn't used, you are right, it should return something useful in prod rather than for tests.

So should, in practice, two tiles with the same position, time and data but different creation and expiration times considered the same? I guess not, if they were the same, they could be used interchangeable without any implications and with the cache enabled this is not the case.

So I guess I should include the cache hint in the comparison and adjust the tests

datatypes/src/primitives/ttl.rs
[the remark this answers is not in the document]
michaelmattig: I removed it f610f2dc12d34978917b3d46dc4880985e5f12e8

datatypes/src/primitives/ttl.rs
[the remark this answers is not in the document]
michaelmattig: I simplified it using only seconds: f610f2dc12d34978917b3d46dc4880985e5f12e8

datatypes/src/primitives/ttl.rs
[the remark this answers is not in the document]
michaelmattig: f610f2dc12d34978917b3d46dc4880985e5f12e8

operators/src/pro/cache/tile_cache.rs
michaelmattig: The closure in 126 returns a bool and not a CacheEntryId I could put it in 127 but then rustfmt breaks the stuff down even further and we end up in line 131 instead of 129, so haven't really saved anything.

operators/src/processing/map_query.rs
michaelmattig: yes, but in this branch of the if-statement, the query couldn't be rewritten and we start from an empty stream. If I am not mistaken.

operators/src/processing/interpolation/mod.rs
michaelmattig: I implemented a filler strategy to derive the cache hint from the surrounding tiles 9b5283436f1ac6f6344a5396a0d0ddda4be63040

operators/src/processing/neighborhood_aggregate/mod.rs
michaelmattig: 9b5283436f1ac6f6344a5396a0d0ddda4be63040

operators/src/processing/reprojection.rs
michaelmattig: 9b5283436f1ac6f6344a5396a0d0ddda4be63040

services/src/datasets/external/gbif.rs
michaelmattig: 5c04d280f009ec129d8b8fb5a92e3fc1ca61c65b

services/src/datasets/external/aruna/mod.rs
michaelmattig: I changed this to using the specified ttl in the provider definition: 5c04d280f009ec129d8b8fb5a92e3fc1ca61c65b

services/src/datasets/external/nature40.rs
michaelmattig: I changed this to using the ttl from the provider definition 5c04d280f009ec129d8b8fb5a92e3fc1ca61c65b

services/src/datasets/external/nature40.rs
michaelmattig: 5c04d280f009ec129d8b8fb5a92e3fc1ca61c65b

services/src/datasets/external/netcdfcf/mod.rs
michaelmattig: 5c04d280f009ec129d8b8fb5a92e3fc1ca61c65b

operators/src/processing/map_query.rs
ChristianBeilschmidt: Could you add this as a comment? This is still unclear to me.

services/src/handlers/wms.rs
```diff
@@ -988,4 +990,29 @@ mod tests {
...
+        assert_eq!(
+            response.status(),
+            200,
+            "{:?}",
+            actix_web::test::read_body(response).await
+        );
+
+        assert_eq!(
+            response.headers().get(header::CACHE_CONTROL).unwrap(),
+            "no-cache"
```
ChristianBeilschmidt, line 1015: Could you add a test for not "no-cache"?

ChristianBeilschmidt (changes requested):

operators/src/processing/map_query.rs
michaelmattig: 4a6cb5331c6f8b07933327ef42b8ae8c924bb066

services/src/handlers/wms.rs
michaelmattig: 4a6cb5331c6f8b07933327ef42b8ae8c924bb066

ChristianBeilschmidt (approved):

PR #805, opened 2023-06-19
calculate grid bytes for cache entries

ChristianBeilschmidt:

---

<TEXT>

### DIFF
--- a/datatypes/src/raster/grid.rs
+++ b/datatypes/src/raster/grid.rs
@@ -509,14 +509,14 @@ impl<D, P> ByteSize for Grid<D, P>
 where
     P: Pixel,
 {
-    fn byte_size(&self) -> usize {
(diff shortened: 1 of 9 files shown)


datatypes/src/raster/grid.rs
```diff
@@ -505,16 +505,31 @@ where
     }
 }
 
+impl<D, P> ByteSize for Grid<D, P>
+where
+    P: Pixel,
```
jdroenner, line 510: why do we need P: Pixel? Wouldn't P: Sized also work?

ChristianBeilschmidt: [15cc8a5](https://github.com/geo-engine/geoengine/pull/805/commits/15cc8a55c69d21b48e92135c7d0d40a3e7d4c5c9)

jdroenner (approved):

PR #808, opened 2023-06-22
enhance ml exmple performance

jdroenner:
- I added a ChunkMerger adapter after the OGR Source since the source produces chunks that are smaller that the chunks produced by the ChunkMerger.
-  Skip raster queries in RasterVectorJoin if  there is no intersection between chunk and query
- increase cache entry temporal and spatial bounds to the tiles covered.
- enable cache between all operators (we should discuss that)

---

<TEXT>

### DIFF
--- a/operators/src/pro/cache/tile_cache.rs
+++ b/operators/src/pro/cache/tile_cache.rs
@@ -477,6 +477,14 @@ impl TileCache {
             .get_mut(&query_id)
             .ok_or(super::error::CacheError::QueryNotFoundInLandingZone)?;
 
+        // TODO: rollback if one fails?
+        entry.query.spatial_bounds = entry.query.spatial_bounds.extend(&tile.spatial_partition()); // since the source should only produce tiles that intersect with the query, we can extend the query bounds
+        entry.query.time_interval = entry
+            .query
+            .time_interval
+            .union(&tile.time)
+            .expect("time of tile must overlap with query");
+
         T::insert_tile(&mut entry.tiles, tile)?;
 
         backend.landing_zone_byte_size_used += entry.byte_size();
--- a/operators/src/source/ogr_source/mod.rs
+++ b/operators/src/source/ogr_source/mod.rs
@@ -36,6 +36,7 @@ use geoengine_datatypes::primitives::{
 };
 use geoengine_datatypes::util::arrow::ArrowTyped;
 
+use crate::adapters::FeatureCollectionStreamExt;
 use crate::engine::{
     CanonicOperatorName, OperatorData, OperatorName, QueryProcessor, WorkflowOperatorPath,
 };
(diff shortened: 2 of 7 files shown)


jdroenner:
[Analyse des Workflows aus dem ML-Beispiel & erstellen einer Strategie die Performance zu verbessern.](https://trello.com/c/hOg2MIlH/1293-analyse-des-workflows-aus-dem-ml-beispiel-erstellen-einer-strategie-die-performance-zu-verbessern)

operators/src/pro/cache/tile_cache.rs
```diff
@@ -477,6 +477,14 @@ impl TileCache {
             .get_mut(&query_id)
             .ok_or(super::error::CacheError::QueryNotFoundInLandingZone)?;
 
+        // TODO: rollback if one fails?
```
michaelmattig, line 480: do you mean one `extend`? this should never happen, right? Otherwise there is a bug in the engine somewhere. So you can remove the TODO(?)

operators/src/source/ogr_source/mod.rs
```diff
@@ -4069,9 +4072,9 @@ mod tests {
 
         let result: Vec<MultiPointCollection> = query.try_collect().await.unwrap();
 
-        assert_eq!(result.len(), 1);
-
-        assert!(result[0].is_empty());
+        // FIXME: this should be an empty collection. The ChunkMerger does not forward a single empty collection
```
michaelmattig, line 4075: Should we create a follow up task?

operators/src/source/ogr_source/mod.rs
```diff
@@ -525,6 +526,7 @@ where
             self.attribute_filters.clone(),
         )
         .await?
+        .merge_chunks(ctx.chunk_byte_size().into()) // rechunk the data if necessary TODO: remove when source produces the right chunk sizes
```
michaelmattig, line 529: Should we create a follow up task?

jdroenner: yes it is the one with the builder byte size stuff.

operators/src/pro/cache/tile_cache.rs
jdroenner: [ec74825](https://github.com/geo-engine/geoengine/pull/808/commits/ec74825ab6fe080db6782663ec4ae2274c294c62)

operators/src/source/ogr_source/mod.rs
jdroenner: [a31c549](https://github.com/geo-engine/geoengine/pull/808/commits/a31c549aee11433c22f621cb2052ac8426ad8a47)

michaelmattig (approved):

PR #818, opened 2023-07-07
allow name for datasets from workflow

jdroenner:

---

create dataset from worklfow now has the option to set a name for the new dataset

### DIFF
--- a/services/src/handlers/layers.rs
+++ b/services/src/handlers/layers.rs
@@ -628,7 +628,8 @@ async fn layer_to_dataset<C: ApplicationContext>(
     };
 
     let from_workflow = RasterDatasetFromWorkflow {
-        name: layer.name,
+        name: None,
+        display_name: layer.name,
         description: Some(layer.description),
         query: qr,
         as_cog: true,
--- a/services/src/datasets/create_from_workflow.rs
+++ b/services/src/datasets/create_from_workflow.rs
@@ -31,9 +31,11 @@ use utoipa::ToSchema;
 
 /// parameter for the dataset from workflow handler (body)
 #[derive(Clone, Debug, Deserialize, Serialize, ToSchema)]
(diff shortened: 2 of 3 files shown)


jdroenner:
[name for dataset from workflow](https://trello.com/c/39aZLCD1/1378-name-for-dataset-from-workflow)

services/src/handlers/layers.rs
```diff
@@ -628,7 +628,8 @@ async fn layer_to_dataset<C: ApplicationContext>(
     };
 
     let from_workflow = RasterDatasetFromWorkflow {
-        name: layer.name,
+        name: None,
```
michaelmattig, line 631: do you really want a random Uuid as name here? Maybe the name should be a parameter of the handler, or: derived from the layer name by replacing all disallowed characters with _ or something like that.

jdroenner: we can talk about that. This PR does not change the behavior of that functionality. 
If we change it: It will fail once you do that twice. So we need a mechanism to delete the datasets?

michaelmattig: but it does change the behavior, because previously the name was the layer name and now it is random.

> If we change it: It will fail once you do that twice. So we need a mechanism to delete the datasets?

that's why I think the name should be a mandatory user parameter. The layer name most likely wouldn't be a valid name either because of the limited allowed characters. And random names defeat the whole purpose of the names.

jdroenner: No it does not change the behavior.  Previously the RasterDatasetFromWorkflow had no field for the dataset name. The old "name" really is the display_name. Just look how the DatasetDefinition is created in raster_dataset_from_workflow. There the name was always set to none and the display_name was set from name.

michaelmattig (approved):

PR #819, opened 2023-07-07
Tile-cache-cleanup

jdroenner:

---

<TEXT>

### DIFF
--- a/datatypes/src/primitives/spatial_partition.rs
+++ b/datatypes/src/primitives/spatial_partition.rs
@@ -167,7 +167,9 @@ impl SpatialPartition2D {
         }
     }
 
-    /// Return true if the partition contains the `other`
+    /// Return true if the partition contains the `other`.
+    /// A partition contains another partition if it contains all of its points
+    /// OR the other partition is equal to it.
     pub fn contains(&self, other: &Self) -> bool {
         self.contains_x(other) && self.contains_y(other)
     }
@@ -177,7 +179,7 @@ impl SpatialPartition2D {
             other.upper_left_coordinate.x,
             self.upper_left_coordinate.x,
             self.lower_right_coordinate.x,
-        ) && crate::util::ranges::value_in_range(
+        ) && crate::util::ranges::value_in_range_inv(
             other.lower_right_coordinate.x,
             self.upper_left_coordinate.x,
             self.lower_right_coordinate.x,
@@ -185,7 +187,7 @@ impl SpatialPartition2D {
     }
 
     fn contains_y(&self, other: &Self) -> bool {
-        crate::util::ranges::value_in_range_inv(
+        crate::util::ranges::value_in_range(
             other.lower_right_coordinate.y,
             self.lower_right_coordinate.y,
             self.upper_left_coordinate.y,
@@ -371,6 +373,12 @@ mod tests {
 
         assert!(p1.contains_coordinate(&c1));
         assert!(!p1.contains_coordinate(&c2));
+    }
+
+    #[test]
+    fn it_contains_self() {
+        let p1 = SpatialPartition2D::new_unchecked((0., 1.).into(), (1., 0.).into());
+        assert!(p1.contains(&p1));
     }
 
     #[test]
--- a/datatypes/src/primitives/spatial_partition.rs
+++ b/datatypes/src/primitives/spatial_partition.rs
@@ -226,7 +226,7 @@ impl SpatialPartition2D {
     }
 
     #[must_use]
-    pub fn extend(&mut self, other: &Self) -> Self {
+    pub fn extended(self, other: &Self) -> Self {
         Self {
             upper_left_coordinate: Coordinate2D::new(
                 self.upper_left().x.min(other.upper_left().x),
@@ -237,6 +237,13 @@ impl SpatialPartition2D {
                 self.lower_right().y.min(other.lower_right().y),
             ),
         }
+    }
+
+    pub fn extend(&mut self, other: &Self) {
+        self.upper_left_coordinate.x = self.upper_left().x.min(other.upper_left().x);
+        self.upper_left_coordinate.y = self.upper_left().y.max(other.upper_left().y);
+        self.lower_right_coordinate.x = self.lower_right().x.max(other.lower_right().x);
+        self.lower_right_coordinate.y = self.lower_right().y.min(other.lower_right().y);
     }
 }
 
@@ -311,7 +318,7 @@ pub fn partitions_extent<I: Iterator<Item = Option<SpatialPartition2D>>>(
 
     for bbox in bboxes {
         if let Some(bbox) = bbox {
-            extent = extent.extend(&bbox);
+            extent.extend(&bbox);
         } else {
             return None;
         }
@@ -515,6 +522,27 @@ mod tests {
                 .into_iter()
             ),
             None
+        );
+    }
+
+    #[test]
+    fn extend() {
+        let mut p1 = SpatialPartition2D::new_unchecked((0., 1.).into(), (1., 0.).into());
+        let p2 = SpatialPartition2D::new_unchecked((1., 2.).into(), (2., 0.).into());
+        p1.extend(&p2);
+        assert_eq!(
+            p1,
+            SpatialPartition2D::new_unchecked((0., 2.).into(), (2., 0.).into())
+        );
+    }
+
+    #[test]
+    fn extended() {
+        let p1 = SpatialPartition2D::new_unchecked((0., 1.).into(), (1., 0.).into());
+        let p2 = SpatialPartition2D::new_unchecked((1., 2.).into(), (2., 0.).into());
+        assert_eq!(
+            p1.extended(&p2),
+            SpatialPartition2D::new_unchecked((0., 2.).into(), (2., 0.).into())
         );
     }
 }
--- a/datatypes/src/primitives/spatial_partition.rs
+++ b/datatypes/src/primitives/spatial_partition.rs
@@ -227,16 +227,9 @@ impl SpatialPartition2D {
 
     #[must_use]
     pub fn extended(self, other: &Self) -> Self {
-        Self {
-            upper_left_coordinate: Coordinate2D::new(
-                self.upper_left().x.min(other.upper_left().x),
-                self.upper_left().y.max(other.upper_left().y),
-            ),
-            lower_right_coordinate: Coordinate2D::new(
-                self.lower_right().x.max(other.lower_right().x),
-                self.lower_right().y.min(other.lower_right().y),
-            ),
-        }
+        let mut extended = self;
+        extended.extend(other);
+        extended
     }
 
     pub fn extend(&mut self, other: &Self) {
--- a/operators/src/pro/cache/tile_cache.rs
+++ b/operators/src/pro/cache/tile_cache.rs
@@ -474,6 +474,12 @@ impl TileCache {
             });
         }
 
+        if landing_zone_ratio >= 1.0 {
+            return Err(crate::error::Error::QueryingProcessorFailed {
+                source: Box::new(CacheError::LandingZoneRatioMustBeSmallerThanOne),
+            });
+        }
+
         let cache_size_bytes =
             (cache_size_in_mb as f64 * (1.0 - landing_zone_ratio) * 1024.0 * 1024.0) as usize;
 
--- a/operators/src/pro/cache/tile_cache.rs
+++ b/operators/src/pro/cache/tile_cache.rs
@@ -208,7 +208,7 @@ impl TileCacheBackend {
         key: &CanonicOperatorName,
         query: &RasterQueryRectangle,
     ) -> Result<QueryId, CacheError> {
-        // if there is no cache for this operator graph, we create one. However, it is not clear yet, if we will actually cache anything.
+        // if there is no cache for this operator graph, we create one. However, it is not clear yet, if we will actually cache anything. The underling query might be aborted.
         if !self.operator_caches.contains_key(key) {
             self.operator_caches.insert(key.clone(), Default::default());
             // self.cache_size.try_add_element_bytes(&key)?; // TODO: can we delete this at a later point?
@@ -233,10 +233,10 @@ impl TileCacheBackend {
             .try_add_element_bytes(&landing_zone_entry)?;
         let old_entry = landing_zone.insert(query_id, landing_zone_entry);
 
-        if let Some(old_entry) = old_entry {
-            debug!("Replacing old entry in landing zone!");
-            self.landing_zone_size.remove_element_bytes(&old_entry); // Since the old entry already was in the landing zone, we need to remove it from the size.
-        }
+        debug_assert!(
+            old_entry.is_none(),
+            "There must not be an entry for the same query id in the landing zone! This is a bug."
+        );
 
         Ok(query_id)
     }
--- a/operators/src/pro/cache/tile_cache.rs
+++ b/operators/src/pro/cache/tile_cache.rs
@@ -411,10 +411,10 @@ impl TileCacheBackend {
                 let cache = &mut self
                     .operator_caches
                     .get_mut(&pop_entry_key)
-                    .expect("There must be cache elements entry if there was an LRU entry")
-                    .entries;
+                    .expect("There must be cache elements entry if there was an LRU entry");
+                let cache_entries = &mut cache.entries;
                 {
-                    let old_cache_entry = cache
+                    let old_cache_entry = cache_entries
                         .remove(&pop_entry_id)
                         .expect("LRU entry must be in cache");
                     self.cache_size.remove_element_bytes(&pop_entry_id);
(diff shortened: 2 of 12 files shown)


jdroenner:
[Cache Size Accounting Refactoring](https://trello.com/c/08J28uBS/1379-cache-size-accounting-refactoring)

datatypes/src/primitives/spatial_partition.rs
```diff
@@ -238,6 +238,13 @@ impl SpatialPartition2D {
             ),
         }
     }
+
+    pub fn extend(&mut self, other: &Self) {
+        self.upper_left_coordinate.x = self.upper_left().x.min(other.upper_left().x);
```
michaelmattig, line 243: can you do `self = self.extened(&other)` instead of repeating the logic?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
-    landing_zone_byte_size_used: usize,
+    cache_size: TileCacheBackendSize,
+    landing_zone_size: TileCacheBackendSize,
 
     // we only use the LruCache for determining the least recently used elements and evict as many entries as needed to fit the new one
     lru: LruCache<CacheEntryId, CanonicOperatorName>,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct TileCacheBackendSize {
```
michaelmattig, line 41: is this in any way specific to the tile cache? If not, I think you could name it differently and put it somewhere else, e.g. to the ByteSize trait or to a util file.

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+            self.can_fit_bytes(bytes),
+            "adding too many bytes {} for free capacity {}",
+            bytes,
+            self.byte_size_free()
+        );
+        self.byte_size_used += bytes;
+    }
+
+    #[inline]
+    #[allow(dead_code)] // used for tests
```
michaelmattig, line 105: would cfg(test) make sense here?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+                "overflowing cache size by {} bytes, total size: {}, added bytes: {}",
+                self.byte_size_used - self.byte_size_total,
+                self.byte_size_total,
+                bytes
+            );
+        }
+    }
+}
+
+type CacheSizeLruRef<'a> = (
```
michaelmattig, line 172: why do you need these helper types?
Maybe make a struct out of it instead so it's more apparent what each element is?

Same for the other helper types

jdroenner: clippy wanted that type

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+            }
+        }
+    }
+
+    fn remove_querys_from_cache_and_lru(
+        &mut self,
+        key: &CanonicOperatorName,
+        cache_entry_ids: &[CacheEntryId],
+    ) {
+        if let Some((cache, cache_size, cache_lru)) = self.cache_and_size_and_lru_mut(key) {
```
michaelmattig, line 440: I think you can avoid looking up the cache again, and instead take it into this function as a parameter

operators/src/pro/cache/tile_cache.rs
jdroenner: i do in 379 and 440

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+            for cache_entry_id in cache_entry_ids {
+                let _old_lru_entry = cache_lru.pop_entry(cache_entry_id);
+                if let Some(old_cache_entry) = cache.remove(cache_entry_id) {
```
michaelmattig, line 443: it's nice that you wrapped the functionality for calculating the cache size. Would it make sense to wrap the whole cache Hashmap and the lru type with a custom type that autoamtically adjusts the byte size on remove/pop so that we can't forget it?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -52,20 +462,18 @@ impl TileCache {
             });
         }
 
+        let cache_size_bytes =
```
michaelmattig, line 465: regarding the check above this line: maybe we should also check that the ratio is < 1

operators/src/pro/cache/tile_cache.rs
```diff
@@ -148,13 +517,16 @@ pub struct CacheEntry {
...
+impl ByteSize for CacheEntryId {}
+
 impl CacheEntry {
     /// Return true if the query can be answered in full by this cache entry
     /// For this, the bbox and time has to be fully contained, and the spatial resolution has to match
     pub fn matches(&self, query: &RasterQueryRectangle) -> bool {
         (self.query.spatial_bounds == query.spatial_bounds
             || self.query.spatial_bounds.contains(&query.spatial_bounds))
-            && self.query.time_interval.contains(&query.time_interval)
+            && (self.query.time_interval == query.time_interval
```
michaelmattig, line 528: the contains method already checks for equality or am I wrong?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+
+    pub fn query_cache<T>(
+        &mut self,
+        key: &CanonicOperatorName,
+        query: &RasterQueryRectangle,
+    ) -> Option<CacheTileStream<T>>
+    where
+        T: Pixel + Cachable,
+    {
+        let (cache, _cache_size, cache_lru) = self.cache_and_size_and_lru_mut(key)?;
```
michaelmattig, line 207: make a custom method withotu _cache_size if it's ignored?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+
+        res
+    }
+
+    pub fn insert_query<T: Pixel + Cachable>(
+        &mut self,
+        key: &CanonicOperatorName,
+        query: &RasterQueryRectangle,
+    ) -> Result<QueryId, CacheError> {
+        // if there is no cache for this operator graph, we create one. However, it is not clear yet, if we will actually cache anything.
```
michaelmattig, line 237: why not? You mean because the query insertion may be aborted?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+        if !self.operator_caches.contains_key(key) {
+            self.operator_caches.insert(key.clone(), Default::default());
+            // self.cache_size.try_add_element_bytes(&key)?; // TODO: can we delete this at a later point?
```
michaelmattig, line 240: the query should be either finished or aborted, either during tile insertion (if a tile is expired) or explicitly when the wrapped stream of the cache operator is dropped.

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+            .landing_zone_and_size_mut(key)
+            .expect("Landing zone must must exist since we just created it");
+
+        // This gets or creates the cache entry for the operator graph. If it already exists, we don't need to add the size of the key again (see if case above).
+        landing_zone_size.try_add_element_bytes(&query_id)?;
+        landing_zone_size.try_add_element_bytes(&landing_zone_entry)?;
+        let old_entry = landing_zone.insert(query_id, landing_zone_entry);
+
+        if let Some(old_entry) = old_entry {
+            debug!("Replacing old entry in landing zone!");
```
michaelmattig, line 260: can this ever happen? The query id is randomly generated. If it indeed already exists, we should throw an error instead of replacing the old query. We do not have to handle this error really though, because it should never happen in practice.

operators/src/pro/cache/tile_cache.rs
michaelmattig: > clippy wanted that type

why though? Why doesn't it let you access two fields of you struct? What was the error message?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+            landing_zone_size.remove_element_bytes(&old_entry); // Since the old entry already was in the landing zone, we need to remove it from the size.
+        }
+
+        Ok(query_id)
+    }
+
+    fn remove_query_from_landing_zone(&mut self, key: &CanonicOperatorName, query_id: &QueryId) {
```
michaelmattig, line 267: would it make sense to delete the whole `operator_caches` entry if the entry would be empty (no landing zone entries and no cache entries)?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+
+        let (landing_zone, landing_zone_size) = self
+            .landing_zone_and_size_mut(key)
+            .ok_or(super::error::CacheError::QueryNotFoundInLandingZone)?;
+
+        let entry = landing_zone
+            .get_mut(query_id)
+            .ok_or(super::error::CacheError::QueryNotFoundInLandingZone)?;
+
+        entry.query.spatial_bounds.extend(&tile.spatial_partition()); // since the source should only produce tiles that intersect with the query, we can extend the query bounds
```
michaelmattig, line 322: can you maybe improve the comment here a bit? Like, that we are not using the spatial_bounds of the query but of the result, which may be larger then the original query because we always compute full tiles?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+        // this should always work, because the query was inserted at some point and then the cache entry was created
+        let (landing_zone, landing_zone_size) = self
+            .landing_zone_and_size_mut(key)
+            .ok_or(super::error::CacheError::QueryNotFoundInLandingZone)?;
+
+        let active_query = landing_zone
+            .remove(query_id)
+            .ok_or(super::error::CacheError::QueryNotFoundInLandingZone)?;
+
+        // remove the size of the query from the landing zone
```
michaelmattig, line 362: maybe put it next to the evict_all_elements... method so that the size computations happens all in one place?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -29,16 +30,425 @@ pub struct TileCacheBackend {
...
+            .expect("There must be cache elements entry if there was a loading zone");
+
+        let old_cache_entry = cache.insert(cache_entry_id, cache_entry);
+        let old_lru_entry = cache_lru.push(cache_entry_id, key.clone());
+        assert!(old_lru_entry.is_none()); // this should always work, because we just inserted a new CacheEntryId
+        assert!(old_cache_entry.is_none()); // this should always work, because we just inserted a new CacheEntryId
+
+        // cache bound can be temporarily exceeded as the entry is moved form the landing zone into the cache
+        // but the total of cache + landing zone is still below the bound
+        cache_size.add_bytes_allow_overflow(cache_entry_size_bytes);
```
michaelmattig, line 390: you could, instead, have a method like `self.evict_elements_to_free_n_bytes(cache_entry_size_bytes)` and then add the `cache_entry_size_bytes` regularly.

Otherwise: is overflow the right notion here? maybe just call it unchecked? or without_bound_check?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -435,218 +809,59 @@ impl TileCache {
...
-        }
-
-        // unpack match if there is one
-        let (entry_id, typed_stream) = result?;
-
-        // set as most recently used
-        backend.lru.promote(&entry_id);
-
-        T::stream(typed_stream)
+        backend.query_cache::<T>(key, query)
```
michaelmattig, line 819: if you move all the logic to the backend, do we really need the `TileCache` type anymore?

operators/src/pro/cache/tile_cache.rs
```diff
@@ -714,13 +926,18 @@ mod tests {
     #[tokio::test]
     async fn it_evicts_lru() {
         // Create cache entry and landing zone entry to geht the size of both
+        //let op0 = op(0);
```
michaelmattig, line 929: can be removed?

operators/src/pro/cache/tile_cache.rs
jdroenner: i think this is a comment from your code

operators/src/pro/cache/tile_cache.rs
jdroenner: u guess it is a leftover from debuging

operators/src/pro/cache/tile_cache.rs
jdroenner: that depends. If we account for the "OperatorCache" then we need that.

datatypes/src/primitives/spatial_partition.rs
jdroenner: [c170029](https://github.com/geo-engine/geoengine/pull/819/commits/c170029eecd40f6e4cc17d086b59812dea8c4f30)

operators/src/pro/cache/tile_cache.rs
jdroenner: [0739e8b](https://github.com/geo-engine/geoengine/pull/819/commits/0739e8b3e781ca3c6444c168e43f815b21ff6d9d)

operators/src/pro/cache/tile_cache.rs
jdroenner: yes, will try to do that when i finally start building the vector cache

operators/src/pro/cache/tile_cache.rs
jdroenner: i guess thats an open discussion at the moment?

operators/src/pro/cache/tile_cache.rs
jdroenner: i removed the methods

jdroenner: [0739e8b](https://github.com/geo-engine/geoengine/pull/819/commits/0739e8b3e781ca3c6444c168e43f815b21ff6d9d)

operators/src/pro/cache/tile_cache.rs
jdroenner: will try to do that when building the vector cache

operators/src/pro/cache/tile_cache.rs
jdroenner: [4e31afd](https://github.com/geo-engine/geoengine/pull/819/commits/4e31afded177a8a3465ddf09f665b7ea891868e3)

operators/src/pro/cache/tile_cache.rs
jdroenner: d06117e815802be648856c9b0850248e4eca1fcc

operators/src/pro/cache/tile_cache.rs
jdroenner: 587fb99972ca447a9052573250432a5023946202

operators/src/pro/cache/tile_cache.rs
jdroenner: 0789dd4330c15bc9f11f40c260d8493037b4c1ee

operators/src/pro/cache/tile_cache.rs
jdroenner: a0fb56f2e65b6ef68a5b30f60f1e6e459fc04b97 now it does

operators/src/pro/cache/tile_cache.rs
jdroenner: i don't know. Lets keep it until the next refactoring round. It makes things a bit more easy to read. The compiler will take care of all the overhead...

operators/src/pro/cache/tile_cache.rs
jdroenner: a285a865bc1432ceaa236ad0f1749031b9b5dc38

operators/src/pro/cache/tile_cache.rs
jdroenner: lets keep that as a reminder that we do not store the size of the TileCacheEntry Struct. I will look at this when i refactor the cache to also hold vector data.

operators/src/pro/cache/tile_cache.rs
jdroenner: yes but it is  not that easy since you can't borrow &mut self and  &mut self.something at the same time. So lets keep it for the next round. I think i removed key lookups at some other places.

operators/src/pro/cache/tile_cache.rs
jdroenner: also [7ed4001](https://github.com/geo-engine/geoengine/pull/819/commits/7ed40014e78b19ba5ecd79fd62570a35caa30ffe) and [38bdd50](https://github.com/geo-engine/geoengine/pull/819/commits/38bdd50817d933563bd94bdbff518a1d93b5558a)

michaelmattig (approved):

PR #822, opened 2023-07-10
Fix contains for TimeInterval (incl. TimeInstance)

jdroenner:

---

If the TimeInterval is a TimeInstance then the old contans method will return true for [1,2).contains([2,2)]  and thats wrong.

### DIFF
--- a/datatypes/src/primitives/time_interval.rs
+++ b/datatypes/src/primitives/time_interval.rs
@@ -167,7 +167,8 @@ impl TimeInterval {
     /// ```
     ///
     pub fn contains(&self, other: &Self) -> bool {
-        self.start <= other.start && self.end >= other.end
+        // TODO: should contains cover equals?
+        self == other || (self.start..self.end).contains(&other.start) && self.end >= other.end
     }
 
     /// Returns whether the given interval intersects this interval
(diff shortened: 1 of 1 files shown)


jdroenner:
[Contains für TimeIntervall ist falsch](https://trello.com/c/a2q3f6S4/1380-contains-f%C3%BCr-timeintervall-ist-falsch)

datatypes/src/primitives/time_interval.rs
```diff
@@ -167,7 +167,8 @@ impl TimeInterval {
     /// ```
     ///
     pub fn contains(&self, other: &Self) -> bool {
-        self.start <= other.start && self.end >= other.end
+        // TODO: should contains cover equals?
```
michaelmattig, line 170: I guess we should just conform to standardized semantics https://en.wikipedia.org/wiki/Allen%27s_interval_algebra

datatypes/src/primitives/time_interval.rs
```diff
@@ -167,7 +167,8 @@ impl TimeInterval {
...
+        self == other || (self.start..self.end).contains(&other.start) && self.end >= other.end
```
michaelmattig, line 171: can you put parentheses around the seconds part so that it's clearer how this is evaluated?

jdroenner: i can try later. I guess clippy will not be happy about that

datatypes/src/primitives/time_interval.rs
michaelmattig: there is no contains in the algebra. So I guess you would use `A == B OR A during B` instead of contains in an algorithm.

jdroenner: ok so lets keep it like that. We use it like this at the moment.

datatypes/src/primitives/time_interval.rs
jdroenner: [0277db9](https://github.com/geo-engine/geoengine/pull/822/commits/0277db9c0d62b3af008c94b3d19382ff5eea9c8f)

michaelmattig (approved):

PR #823, opened 2023-07-10
batch quota updates send to the quota manager

jdroenner:

---

<TEXT>


jdroenner:

PR #824, opened 2023-07-11
handle empty collections early in rastervectorjoin

jdroenner:

---

<TEXT>

### DIFF
--- a/operators/src/processing/raster_vector_join/non_aggregated.rs
+++ b/operators/src/processing/raster_vector_join/non_aggregated.rs
@@ -94,9 +94,7 @@ where
         ignore_no_data: bool,
     ) -> Result<BoxStream<'a, Result<FeatureCollection<G>>>> {
         if collection.is_empty() {
-            log::debug!(
-                "input collection is empty, returning empty collection, skipping raster query"
-            );
+            log::debug!("input collection is empty, returning empty collection, skipping raster query");
 
             return Self::collection_with_new_null_column(
                 &collection,
--- a/operators/src/processing/raster_vector_join/non_aggregated.rs
+++ b/operators/src/processing/raster_vector_join/non_aggregated.rs
@@ -94,7 +94,9 @@ where
         ignore_no_data: bool,
     ) -> Result<BoxStream<'a, Result<FeatureCollection<G>>>> {
         if collection.is_empty() {
-            log::debug!("input collection is empty, returning empty collection, skipping raster query");
+            log::debug!(
+                "input collection is empty, returning empty collection, skipping raster query"
+            );
 
             return Self::collection_with_new_null_column(
                 &collection,
@@ -117,7 +119,7 @@ where
             log::debug!(
                 "spatial or temporal intersection is empty, returning the same collection, skipping raster query"
             );
-            
+
             return Self::collection_with_new_null_column(
                 &collection,
                 new_column_name,
@@ -294,7 +296,7 @@ where
     }
 
     fn into_collection(self, new_column_name: &str) -> Result<FeatureCollection<G>> {
-        let Some(state) = self.state else {            
+        let Some(state) = self.state else {
             return Err(Error::EmptyInput); // TODO: maybe output empty dataset or just nulls
         };
         let mut new_collection = state
--- a/operators/src/processing/raster_vector_join/non_aggregated.rs
+++ b/operators/src/processing/raster_vector_join/non_aggregated.rs
@@ -93,36 +93,70 @@ where
         aggregation_method: FeatureAggregationMethod,
         ignore_no_data: bool,
     ) -> Result<BoxStream<'a, Result<FeatureCollection<G>>>> {
-        if !collection.is_empty() {
-            let bbox = collection
-                .bbox()
-                .and_then(|bbox| bbox.intersection(&query.spatial_bounds));
+        if collection.is_empty() {
+            log::debug!(
+                "input collection is empty, returning empty collection, skipping raster query"
+            );
 
-            let time = collection
-                .time_bounds()
-                .and_then(|time| time.intersect(&query.time_interval));
+            return Self::collection_with_new_null_column(
+                &collection,
+                new_column_name,
+                raster_processor.raster_data_type().into(),
+            );
+        }
 
-            // TODO: also intersect with raster spatial / time bounds
+        let bbox = collection
+            .bbox()
+            .and_then(|bbox| bbox.intersection(&query.spatial_bounds));
 
-            if let (Some(q_bbox), Some(q_time)) = (bbox, time) {
-                let query = VectorQueryRectangle {
-                    spatial_bounds: q_bbox,
-                    time_interval: q_time,
-                    spatial_resolution: query.spatial_resolution,
-                }
-                .into();
+        let time = collection
+            .time_bounds()
+            .and_then(|time| time.intersect(&query.time_interval));
 
-                let res_stream = call_on_generic_raster_processor!(raster_processor, raster_processor => {
-                    Self::process_typed_collection_chunk(collection, raster_processor, new_column_name, query, ctx, aggregation_method, ignore_no_data).await
-                });
-                return res_stream;
-            }
+        // TODO: also intersect with raster spatial / time bounds
+
+        let (Some(spatial_bounds), Some(time_interval)) = (bbox, time) else {
+            log::debug!(
+                "spatial or temporal intersection is empty, returning the same collection, skipping raster query"
+            );
+            
+            return Self::collection_with_new_null_column(
+                &collection,
+                new_column_name,
+                raster_processor.raster_data_type().into(),
+            );
         };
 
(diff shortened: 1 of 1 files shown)


jdroenner:
[handle empty collections early in RasterVectorJoin](https://trello.com/c/zWKWWhSR/1384-handle-empty-collections-early-in-rastervectorjoin)

operators/src/processing/raster_vector_join/non_aggregated.rs
```diff
@@ -256,6 +263,7 @@ where
 
     fn into_collection(self, new_column_name: &str) -> Result<FeatureCollection<G>> {
         let Some(state) = self.state else {
+            log::warn!("Empty input for vector raster join");
```
ChristianBeilschmidt, line 266: Do we need an extra warning when we throw an error?

operators/src/processing/raster_vector_join/non_aggregated.rs
```diff
@@ -288,6 +296,7 @@ where
         for (raster_processor, new_column_name) in
             self.raster_processors.iter().zip(&self.column_names)
         {
+            log::debug!("processing raster for new column {:?}", new_column_name);
```
ChristianBeilschmidt, line 299:
```suggestion
            log::debug!("processing raster for new column {:?}", new_column_name);
            
```

operators/src/processing/raster_vector_join/non_aggregated.rs
```diff
@@ -93,61 +93,68 @@ where
...
+                move |accum, raster| {
+                    let collection = collection.clone();
+                    async move {
+                        let accum = accum?;
+                        let raster = raster?;
+                        accum.extract_raster_values(&collection, &raster)
+                    }
+                },
+            )
+            .map(move |accum| accum?.into_collection(new_column_name));
```
ChristianBeilschmidt, line 155: Why did you (re)move the logic of modifying the boxes, etc.?

I would put it back since you did not win anything. In the block, where it is now, you handle two branches. Thus, it is better to read if you only see that I guess.

operators/src/processing/raster_vector_join/non_aggregated.rs
```diff
@@ -93,61 +93,68 @@ where
         aggregation_method: FeatureAggregationMethod,
         ignore_no_data: bool,
     ) -> Result<BoxStream<'a, Result<FeatureCollection<G>>>> {
-        call_on_generic_raster_processor!(raster_processor, raster_processor => {
-            Self::process_typed_collection_chunk(collection, raster_processor, new_column_name, query, ctx, aggregation_method, ignore_no_data).await
-        })
+        if !collection.is_empty() {
```
ChristianBeilschmidt, line 96: The early return case should be the empty case and not the default case.

operators/src/processing/raster_vector_join/non_aggregated.rs
```diff
@@ -93,61 +93,68 @@ where
...
+                .into();
+
+                let res_stream = call_on_generic_raster_processor!(raster_processor, raster_processor => {
+                    Self::process_typed_collection_chunk(collection, raster_processor, new_column_name, query, ctx, aggregation_method, ignore_no_data).await
+                });
+                return res_stream;
+            }
+        };
+
+        log::debug!("input collection is empty, returning empty collection, skipping raster query");
```
ChristianBeilschmidt, line 122:
```suggestion
        log::debug!("input collection is empty, returning empty collection, skipping raster query");
        
```

ChristianBeilschmidt (changes requested):

operators/src/processing/raster_vector_join/non_aggregated.rs
jdroenner: no its a debug leftover. Will remove it later.

operators/src/processing/raster_vector_join/non_aggregated.rs
jdroenner: yes,but if there is no intersection in time or space we also can just return the collection with a null column. If i handle the empty case as an early return, then i have to duplicate the code for that.

operators/src/processing/raster_vector_join/non_aggregated.rs
jdroenner: it will at least save some generated code.  Building the query before we switch to the code path with generics will result in less generic code. Also in the macro we do a match block so there are also multiple branches.

While the empty collection will not be the default case, There will be many cases where there is no intersection in time and space.

operators/src/processing/raster_vector_join/non_aggregated.rs
jdroenner: do you think:

```

if bbox.is_none() or time.is_none() or collection.is_empty() {
// return early
}

bbox = bbox.unwrap()
time = time.unwrap()

// default case
```
is better? This would also calculate the bbox and the time intersection even if the collection is empty.
If you have an idea how refactor please tell me. We could crate a  new method  that we call but thats just more code for the same effect...

operators/src/processing/raster_vector_join/non_aggregated.rs
```diff
@@ -93,61 +93,68 @@ where
...
+        let data = FeatureDataType::from(raster_processor.raster_data_type())
+            .null_feature_data(collection.len());
+        let collection = collection.add_column(new_column_name, data)?;
+        let collection_stream = once_stream(async move { Ok(collection) }).boxed();
+        Ok(collection_stream)
```
ChristianBeilschmidt, line 127:
```suggestion
        if collection.is_empty() {
            log::debug!(
                "input collection is empty, returning empty collection, skipping raster query"
            );

            return Self::collection_with_new_null_column(
                &collection,
                new_column_name,
                raster_processor.raster_data_type().into(),
            );
        }

        let bbox = collection
            .bbox()
            .and_then(|bbox| bbox.intersection(&query.spatial_bounds));

        let time = collection
            .time_bounds()
            .and_then(|time| time.intersect(&query.time_interval));

        // TODO: also intersect with raster spatial / time bounds

        let (Some(spatial_bounds), Some(time_interval)) = (bbox, time) else {
            log::debug!(
                "spatial or temporal intersection is empty, returning the same collection, skipping raster query"
            );
            
            return Self::collection_with_new_null_column(
                &collection,
                new_column_name,
                raster_processor.raster_data_type().into(),
            );
        };

        let query = VectorQueryRectangle {
            spatial_bounds,
            time_interval,
            spatial_resolution: query.spatial_resolution,
        }
        .into();

        call_on_generic_raster_processor!(raster_processor, raster_processor => {
            Self::process_typed_collection_chunk(
                collection,
                raster_processor,
                new_column_name,
                query,
                ctx,
                aggregation_method,
                ignore_no_data,
            )
            .await
        })
    }

    fn collection_with_new_null_column<'a>(
        collection: &FeatureCollection<G>,
        new_column_name: &str,
        feature_data_type: FeatureDataType,
    ) -> Result<BoxStream<'a, Result<FeatureCollection<G>>>> {
        let collection = collection.add_column(
            new_column_name,
            feature_data_type.null_feature_data(collection.len()),
        )?;
        let collection_stream = once_stream(async move { Ok(collection) }).boxed();
        Ok(collection_stream)
```

ChristianBeilschmidt: I find this kind of clean.

operators/src/processing/raster_vector_join/non_aggregated.rs
ChristianBeilschmidt: I've added a suggestion below.
This should compile and successfully run the tests.

ChristianBeilschmidt (changes requested):

operators/src/processing/raster_vector_join/non_aggregated.rs
jdroenner: i don't really like the 'let (Some(spatial_bounds), Some(time_interval)) = (bbox, time) else {' because it is easy to overlook the "if" case.

operators/src/processing/raster_vector_join/non_aggregated.rs
jdroenner: [68e415f](https://github.com/geo-engine/geoengine/pull/824/commits/68e415f0df584517e614e3d889c52ce6f727c373)

operators/src/processing/raster_vector_join/non_aggregated.rs
jdroenner: i tried that but git thinks there is no difference

operators/src/processing/raster_vector_join/non_aggregated.rs
jdroenner: while i think that the other way around it was less code i am fine with your suggestion.
It did not compile because of whitespaces that fmt was complaining about.

ChristianBeilschmidt (approved):

PR #825, opened 2023-07-13
quota mode and update buffering

michaelmattig:

---

<TEXT>

### DIFF
--- a/services/src/pro/contexts/postgres.rs
+++ b/services/src/pro/contexts/postgres.rs
@@ -14,7 +14,7 @@ use crate::pro::permissions::Role;
 use crate::pro::quota::{initialize_quota_tracking, QuotaTrackingFactory};
 use crate::pro::tasks::{ProTaskManager, ProTaskManagerBackend};
 use crate::pro::users::{OidcRequestDb, UserAuth, UserSession};
-use crate::pro::util::config::{Cache, Oidc};
+use crate::pro::util::config::{Cache, Oidc, Quota};
 
 use crate::tasks::SimpleTaskManagerContext;
 use crate::util::config::get_config_element;
@@ -76,6 +76,7 @@ where
         tls: Tls,
         exe_ctx_tiling_spec: TilingSpecification,
         query_ctx_chunk_size: ChunkByteSize,
+        quota_config: Quota,
     ) -> Result<Self> {
         let pg_mgr = PostgresConnectionManager::new(config, tls);
 
@@ -83,7 +84,12 @@ where
         Self::update_schema(pool.get().await?).await?;
 
         let db = PostgresDb::new(pool.clone(), UserSession::admin_session());
-        let quota = initialize_quota_tracking(db);
+        let quota = initialize_quota_tracking(
+            quota_config.mode,
+            db,
+            quota_config.increment_quota_buffer_size,
+            quota_config.increment_quota_buffer_timeout_seconds,
+        );
 
         Ok(PostgresContext {
             task_manager: Default::default(),
@@ -111,6 +117,7 @@ where
         query_ctx_chunk_size: ChunkByteSize,
         oidc_config: Oidc,
         cache_config: Cache,
+        quota_config: Quota,
     ) -> Result<Self> {
         let pg_mgr = PostgresConnectionManager::new(config, tls);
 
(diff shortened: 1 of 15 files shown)


michaelmattig:

services/src/pro/quota/mod.rs
```diff
@@ -49,47 +56,119 @@ impl TestDefault for QuotaTrackingFactory {
...
-                let user = UserId(computation.issuer);
-                // TODO: what to do if this fails (quota can't be recorded)? Try again later?
-                let r = self.user_db.increment_quota_used(&user, 1).await;
-
-                if r.is_err() {
-                    log::error!("Could not increment quota for user {}", user);
+            while let Some(message) = self.quota_receiver.recv().await {
+                let flush_buffer = match message {
+                    QuotaMessage::ComputationUnit(computation) => {
+                        // TODO: issue a tracing event instead?
```
jdroenner, line 106: you do send a tracing message, right?

services/src/pro/contexts/postgres.rs
```diff
@@ -880,6 +893,7 @@ mod tests {
                             tokio_postgres::NoTls,
                             TestDefault::test_default(),
                             TestDefault::test_default(),
+                            get_config_element::<Quota>().unwrap(),
```
jdroenner, line 896: this method is only used in tests, right? Is it a good idea to use the settings.toml for tests?

services/src/pro/handlers/datasets.rs
```diff
@@ -328,6 +331,7 @@ mod tests {
             exe_ctx_tiling_spec,
             TestDefault::test_default(),
             TestDefault::test_default(),
+            get_config_element::<Quota>().unwrap(),
         );
```
jdroenner, line 335: does this test need to read the config.toml? can't we use a test_default here?

services/src/pro/handlers/users.rs
```diff
@@ -1255,7 +1256,11 @@ mod tests {
         let server_url = format!("http://{}", server.addr());
         let request_db = single_state_nonce_request_db(server_url);
 
-        let ctx = ProInMemoryContext::new_with_oidc(request_db, TestDefault::test_default());
+        let ctx = ProInMemoryContext::new_with_oidc(
+            request_db,
+            TestDefault::test_default(),
+            get_config_element::<crate::pro::util::config::Quota>().unwrap(),
+        );
```
jdroenner, line 1263: does this test need to read the config.toml? can't we use a test_default here?

services/src/pro/handlers/users.rs
```diff
@@ -1286,7 +1291,11 @@ mod tests {
             ),
         );
 
-        let ctx = ProInMemoryContext::new_with_oidc(request_db, TestDefault::test_default());
+        let ctx = ProInMemoryContext::new_with_oidc(
+            request_db,
+            TestDefault::test_default(),
+            get_config_element::<crate::pro::util::config::Quota>().unwrap(),
+        );
```
jdroenner, line 1298: does this test need to read the config.toml? can't we use a test_default here?
Same for all the other test where you now use get_config_element...

services/src/pro/users/postgres_userdb.rs
```diff
@@ -50,7 +50,7 @@ where
             .await?;
 
         let quota_available =
-            crate::util::config::get_config_element::<crate::pro::util::config::User>()?
+            crate::util::config::get_config_element::<crate::pro::util::config::Quota>()?
```
jdroenner, line 53: I know it was like that before but, can't we store that in the context? Its an i64 so it will not really hurt. We have to think about updating things from the settings separatly.

services/src/pro/contexts/postgres.rs
michaelmattig: that's what the Settings-test.toml is there for.

services/src/pro/handlers/datasets.rs
michaelmattig: the config for the tests is in Settings-test.toml. We don't do it uniformly, but e.g. the Postgres config we read from the Settings-test.toml while other config elements also implement the TestDefault trait.  But I think it's not ideal to have two places for setting the test configuration.

services/src/pro/users/postgres_userdb.rs
michaelmattig: We could. But then do we also store the admin username and password there?
In general, the config is, in my understanding, just a typed map. So looking something up costs nothing. And it is easily extensible with new fields without having to change dozens of tests.

jdroenner (approved):

PR #836, opened 2023-07-24
add bulk updates for quotas

michaelmattig:

---

<TEXT>

### DIFF
--- a/services/src/pro/quota/mod.rs
+++ b/services/src/pro/quota/mod.rs
@@ -128,12 +128,12 @@ impl<U: UserDb + 'static> QuotaManager<U> {
 
                 if flush_buffer {
                     // TODO: what to do if this fails (quota can't be recorded)? Try again later?
-                    for (user, quota_used) in self.increment_quota_buffer.drain() {
-                        let r = self.user_db.increment_quota_used(&user, quota_used).await;
+                    let quota_updates = self.increment_quota_buffer.drain().collect::<Vec<_>>();
 
-                        if r.is_err() {
-                            log::error!("Could not increment quota for user {}", user);
-                        }
+                    let r = self.user_db.bulk_increment_quota_used(&quota_updates).await;
+
+                    if r.is_err() {
+                        log::error!("Could not increment quota for users {:?}", quota_updates);
                     }
                     self.increment_quota_buffer_used = 0;
                 }
--- a/services/src/pro/quota/mod.rs
+++ b/services/src/pro/quota/mod.rs
@@ -128,13 +128,15 @@ impl<U: UserDb + 'static> QuotaManager<U> {
 
                 if flush_buffer {
                     // TODO: what to do if this fails (quota can't be recorded)? Try again later?
-                    let quota_updates = self.increment_quota_buffer.drain().collect::<Vec<_>>();
 
-                    let r = self.user_db.bulk_increment_quota_used(&quota_updates).await;
-
-                    if r.is_err() {
-                        log::error!("Could not increment quota for users {:?}", quota_updates);
+                    if let Err(error) = self
+                        .user_db
+                        .bulk_increment_quota_used(self.increment_quota_buffer.drain())
+                        .await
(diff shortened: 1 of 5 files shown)


michaelmattig:

services/src/pro/users/postgres_userdb.rs
```diff
@@ -509,6 +509,33 @@ where
...
+            users.push(*user);
+            quotas.push(*quota as i64);
+        }
+
+        let query = "
+            UPDATE users
+            SET quota_available = quota_available - quota_changes.quota, 
+                quota_used = quota_used + quota_changes.quota
+            FROM 
+                (SELECT * FROM UNNEST($1::uuid[], $2::bigint[]) AS t(id, quota)) AS quota_changes
```
ChristianBeilschmidt, line 530: ouldn't you insert a two-dim array instead of pushing two arrays.

services/src/pro/quota/mod.rs
```diff
@@ -128,12 +128,12 @@ impl<U: UserDb + 'static> QuotaManager<U> {
 
                 if flush_buffer {
                     // TODO: what to do if this fails (quota can't be recorded)? Try again later?
-                    for (user, quota_used) in self.increment_quota_buffer.drain() {
-                        let r = self.user_db.increment_quota_used(&user, quota_used).await;
+                    let quota_updates = self.increment_quota_buffer.drain().collect::<Vec<_>>();
```
ChristianBeilschmidt, line 131: Do you convert a HashMap from user_id -> number of chunks to an array.
Then, you split the array into two arrays and put it into the db an then unnest it?

You could stick with the map, right?
Or you could insert keys() and values() so that you don't need to copy the vec twice.

ChristianBeilschmidt (changes requested):

services/src/pro/users/postgres_userdb.rs
michaelmattig: I tried passing it as a Vec of tuples, but that doesn't work. We'd have to implement ToSql for the type `(UserId, i64)`

services/src/pro/quota/mod.rs
michaelmattig: it could certainly be simplified. But I think u64 makes more sense for the interface (trait method), as the used quota should never be negative. But: Postgres has no unsigned datatype, so we have to convert it to i64 before executing the query.

> You could stick with the map, right?

I don't think it's possible to insert a map into postgres, if that's what you mean. But if you have an idea let me know.

> Or you could insert keys() and values() so that you don't need to copy the vec twice.

I don't think it's that easy, because then we would lose the correspondence between key and value. We need to collect the entries from the hasmap I think. Splitting it up into separate Vecs could be avoided I guess but I don't really think it's worth it.

ChristianBeilschmidt: Ok, then just change the parameter to be an Iter<(UserId, u64)>, then you don't have to create a vec and then create two new vecs.

michaelmattig: what do you mean with `Iter<(UserId, u64)>`? 

Do you mean `std::collections::hash_map::Iter`? Then we'd be limited to HashMaps and could pass the HashMap as it is.

Or do you mean `Iterator`? That's a trait, which means we would have to make the function generic, or pass a fat pointer.

ChristianBeilschmidt: You can just make it `function<I: Iterator<Item=(UserId, u64)>>`.

michaelmattig: yeah that's what I meant. It would make the function generic, leading to more code being generated and restricting the use of this trait e.g. for usage as a trait object. Is it really worth it?

michaelmattig: 8ac8eafc4fe835edc5611a01f9f00ac1421a6b0d

ChristianBeilschmidt (approved):

PR #840, opened 2023-07-27
Remove in memory contexts and dbs

michaelmattig:

---

<TEXT>


michaelmattig:
[Entfernen der InMemory-DBs](https://trello.com/c/SMI0ORTw/1387-entfernen-der-inmemory-dbs)

services/src/pro/apidoc.rs
```diff
@@ -408,28 +408,24 @@ impl Modify for ApiDocInfo {
...
+    use crate::pro::util::tests::with_pro_temp_context;
 
     #[test]
     fn can_resolve_api() {
-        crate::api::can_resolve_api(ApiDoc::openapi());
+        crate::util::openapi_examples::can_resolve_api(ApiDoc::openapi());
     }
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
ChristianBeilschmidt, line 419: didn't you want to create a macro that captures test, tokio and the pg context?

services/src/pro/handlers/drone_mapping.rs
```diff
@@ -396,219 +394,226 @@ mod tests {
...
+            );
+
+            // manipulate config to use the mock nodeodm server
+            config::set_config("odm.endpoint", mock_nodeodm.url_str("/")).unwrap();
+
+            let session = create_session_helper(&app_ctx).await;
+            let ctx = app_ctx.session_context(session.clone());
+
+            // file upload into geo engine
+            // TODO: properly upload the data using the handler once this is possible in a test (after migration to actix)
```
ChristianBeilschmidt, line 436: is this easy to do?

we did the migration to actix some year(s) ago

services/src/pro/util/openapi_examples.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Runs all example requests against the provided test server to check for bad documentation,
+/// for example due to incompatible schema changes between the time of writing the request body
+/// and now. It can also detect if the query parameters are not documented correctly or the
+/// request path changed.
+///
+/// # Panics
+///
+/// panics if the creation of an anonymous session fails.
+#[allow(clippy::unimplemented)]
```
ChristianBeilschmidt, line 18:
```suggestion
```

services/src/server.rs
```diff
@@ -193,83 +165,3 @@ where
...
-    async fn wait_for_server(base_url: &Url) -> bool {
-        for _ in 0..WAIT_SERVER_RETRIES {
-            if reqwest::get(base_url.clone()).await.is_ok() {
-                return true;
-            }
-            std::thread::sleep(std::time::Duration::from_secs(WAIT_SERVER_RETRY_INTERVAL));
-        }
-        false
-    }
-}
```
ChristianBeilschmidt, line 275: why are the tests gone?

Settings-default.toml
```diff
@@ -29,6 +28,7 @@ database = "geoengine"
 schema = "public"
 user = "geoengine"
 password = "geoengine"
+clear_database_on_start = false # set to `true` to start with a fresh database every time. This has to be set at the very first startup with an empty database.
```
ChristianBeilschmidt, line 31: we should log if this is set to true but we ignore it for security reasons

services/src/error.rs
```diff
@@ -128,6 +128,11 @@ pub enum Error {
 
     TokioPostgresTimeout,
```
ChristianBeilschmidt, line 130: When you start the data atlas, you get

```
error | "IdStringMustBeUuid"
-- | --
message | "IdStringMustBeUuid"
```

This worked in the InMemory-Context.

ChristianBeilschmidt (changes requested):

services/src/api/mod.rs
```diff
@@ -1,716 +1 @@
...
-        panic!("Invalid reference type");
-    }
-}
-
-/// Recursively checks that schemas referenced in the given schema object exist in the provided map.
-fn can_resolve_schema(schema: RefOr<Schema>, components: &Components) {
-    match schema {
-        RefOr::Ref(reference) => {
-            throw_if_invalid_ref(&reference, components);
-        }
```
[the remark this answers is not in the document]
michaelmattig: I moved this to util/openapi_examples and pro/util/openapi_examples. This didn't belong here imho.

services/src/pro/apidoc.rs
michaelmattig: I thought about it, but I wasn't really sure about how to do it, because the #[toko::test] is also a macro, so we'd have nested macros, and also: tests can't have arguments, so I wasn't really sure how to properly implement this.

That's why I stuck with the current approach of multi_thread flavor and the with_temp_context methods.

services/src/pro/handlers/drone_mapping.rs
michaelmattig: I think this is possible now, yes. But we wanted to get rid of this handler entirely, or not?

services/src/server.rs
michaelmattig: IMHO these should be integration tests, and not unit tests. I didn't find a way to make these work with the temp context, so instead, I extended the services/tests/startup.rs test to no only check for warnings, but also to check whether the webserver can be connected after startup.

Settings-default.toml
michaelmattig: I do log it at info level on startup

services/src/pro/util/openapi_examples.rs
michaelmattig: 08d1d3105dc77489265f978557567d7cdee0e4cd

services/src/contexts/postgres.rs
```diff
@@ -50,6 +51,11 @@ where
     volumes: Volumes,
 }
 
+enum DatabaseStatus {
+    Unitialized,
+    Initialized { clear_database_on_start: bool },
```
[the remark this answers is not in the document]
michaelmattig: de6bc503b260eeb3bf068f948a1a68341eafc7cf

services/src/handlers/projects.rs
```diff
@@ -583,157 +605,170 @@ mod tests {
...
     }
 
-    async fn update_test_helper(
-        method: Method,
-    ) -> (InMemoryContext, SimpleSession, ProjectId, ServiceResponse) {
-        let app_ctx = InMemoryContext::test_default();
+    // async fn update_test_helper(
+    //     method: Method,
+    // ) -> (InMemoryContext, SimpleSession, ProjectId, ServiceResponse) {
+    //     with_temp_context(|app_ctx, _| async move {
```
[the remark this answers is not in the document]
michaelmattig: I overlooked it: e55d12bda3fcaa02ab51fa24445a2c7947408e85

services/src/error.rs
michaelmattig: the error is:

> IdStringMustBeUuid { found: "c6f78d3b-609b-454f-8ad5-ed6ec12a9ac0c" }

https://www.freecodeformat.com/validate-uuid-guid.php says it is in fact not a valid uuid. I am not sure why.

The in memory backend didn't check this. So I guess we need to change the dataset-defs and the frontend. Or do you see another solution?

ChristianBeilschmidt: Hm, but this site (https://cloud.magiclen.org/us/uuid/checker) says it is valid. Maybe duckduckgo creates uuids that are only valid for certain validators :thinking: 

Can we get more info on that? What does postgres check?

services/src/pro/handlers/drone_mapping.rs
ChristianBeilschmidt: :+1:

services/src/pro/apidoc.rs
ChristianBeilschmidt: * because the #[toko::test] is also a macro
   *  -> in the worst case, we just use the code that is created by the macro. it is just some lines.
 * tests can't have arguments
   * -> our tests could. we would just put our test method with arguments inside a test method

michaelmattig: I think it's possible to do this, but I would vote for doing this in a separate task.

Settings-default.toml
michaelmattig: with clear = true on a database with clear = true
image

with clear = true on a datbase with clear = false
image

services/src/error.rs
michaelmattig: It's not postgres, but our Rust code that checks whether the layer/collection id we receive in the API as a string, is im fact a UUID:

```
  #[test]
    fn it_is_a_uuid() {
        uuid::Uuid::from_str("c6f78d3b-609b-454f-8ad5-ed6ec12a9ac0c").unwrap();
    }
```

> thread 'contexts::tests::it_is_a_uuid' panicked at 'called `Result::unwrap()` on an `Err` value: Error(GroupLength { group: 4, len: 13, index: 25 })', services/src/contexts/mod.rs:523:71

Settings-default.toml
ChristianBeilschmidt: 1. it should be an error if `clear = false` at the first start and `clear = true` at a consecutive one.
2. The error does not state what went wrong.

services/src/pro/handlers/drone_mapping.rs
ChristianBeilschmidt: if it is easy to do, yes

services/src/pro/apidoc.rs
ChristianBeilschmidt: Yeah, let's get this merged first.

services/src/pro/handlers/drone_mapping.rs
michaelmattig: 4bf61de4aadd39d243da6547f8505033e190573a

Settings-default.toml
michaelmattig: 1. that's what I meant and what I did
2. it does state that ClearDatabaseOnStartupNotAllowed. I don't actually know why it doesn't print my snafu display message that explain it in more detail. But that's more of a general error handling issue, than an issue with my PR.

services/src/error.rs
michaelmattig: >  error: SerdeJson: invalid type: map, expected a string consisting of a namespace and name name, separated by a colon, only using alphanumeric characters, underscores & dashes at line 22 column 3

the layer-def in kailin_hatlestad.json still used the old datasetIds. I changed it in the dataset-defs PR

Settings-default.toml
ChristianBeilschmidt: I tried this:

1. Start with `clear = true`
2. Start with `clear = false`
3. Start with `clear = true`

no error

services/src/error.rs
ChristianBeilschmidt: you need to merge the main for dataset-defs in order to try this out

Settings-default.toml
michaelmattig: IMHO it is correct that there is no error. The first time you start the engine you determine whether clearOnStart is possible, and this is saved in the database.

If you then decide to not clear it the next time, it should be fine, everything survives.

If you then start again with clear, it should clear, because it is allowed in the database.

At least that's my thinking. It would protect production databases that initially set clear to true, but allows us to ad-hoc decide whether we want to truncate the data during development and testing.

Settings-default.toml
```diff
@@ -29,6 +28,9 @@ database = "geoengine"
 schema = "public"
 user = "geoengine"
 password = "geoengine"
+# Set to `true` to start with a fresh database every time.
+# This is ignored if the engine has ever started with this being `false`.
+clear_database_on_start = false
```
ChristianBeilschmidt, line 33: We have to document the behavior correctly so that it matches:

```suggestion
# Set to `true` to start with a fresh database each time.
# If this parameter is set to `false` at the first start, subsequent changes to `true` will result in an error.
# This helps to protect production environments from unwanted data loss.
clear_database_on_start = false
```

ChristianBeilschmidt (changes requested):

PR #844, opened 2023-07-31
Pg symbology mapping

ChristianBeilschmidt:

---

<TEXT>


ChristianBeilschmidt:
[Die JSON-Spalten, soweit möglich, als Postgres-Typen abbilden: Symbology](https://trello.com/c/MU7XGVHx/1396-die-json-spalten-soweit-m%C3%B6glich-als-postgres-typen-abbilden)

services/src/contexts/schema.sql
```diff
@@ -50,6 +50,109 @@ CREATE TYPE "TimeStep" AS (
...
+CREATE TYPE "DefaultColors" AS (
+    -- default
+    default_color "RgbaColor",
+    -- over/under
+    over_color "RgbaColor",
+    under_color "RgbaColor"
+);
+
+CREATE TYPE "ColorizerType" AS ENUM (
+    'linearGradient', 'logarithmicGradient', 'palette', 'rgba'
```
michaelmattig, line 79: The other enum variants in this files are capitalized. Should we do this uniformly?

services/src/contexts/schema.sql
```diff
@@ -88,7 +191,7 @@ CREATE TABLE project_version_layers (
     name character varying(256) NOT NULL,
     workflow_id uuid NOT NULL,
     -- TODO: REFERENCES workflows(id)
-    symbology json,
+    symbology "Symbology",
```
michaelmattig, line 194: what about constraints on the symbology? e.g. exactly one of raster/point/line/polygon should be not null. Also for all the other types we could constrain what fields are nullable and which not. Unfortunately postgres doesn't allow specifying this at the type level, so we would have to do it here. But, how far do we want to go? And what happens if e.g. a symbology has a raster and point symbology? Would we check that in the application? Would we just use the first one?

services/src/api/model/db_types.rs
```diff
@@ -0,0 +1,408 @@
...
+};
+
+use super::datatypes::{
+    Breakpoint, Colorizer, DefaultColors, LinearGradient, LogarithmicGradient, OverUnderColors,
+    Palette, RgbaColor,
+};
+
+#[derive(Debug, ToSql, FromSql)]
+#[postgres(name = "DefaultColors")]
+pub struct DefaultColorsDbType {
```
michaelmattig, line 19: does it make sense to have "Type" in the name? Isn't it obvious that this is a type?

services/src/api/model/db_types.rs
```diff
@@ -0,0 +1,408 @@
...
+                line: None,
+                polygon: Some(polygon),
+            } => Ok(Self::Polygon(polygon)),
+            _ => Err(Error::UnexpectedInvalidDbTypeConversion),
+        }
+    }
+}
+
+macro_rules! delegate_from_to_sql {
+    ( $normal:ident, $dbtype:ident ) => {
```
michaelmattig, line 372: make call it "rust" instead of "normale"

services/src/contexts/postgres.rs
```diff
@@ -2435,4 +2441,298 @@ mod tests {
         })
         .await;
     }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[allow(clippy::too_many_lines)]
+    async fn test_types_individually() {
```
michaelmattig, line 2447: what exactly do you test here? That you can insert it as a parameter and get it from a Row? Maybe you can write that in the the test name

services/src/api/model/db_types.rs
```diff
@@ -0,0 +1,408 @@
...
+                point: None,
+                line: None,
+                polygon: Some(polygon),
+            } => Ok(Self::Polygon(polygon)),
+            _ => Err(Error::UnexpectedInvalidDbTypeConversion),
+        }
+    }
+}
+
+macro_rules! delegate_from_to_sql {
```
michaelmattig, line 371: maybe write a quick comment for the future generation here what this does?
"Implements To/FromSql for a type by means of an intermediary database type. This is for cases where a Rust struct or enum cannot directly be represented by Postgres." Oder so.

services/src/api/model/db_types.rs
```diff
@@ -0,0 +1,408 @@
...
+    ( $normal:ident, $dbtype:ident ) => {
+        impl ToSql for $normal {
+            fn to_sql(
+                &self,
+                ty: &postgres_types::Type,
+                w: &mut bytes::BytesMut,
+            ) -> Result<postgres_types::IsNull, Box<dyn std::error::Error + Sync + Send>> {
+                <$dbtype as ToSql>::to_sql(&self.clone().into(), ty, w)
```
michaelmattig, line 379: why do you have to clone here?

services/src/contexts/schema.sql
ChristianBeilschmidt: we discussed this in Slack, we should propose "domain wrapping" to rust-postgres and go from there.

services/src/contexts/schema.sql
ChristianBeilschmidt: [a28413d](https://github.com/geo-engine/geoengine/pull/844/commits/a28413d25a68ac5b2eabdf4c1515ef267b60a5e6)

services/src/api/model/db_types.rs
ChristianBeilschmidt: [55eb248](https://github.com/geo-engine/geoengine/pull/844/commits/55eb24838bdf3d8627168210bdad2731e23b418b)

services/src/api/model/db_types.rs
ChristianBeilschmidt: [55eb248](https://github.com/geo-engine/geoengine/pull/844/commits/55eb24838bdf3d8627168210bdad2731e23b418b)

services/src/contexts/postgres.rs
ChristianBeilschmidt: [0d9f680](https://github.com/geo-engine/geoengine/pull/844/commits/0d9f680b8e88caac754d9c099106bb8a7f2fd9aa)

michaelmattig (approved):

PR #845, opened 2023-08-02
Pg-result-descriptor-mapping

ChristianBeilschmidt:

---

<TEXT>

### DIFF
--- a/services/src/contexts/schema.sql
+++ b/services/src/contexts/schema.sql
@@ -25,8 +25,8 @@ CREATE TYPE "BoundingBox2D" AS (
 );
 
 CREATE TYPE "TimeInterval" AS (
-    start timestamp with time zone,
-    "end" timestamp with time zone
+    start bigint,
+    "end" bigint
 );
 
 CREATE TYPE "STRectangle" AS (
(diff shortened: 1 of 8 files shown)


services/src/contexts/schema.sql
```diff
@@ -25,8 +25,8 @@ CREATE TYPE "BoundingBox2D" AS (
 );
 
 CREATE TYPE "TimeInterval" AS (
-    start timestamp with time zone,
-    "end" timestamp with time zone
+    start bigint,
```
michaelmattig, line 28: I ask myself: Why do we store a int here instead of a timestamp? I know it's because the ResultDescriptor has a TimeInterval, and our time interval uses ints to represent times, because it is efficient. 

However, while it makes sense to do this in e.g. the FeatureCollection, I don't think this is the case for the ResultDescriptor. There is only a single time interval, so maybe we can use a different type here that has actual datetimes? I always found it confusing that the API gives the time interval as timestamps rather than ISO-DateTimes. That's not how APIs usually handle datetimes.

So what do you think about changing the ResultDescriptor in using DateTimes instead of ints for the time interval? That would make the API nicer and also would make the schema better and would actually allow to perform queries on time more easily.

ChristianBeilschmidt: We cannot do it that easily.

Currently, TimeInterval::default() is not possible to store it as timestamp in postgres.
Thus, this would not help in having a postgres timestamp here.

Then, all timeinstances in ge must be convertible, otherwise, we cannot return a good result descriptor.

I'm not entirely against it. We would have to draw it out and see what changes in total, here.
But I wouldn't easily do it here in this PR.

michaelmattig (approved):

PR #847, opened 2023-08-03
layer metadata as sql type

ChristianBeilschmidt:

---

<TEXT>


ChristianBeilschmidt:
[JSON-Spalten als Postgres-Typen abbilden: layers-metadata](https://trello.com/c/hDjaCR09/1417-json-spalten-als-postgres-typen-abbilden-layers-metadata)

services/src/api/model/db_types.rs
```diff
@@ -364,6 +363,40 @@ impl TryFrom<SymbologyDbType> for Symbology {
...
+    }
+}
+
+impl<S: std::hash::BuildHasher + std::default::Default> From<HashMapTextTextDbType>
+    for HashMap<String, String, S>
+{
+    fn from(map: HashMapTextTextDbType) -> Self {
+        map.0
+            .into_iter()
+            .map(|TextTextKeyValue { key, value }| (key, value))
```
michaelmattig, line 395: what happens if a key occurs twice? There is no constraint that prevents this in postgres and it shouldn't crash our program. 

Do we need a test that checks this?

services/src/layers/postgres_layer_db.rs
```diff
@@ -663,7 +660,7 @@ where
             workflow: serde_json::from_value(row.get(2)).context(crate::error::SerdeJson)?,
             symbology: row.get(3),
             properties: row.get(4),
-            metadata: serde_json::from_value(row.get(5)).context(crate::error::SerdeJson)?,
+            metadata: row.get::<_, HashMapTextTextDbType>(5).into(),
```
michaelmattig, line 663: why no delegate here?

services/src/api/model/db_types.rs
ChristianBeilschmidt: First it does not crash: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d9616410a6c778fb727d182d6799e8f7

Second, since we only input hashmaps, I wouldn't check it for now in Postgres.

services/src/layers/postgres_layer_db.rs
ChristianBeilschmidt: I cannot delegate to HashMap (foreign type) and I didn't want to change the field type in the struct.

michaelmattig (approved):

PR #851, opened 2023-08-08
Pg dataset metadata mapping

ChristianBeilschmidt:

---

<TEXT>


ChristianBeilschmidt:
[JSON-Spalten als Postgres-Typen abbilden: Meta_data](https://trello.com/c/BowyvlyK/1415-json-spalten-als-postgres-typen-abbilden-metadata)

services/src/contexts/schema.sql
```diff
@@ -257,8 +257,257 @@ CREATE TYPE "ResultDescriptor" AS (
...
+    multi_line_string "MultiLineString",
+    multi_polygon "MultiPolygon"
+);
+
+CREATE TYPE "OgrSourceDataset" AS (
+    file_name text,
+    layer_name text,
+    data_type "VectorDataType",
+    "time" "OgrSourceDatasetTimeType",
+    default_geometry "TypedGeometry",
```
michaelmattig, line 385: Could we use postgres geom here instead?

services/src/api/model/db_types.rs
```diff
@@ -607,6 +616,960 @@ impl<S: std::hash::BuildHasher + std::default::Default> From<HashMapTextTextDbTy
...
+            attribute_query: other.attribute_query,
+            cache_ttl: other.cache_ttl,
+        })
+    }
+}
+
+#[derive(Debug, ToSql, FromSql)]
+#[postgres(name = "OgrMetaData")]
+pub struct OgrMetaDataDbType {
+    pub loading_info: OgrSourceDataset,
```
michaelmattig, line 1244: why not use the `OgrSourceDatasetDbType` here?

services/src/api/model/operators.rs
```diff
@@ -602,7 +602,7 @@ impl From<OgrSourceTimeFormat> for geoengine_operators::source::OgrSourceTimeFor
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize, ToSchema)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize, Serialize, ToSchema, ToSql, FromSql)]
```
michaelmattig, line 605: general question: Is it a good idea to use the API structs directly for storage in the DB? You also said you wanted to use the API types strictly in the handlers and then convert them to something else immediately.

services/src/datasets/postgres.rs
```diff
@@ -360,30 +357,30 @@ where
     <Tls as MakeTlsConnect<Socket>>::TlsConnect: Send,
     <<Tls as MakeTlsConnect<Socket>>::TlsConnect as TlsConnect<Socket>>::Future: Send,
 {
-    fn to_json(&self) -> Result<DatasetMetaDataJson> {
+    fn to_typed_metadata(&self) -> Result<DatasetMetaData> {
```
michaelmattig, line 360: do we still need this whole ordeal (meaning the PostgresStorable trait) if we no longer have the requirement that something is json serializable?

services/src/datasets/postgres.rs
```diff
@@ -260,11 +252,16 @@ where
...
-            OgrSourceDataset,
-            VectorResultDescriptor,
-            VectorQueryRectangle,
-        > = serde_json::from_value(row.get(0))?;
+        let meta_data: MetaDataDefinition = row.get("meta_data");
+
+        let MetaDataDefinition::OgrMetaData(meta_data) = meta_data else {
+            return Err(geoengine_operators::error::Error::MetaData {
+                source: Box::new(geoengine_operators::error::Error::InvalidType {
+                    expected: "OgrMetaData".to_string(),
```
michaelmattig, line 260: can you use the `type_name` method here?

services/src/pro/datasets/postgres.rs
```diff
@@ -279,11 +279,16 @@ where
...
-            OgrSourceDataset,
-            VectorResultDescriptor,
-            VectorQueryRectangle,
-        > = serde_json::from_value(row.get(0))?;
+        let meta_data: MetaDataDefinition = row.get("meta_data");
+
+        let MetaDataDefinition::OgrMetaData(meta_data) = meta_data else {
+                return Err(geoengine_operators::error::Error::MetaData {
+                    source: Box::new(geoengine_operators::error::Error::InvalidType {
+                        expected: "OgrMetaData".to_string(),
```
michaelmattig, line 287: also here: can you use the `type_name` method?

services/src/api/model/db_types.rs
ChristianBeilschmidt: I want to use them as few as possible. As long as a type implements FromSql/ToSql, we should use it.

I guess we discussed this already before :smile: .

In the worst case, I would have to create extra DbTypes to use them as fields. By doing it this way, we keep it as minimal as possible.

services/src/api/model/operators.rs
ChristianBeilschmidt: Yes, but currently, the API structs are (mostly) used to store them in the DB.

We should tackle this in another task as I don't want to change the API methods or internals other than the FromSql/ToSql part.

Shouldn't be much a hassle in the new task since the fields and everything has the same name.

services/src/datasets/postgres.rs
ChristianBeilschmidt: I wanted to get rid of it, but unfortunately, the API only offers you some Trait type that isn't the enum.
Thus, I still need the function to get something that implements FromSql/ToSql.

services/src/datasets/postgres.rs
ChristianBeilschmidt: Maybe you can give me a hint how to do that since I don't want to construct an OgrMetaData enum variant for that.

services/src/pro/datasets/postgres.rs
ChristianBeilschmidt: Maybe you can give me a hint how to do that since I don't want to construct an OgrMetaData enum variant for that.

services/src/contexts/schema.sql
ChristianBeilschmidt: [cfa23a9](https://github.com/geo-engine/geoengine/pull/851/commits/cfa23a912140f669287176a1ad78c7994e4a87a7)

services/src/api/model/datatypes.rs
```diff
@@ -1953,9 +2046,97 @@ impl From<MultiLineString> for geoengine_datatypes::primitives::MultiLineString
     }
 }
 
+impl ToSql for MultiLineString {
```
michaelmattig, line 2049: why do you not use: postgres_types::Type::PATH::LINE_ARRAY ?

services/src/api/model/db_types.rs
```diff
@@ -607,6 +617,928 @@ impl<S: std::hash::BuildHasher + std::default::Default> From<HashMapTextTextDbTy
...
+    }
+}
+
+#[derive(Debug)]
+pub struct PolygonRef<'p> {
+    pub rings: &'p [Vec<Coordinate2D>],
+}
+
+#[derive(Debug)]
+pub struct PolygonOwned {
```
michaelmattig, line 967: isn't this the same as a MultiLineString? Can you reuse it?

services/src/contexts/postgres.rs
```diff
@@ -3026,6 +3030,1186 @@ mod tests {
...
+                [MultiPoint {
+                    coordinates: vec![
+                        Coordinate2D::new(0.0f64, 0.5).into(),
+                        Coordinate2D::new(2., 1.0).into(),
+                    ],
+                }],
+            )
+            .await;
+
+            // test_type(
```
michaelmattig, line 3156: why is this commented out?

services/src/contexts/schema.sql
```diff
@@ -257,8 +257,245 @@ CREATE TYPE "ResultDescriptor" AS (
...
+    "datetime" text [],
+    rename "TextTextKeyValue" []
+);
+
+CREATE TYPE "OgrSourceErrorSpec" AS ENUM (
+    'Ignore',
+    'Abort'
+);
+
+-- closed path is a polygon without stored bbox
```
michaelmattig, line 357: what is a closed path and how is that comment relevant here?

services/src/api/model/db_types.rs
```diff
@@ -607,6 +617,928 @@ impl<S: std::hash::BuildHasher + std::default::Default> From<HashMapTextTextDbTy
...
+
+        let postgres_types::Kind::Domain(inner_type) = ty.kind() else {
+            return false;
+        };
+
+        let postgres_types::Kind::Array(inner_type) = inner_type.kind() else {
+            return false;
+        };
+
+        matches!(inner_type, &postgres_types::Type::PATH)
```
michaelmattig, line 1022: why not use a line?

services/src/datasets/postgres.rs
michaelmattig: you can use this trait: https://github.com/geo-engine/geoengine/blob/cfa23a912140f669287176a1ad78c7994e4a87a7/operators/src/engine/operator.rs#L389

services/src/contexts/schema.sql
ChristianBeilschmidt: You could wonder why not to use the `polygon` type.

services/src/api/model/db_types.rs
ChristianBeilschmidt: > Lines are represented by the linear equation Ax + By + C = 0, where A and B are not both zero. 

I thought we wouldn't want that.

services/src/api/model/db_types.rs
ChristianBeilschmidt: No, for `MultiLineString` I use and open path, but polygonal paths are closed.

services/src/api/model/datatypes.rs
ChristianBeilschmidt: 1. We need multiple line strings and not lines.

2. 
> Lines are represented by the linear equation Ax + By + C = 0, where A and B are not both zero.

I thought we wouldn't want that.

services/src/datasets/postgres.rs
ChristianBeilschmidt: But `OgrMetaData` is not an operator so I cannot use `OperatorName`, right?
`OperatorName` is implemented for `OgrSource`, which would yield `"OgrSource"`, but I need `"OgrMetaData"`.

services/src/contexts/postgres.rs
ChristianBeilschmidt: [8d4d2e5](https://github.com/geo-engine/geoengine/pull/851/commits/8d4d2e52f63a17281ca50f8bd5f577d011338b98)

I removed the types as they were custom PATH and POINT impls.

services/src/contexts/schema.sql
michaelmattig: i still don't get it. 

Do you mean we define our own polygon type instead of the built-in polygon type because we do not want to store the bbox?

How is it ensured that the path is actually closed?

ChristianBeilschmidt: [7b2f1cd](https://github.com/geo-engine/geoengine/pull/851/commits/7b2f1cd03f520a47b46dca63e9c7d7f3971bfc1d)

Our polygon's have closed paths in our rings so it corresponds 1:1.

PR #856, opened 2023-08-17
LineSymbology Serialization Bug

glombiewski:

LineSymbology was missing a camelCase such that auto_simplified could not be read properly.


glombiewski:
[LineString Daten können nicht angezeigt werden](https://trello.com/c/OGgd1lPS/1450-linestring-daten-k%C3%B6nnen-nicht-angezeigt-werden)

PR #857, opened 2023-08-17
remove feature gates

ChristianBeilschmidt:

---

<TEXT>


ChristianBeilschmidt:
[Update Backend](https://trello.com/c/U4DMtFcR/1420-update-backend)

.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           curl --proto '=https' --tlsv1.2 --retry 10 --retry-connrefused -fsSL "https://sh.rustup.rs" | sh -s -- --profile minimal --default-toolchain none -y
           echo "${CARGO_HOME:-$HOME/.cargo}/bin" >> $GITHUB_PATH
       - name: Install SQLFluff
-        run: pip install sqlfluff==2.1.4
+        run: pip install sqlfluff==2.3.0
```
michaelmattig, line 56: I can't comment at the correct line, but: do we still need --all-features in the CI build strategy? all-features is the same as pro now?

test_data/gbif/init_test_data.sql
```diff
@@ -0,0 +1,6 @@
+CREATE EXTENSION IF NOT EXISTS postgis WITH SCHEMA public;
+
+SET ROLE 'geoengine'; -- noqa: PRS
```
michaelmattig, line 3: what does the comment mean?

services/src/datasets/add_from_directory.rs
```diff
@@ -113,9 +113,7 @@ pub async fn add_providers_from_directory<D: LayerProviderDb>(
         add_from_dir(db, path).await;
     }
 
-    #[cfg(feature = "nfdi")]
```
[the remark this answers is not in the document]
michaelmattig: also: if I do cargo run, shouldn't the ABCD provider etc. be always present now? I don't see them when I run geo engine

test_data/gbif/init_test_data.sql
```diff
@@ -0,0 +1,6 @@
+CREATE EXTENSION IF NOT EXISTS postgis WITH SCHEMA public;
```
michaelmattig, line 1: why with schema public? we don't need it there

README.md
```diff
@@ -75,6 +75,21 @@ During development, you can use the following command to clean the database and
 sudo -u postgres psql -d geoengine -c "drop schema public cascade; create schema public authorization geoengine; create extension postgis;" && cargo run --features pro
 ```
 
+##### NFDI / GFBio / GBIF
+
+For running the NFDI / GFBio / GBIF data providers during development, you need to integrate the test data like this:
+
+```bash
+sudo -u postgres psql --dbname=geoengine -c \
```
michaelmattig, line 83: maybe make subsections for each provider? Or at least say which line is relevant for which provider

test_data/gbif/init_test_data.sql
ChristianBeilschmidt: sqlfluff does not recognize set role xyz, so this mutes the linter

test_data/gbif/init_test_data.sql
ChristianBeilschmidt: Maybe you have an idea, but when I install it into schema gbif and then try to install it again with schema abcd, it says it is already installed.
When I install it into public, it can be used anywhere.

services/src/datasets/add_from_directory.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: [8a21aa0](https://github.com/geo-engine/geoengine/pull/857/commits/8a21aa0ef77810d5366567a15a560a7c02d6216e)

It is in the listing, isn't it?

michaelmattig (approved):

PR #864, opened 2023-08-24
Compressed-vector-cache

jdroenner:

---

- Added compresseion to vector cache
- Uee a single Cache Stream implementation


jdroenner:
[Cache: Kompression für Vektor-Daten](https://trello.com/c/6unFkh41/1409-cache-kompression-f%C3%BCr-vektor-daten)

datatypes/src/collections/feature_collection.rs
```diff
@@ -88,6 +88,36 @@ impl<CollectionType> AsRef<FeatureCollection<CollectionType>>
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct FeatureCollectionInternals<G> {
```
michaelmattig, line 92: why do you not use this in the `FeatureCollection<CollectionType>`? How can we guarantee it's the same otherwise?

operators/src/pro/cache/cache_chunks.rs
```diff
@@ -391,31 +320,337 @@ impl_cache_result_check!(MultiPolygon);
...
+pub type CompressedDataCollection = CompressedFeatureCollection<NoGeometry>;
+pub type CompressedMultiPointCollection = CompressedFeatureCollection<MultiPoint>;
+pub type CompressedMultiLineStringCollection = CompressedFeatureCollection<MultiLineString>;
+pub type CompressedMultiPolygonCollection = CompressedFeatureCollection<MultiPolygon>;
+
+#[derive(Debug)]
+pub struct CompressedFeatureCollectionImpl<G, C> {
+    data: Vec<u8>,
+    spatial_bounds: Option<geoengine_datatypes::primitives::BoundingBox2D>,
+    time_interval: Option<geoengine_datatypes::primitives::TimeInterval>,
```
michaelmattig, line 363: why is this an option? Features always have an attached time. It can just be [begin of time, end of time). Does this require special handling?

operators/src/pro/cache/cache_chunks.rs
```diff
@@ -391,31 +320,337 @@ impl_cache_result_check!(MultiPolygon);
...
+pub type CompressedFeatureCollection<G> = CompressedFeatureCollectionImpl<G, Lz4FlexCompression>;
+pub type CompressedDataCollection = CompressedFeatureCollection<NoGeometry>;
+pub type CompressedMultiPointCollection = CompressedFeatureCollection<MultiPoint>;
+pub type CompressedMultiLineStringCollection = CompressedFeatureCollection<MultiLineString>;
+pub type CompressedMultiPolygonCollection = CompressedFeatureCollection<MultiPolygon>;
+
+#[derive(Debug)]
+pub struct CompressedFeatureCollectionImpl<G, C> {
+    data: Vec<u8>,
+    spatial_bounds: Option<geoengine_datatypes::primitives::BoundingBox2D>,
```
michaelmattig, line 362: I guess this is an Option because it is None for DataCollection? Didn't we say at some point, that Features with no geometry are just valid everywere? Maybe it would be simpler to take the CRS bounds then.

operators/src/pro/cache/cache_chunks.rs
```diff
@@ -186,11 +174,28 @@ where
         // So the query is not updated.
         Ok(())
     }
+
+    fn cache_element_hit(&self, query: &Self::Query) -> bool {
+        let temporal_hit = if let Some(time_bounds) = self.time_interval {
+            time_bounds == query.time_interval || time_bounds.intersects(&query.time_interval)
```
michaelmattig, line 180: shouldn't it be `contains`?

operators/src/pro/cache/cache_chunks.rs
```diff
@@ -186,11 +174,28 @@ where
...
+        } else {
+            false // TODO: should we return true here?
```
michaelmattig, line 182: I think we have to think about if it makes sense that time_bounds can be None. But if they can, I think it has to be false here. As we do not know the bounds, we cannot guarantee that all elements of the query are contained in the element.

operators/src/pro/cache/cache_chunks.rs
```diff
@@ -186,11 +174,28 @@ where
...
+        };
+
+        let spatial_hit = if let Some(spatial_bounds) = self.spatial_bounds {
+            spatial_bounds == query.spatial_bounds
+                || spatial_bounds.intersects_bbox(&query.spatial_bounds)
+        } else {
+            !G::IS_GEOMETRY // return true if the geometry is not spatial
```
michaelmattig, line 189: isn't the only case where spatial_bounds are None, if the collection is DataCollection? And DataCollection is the only one where IS_GEMOTRY is false right? So we could just say true here. Because in a DataCollection, every  feature is valid everywhere and thus contained in any element, regardless of the spatial bounds.

operators/src/pro/cache/cache_chunks.rs
```diff
@@ -391,31 +320,337 @@ impl_cache_result_check!(MultiPolygon);
...
+    where
+        Self: Sized;
+
+    fn array_to_bytes(table: StructArray) -> Result<Vec<u8>, CacheError> {
+        let record_batch = RecordBatch::from(&table);
+
+        let mut file_writer = FileWriter::try_new_with_options(
+            Vec::new(),
+            record_batch.schema().as_ref(),
+            IpcWriteOptions::default(), // TODO: Add ".try_with_compression(COMPRESSION)?," once issues are resolved
```
michaelmattig, line 438: maybe specify the "issues" for future generations

operators/src/pro/cache/cache_chunks.rs
jdroenner: No this is the check if a single element should be part of the query answer.

operators/src/pro/cache/cache_chunks.rs
jdroenner: [21fe3b6](https://github.com/geo-engine/geoengine/pull/864/commits/21fe3b6b20248c32d049f0099b08426b18ff1767)

operators/src/pro/cache/cache_stream.rs
```diff
@@ -0,0 +1,262 @@
...
+use super::shared_cache::{CacheBackendElementExt, CacheElement};
+use crate::util::Result;
+use futures::stream::FusedStream;
+use futures::{Future, Stream};
+use pin_project::pin_project;
+use std::{pin::Pin, sync::Arc};
+
+type DecompressorFutureType<X> = tokio::task::JoinHandle<std::result::Result<X, CacheError>>;
+
+/// Our own tile stream that "owns" the data (more precisely a reference to the data)
```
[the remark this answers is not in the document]
jdroenner: [6936933](https://github.com/geo-engine/geoengine/pull/864/commits/6936933b0e229e8ad0d9206efcac9fdb3a206b66)

[5443506](https://github.com/geo-engine/geoengine/pull/864/commits/5443506052d3c3c59a04b5faca883ab8314d8c37)

datatypes/src/collections/feature_collection.rs
jdroenner: this construct helps to transform the data object into a data struct and back. Don't know if that is a good pattern but it works.

operators/src/pro/cache/cache_chunks.rs
jdroenner: The option handles the case when the collection is empty

operators/src/pro/cache/cache_chunks.rs
jdroenner: it must be true because if it is the empty collection it might be the only empty in the cache and the cache must produce something.

operators/src/pro/cache/cache_chunks.rs
jdroenner: Besides the DataCollection it is empty if the collection is empty. In both cases it must return true. It is filtered in later step.

operators/src/pro/cache/cache_chunks.rs
jdroenner: [75cf335](https://github.com/geo-engine/geoengine/pull/864/commits/75cf335a877d4817e8b2116cbb12990a519fff5a)

operators/src/pro/cache/cache_chunks.rs
jdroenner: [75cf335](https://github.com/geo-engine/geoengine/pull/864/commits/75cf335a877d4817e8b2116cbb12990a519fff5a)

operators/src/pro/cache/cache_chunks.rs
jdroenner: [d8ac3d3](https://github.com/geo-engine/geoengine/pull/864/commits/d8ac3d359f68ad85ada90e3ef513af689250dbbe)

michaelmattig (approved):

PR #867, opened 2023-08-25
Use postgres port on startup

glombiewski:

---

- Added postgres port option to server startup.

### DIFF
--- a/services/src/pro/server.rs
+++ b/services/src/pro/server.rs
@@ -216,18 +216,9 @@ async fn start_postgres(
 ) -> Result<()> {
     {
         let db_config = config::get_config_element::<config::Postgres>()?;
-        let mut pg_config = bb8_postgres::tokio_postgres::Config::new();
-        pg_config
-            .user(&db_config.user)
-            .password(&db_config.password)
-            .host(&db_config.host)
-            .dbname(&db_config.database)
-            .port(db_config.port)
-            // fix schema by providing `search_path` option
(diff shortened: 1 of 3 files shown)


glombiewski:
I did not see any server startup tests, but if there are any, let me know and I'll add a test for this case.

glombiewski:
[Postgres Config Port wird nicht verwendet](https://trello.com/c/CFJQsBcG/1456-postgres-config-port-wird-nicht-verwendet)

services/src/pro/server.rs
```diff
@@ -222,6 +222,7 @@ async fn start_postgres(
             .password(&db_config.password)
             .host(&db_config.host)
             .dbname(&db_config.database)
+            .port(db_config.port)
```
michaelmattig, line 225: can you also add this to the non-pro version: https://github.com/glombiewski/geoengine/blob/814d06609161cbdc75cc96bc3a58f1112d068630/services/src/server.rs#L53

michaelmattig:
Startup tests are here; https://github.com/glombiewski/geoengine/blob/814d06609161cbdc75cc96bc3a58f1112d068630/services/tests/startup.rs#L49

services/src/pro/server.rs
glombiewski: 9cd7f7ac6e175ab4a0b9af46e88a8f798e7c8fc3

michaelmattig (approved):

PR #881, opened 2023-09-25
first version of database migration

michaelmattig:

---

<TEXT>


michaelmattig:
[Migrations-Funktionalität bauen](https://trello.com/c/rW1XOTBH/1389-migrations-funktionalit%C3%A4t-bauen)

services/src/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use super::database_migration::{DatabaseVersion, Migration};
+
+pub struct Migration0000Initial;
+
+#[async_trait]
+impl Migration for Migration0000Initial {
+    fn prev_version(&self) -> Option<DatabaseVersion> {
+        None
+    }
```
ChristianBeilschmidt, line 19: Why do we need this here?
What about the list of all migrations… we could have a discrepancy, there.

services/src/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,105 @@
...
+            ) VALUES (
+                $1,
+                'Layers',
+                'All available Geo Engine layers',
+                ARRAY[]::"PropertyType"[]
+            );"#,
+            )
+            .await?;
+
+        tx.execute(&stmt, &[&INTERNAL_LAYER_DB_ROOT_COLLECTION_ID])
```
ChristianBeilschmidt, line 65: This is critical. If we change the constant (and we will), this changes without being a new migration.

services/src/contexts/migrations/database_migration.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    let current_version = determine_current_database_version(conn)
+        .await?
+        .expect("after migration, there should be a current version.");
+
+    let latest_version = migrations
+        .last()
+        .expect("there should be at least one migration, namely the inital migration.")
+        .version();
```
ChristianBeilschmidt, line 126: Can't we do this here instead of just checking it here?

services/src/pro/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,122 @@
...
+                ($3, 'anonymous');"#,
+            )
+            .await?;
+
+        tx.execute(
+            &stmt,
+            &[
+                &Role::admin_role_id(),
+                &Role::registered_user_role_id(),
+                &Role::anonymous_role_id(),
```
ChristianBeilschmidt, line 39: Here, we also have the problem that this values can change and then we have a different state 0.

services/src/contexts/migrations/database_migration.rs
```diff
@@ -0,0 +1,300 @@
...
+
+        migration.migrate(&tx).await?;
+
+        let stmt = tx
+            .prepare("UPDATE geoengine SET database_version = $1")
+            .await?;
+
+        tx.execute(&stmt, &[&migration.version()]).await?;
+
+        tx.commit().await?;
```
ChristianBeilschmidt, line 114: Maybe, we could check at the end, that all workflows adhere to our JSON schema? We could determine faulty migrations this way.

ChristianBeilschmidt (changes requested):

services/src/contexts/migrations/migration_0000_initial.rs
michaelmattig: yeah, but that discrepancy would be caught. If we do not specify the previous version, we could forget one in the list of all migrations and it wouldn't get noticed.

services/src/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        tx.batch_execute(include_str!("migration_0000_initial.sql"))
+            .await?;
+
+        let stmt = tx
+            .prepare(
+                "
+            INSERT INTO geoengine (clear_database_on_start, database_version) VALUES ($1, '0000_initial');",
+            )
+            .await?;
```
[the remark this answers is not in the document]
michaelmattig: the `migrate_database` method checks this. I think it's better to do it at one place, because an individual migration could forget to check it. Now it is automatically checked.

The version is also only inserted in the initial migration. The migration logic in `migrate_database` updates the current version automatically after each migration.

services/src/contexts/migrations/migration_0000_initial.rs
michaelmattig: what would be the solution? Create a new constant in the migration?

services/src/contexts/migrations/database_migration.rs
```diff
@@ -0,0 +1,300 @@
...
+
+/// The logic for migrating the database from one version to another.
+#[async_trait]
+pub trait Migration: Send + Sync {
+    /// The previous version of the database. `None` if this is the first migration.
+    /// The database must be in this version in order for the migration to be applied.
+    fn prev_version(&self) -> Option<DatabaseVersion>;
+
+    /// The new version of the database after applying this migration.
+    fn version(&self) -> DatabaseVersion;
```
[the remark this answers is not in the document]
michaelmattig: sure, but that still doesn't mean that only because the version of a migration is larger then the version of another migration, that it can be applied. It could be that there are other versions in between two versions.

We could make version an integer to avoid this. But that removes a lot of flexiblity, like linking the version so the major release, or adding some comment that gives some context.

services/src/contexts/migrations/database_migration.rs
michaelmattig: do what exactly?

services/src/pro/contexts/migrations/migration_0000_initial.rs
michaelmattig: so should I add a new constant here and use that instead?

services/src/pro/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,122 @@
...
+                'admin',
+                $4,
+                true
+            );"#,
+            )
+            .await?;
+
+        let quota_available =
+            crate::util::config::get_config_element::<crate::pro::util::config::Quota>()?
+                .default_available_quota;
```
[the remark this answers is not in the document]
michaelmattig: that's true. But I think in the case that the config changes we need to adjust the migration here, to translate the future quota config to the legacy quota config in order to create the initial schema.

Or: we could also just give the admin unlimited quota.

services/src/contexts/migrations/database_migration.rs
michaelmattig: That's possible but could also be very costly. If you have thousands of workflows stored you have to load them all from the database and parse them.

I think it suffices to write tests for the migrations that actually change something in the operators and test exactly what they change to not break anything.

services/src/contexts/postgres.rs
```diff
@@ -194,91 +191,24 @@ where
...
-                &INTERNAL_LAYER_DB_ROOT_COLLECTION_ID,
-                &UNSORTED_COLLECTION_ID,
-            ],
-        )
-        .await?;
+        Ok(())
+    }
 
-        tx.commit().await?;
+    #[allow(clippy::too_many_lines)]
```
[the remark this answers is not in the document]
michaelmattig: 8165d13dc08322e227e5d0dee1fe0526e02fda60

services/src/contexts/migrations/database_migration.rs
ChristianBeilschmidt: 1. The necessity of loading it out of the database means you ignored the task hint for pg_jsonschema (https://github.com/supabase/pg_jsonschema). This way, you don't need to load it out to check it.

2. Could you point to where you checked this?

services/src/contexts/migrations/migration_0000_initial.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Could you point to the place where it automatically updates it?

services/src/contexts/migrations/migration_0000_initial.rs
ChristianBeilschmidt: I think we have to do this. You also could not say `UPDATE id = 1234 where ID = ?` and then add the constant since it could be changed and the update would not have an effect.

services/src/contexts/migrations/database_migration.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: Ok, but we shouldn't point to the migration directly? Like a backward-linked list? Then we don't need the array? What is your take on this?

services/src/contexts/migrations/database_migration.rs
ChristianBeilschmidt: update the version instead of determining that the version is correct

services/src/pro/contexts/migrations/migration_0000_initial.rs
ChristianBeilschmidt: Yeah, it must be locally, I guess

services/src/pro/contexts/migrations/migration_0000_initial.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: We need to somehow differentiate between migration and configuration.
My stomach says we shouldn't mix this. Moreover, the config can be changed after the first start.
How do we deal wit this?

services/src/contexts/migrations/migration_0000_initial.rs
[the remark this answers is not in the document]
michaelmattig: https://github.com/geo-engine/geoengine/pull/881/files#diff-5f7b6dfc9a5e3b93a8b875257f9f86c8c9e138aba9864ac9e166f9cc39e26e4bR108

services/src/pro/contexts/migrations/migration_0000_initial.rs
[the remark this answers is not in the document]
michaelmattig: after discussion with Johannes: make a constant for admin available quota and do not incorporate the config into the migration at all. For the future we try to not mix them.

services/src/contexts/migrations/database_migration.rs
michaelmattig: Johannes: write a todo and/or make new card

michaelmattig: I created a new card https://trello.com/c/SH9dewSd

services/src/contexts/migrations/migration_0000_initial.rs
michaelmattig: 045393894029fd54081e75acd1ba1c940f6bd3b8

services/src/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,105 @@
...
+            ) VALUES (
+                $1,
+                'Unsorted',
+                'Unsorted Layers',
+                ARRAY[]::"PropertyType"[]
+            );"#,
+            )
+            .await?;
+
+        tx.execute(&stmt, &[&UNSORTED_COLLECTION_ID]).await?;
```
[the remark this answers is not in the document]
michaelmattig: 045393894029fd54081e75acd1ba1c940f6bd3b8

services/src/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,105 @@
...
+            )
+            .await?;
+
+        tx.execute(
+            &stmt,
+            &[
+                &INTERNAL_LAYER_DB_ROOT_COLLECTION_ID,
+                &UNSORTED_COLLECTION_ID,
+            ],
+        )
```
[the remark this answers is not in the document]
michaelmattig: 045393894029fd54081e75acd1ba1c940f6bd3b8

services/src/pro/contexts/migrations/migration_0000_initial.rs
michaelmattig: 045393894029fd54081e75acd1ba1c940f6bd3b8

services/src/pro/contexts/migrations/migration_0000_initial.rs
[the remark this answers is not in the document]
michaelmattig: 045393894029fd54081e75acd1ba1c940f6bd3b8

services/src/pro/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,122 @@
...
+            .prepare(
+                r#"
+            INSERT INTO user_roles 
+                (user_id, role_id)
+            VALUES 
+                ($1, $1);"#,
+            )
+            .await?;
+
+        tx.execute(&stmt, &[&Role::admin_role_id()]).await?;
```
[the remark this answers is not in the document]
michaelmattig: 045393894029fd54081e75acd1ba1c940f6bd3b8

services/src/pro/contexts/migrations/migration_0000_initial.rs
```diff
@@ -0,0 +1,122 @@
...
+            .await?;
+
+        tx.execute(
+            &stmt,
+            &[
+                &Role::admin_role_id(),
+                &Role::registered_user_role_id(),
+                &Role::anonymous_role_id(),
+                &INTERNAL_LAYER_DB_ROOT_COLLECTION_ID,
+                &UNSORTED_COLLECTION_ID,
```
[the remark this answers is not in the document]
michaelmattig: 045393894029fd54081e75acd1ba1c940f6bd3b8

jdroenner (approved):

PR #890, opened 2023-10-30
use options in list datasets

michaelmattig:

---

<TEXT>

### DIFF
--- a/services/src/api/handlers/datasets.rs
+++ b/services/src/api/handlers/datasets.rs
@@ -1158,7 +1158,7 @@ mod tests {
             .unwrap();
 
         let ds = AddDataset {
-            name: Some(DatasetName::new(None, "My_Dataset2")),
+            name: Some(DatasetName::new(None, "My Dataset2")),
             display_name: "OgrDataset2".to_string(),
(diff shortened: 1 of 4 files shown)


michaelmattig:
[Die Options in den PostgresDbs werden teilweise ignoriert](https://trello.com/c/nJQiUIuZ/1493-die-options-in-den-postgresdbs-werden-teilweise-ignoriert)

services/src/api/handlers/datasets.rs
```diff
@@ -1226,7 +1226,7 @@ mod tests {
                 "symbology": null
             },{
                 "id": id2,
-                "name": "My_Dataset2",
+                "name": "My Dataset2",
```
jdroenner, line 1229: i think names are not allowed to contain " " (spaces). Only DisplayNames may contain them.

jdroenner (changes requested):

services/src/api/handlers/datasets.rs
michaelmattig: da37fa54d75953a358c9ff55cec07e4905b4b346

jdroenner (approved):

PR #908, opened 2023-12-20
Layer Search

ChristianBeilschmidt:

---

<TEXT>


ChristianBeilschmidt:
Note: This works for now only for the non-pro LayerDb and the GBIF provider.

services/src/api/handlers/layers.rs
```diff
@@ -40,6 +41,25 @@ where
...
             .service(
                 web::scope("/collections")
+                    .service(
+                        web::scope("/search")
+                            .route(
+                                r#"autocomplete/{provider}/{collection}"#,
+                                web::get().to(autocomplete_handler::<C>),
+                            )
+                            .route(
+                                r#"capabilities/{provider}"#,
```
michaelmattig, line 51: wouldn't it make more sense to have this under /layers/{provider}/capabilities?

services/src/api/handlers/layers.rs
```diff
@@ -40,6 +41,25 @@ where
...
+                                web::get().to(search_capabilities_handler::<C>),
+                            )
+                            .route(
+                                r#"{provider}/{collection}"#,
+                                web::get().to(search_handler::<C>),
+                            ),
+                    )
+                    .service(web::scope("/autocomplete").route(
```
michaelmattig, line 59: why are there two autocomplete routes?

services/src/api/handlers/layers.rs
```diff
@@ -314,6 +334,218 @@ async fn list_collection_handler<C: ApplicationContext>(
     Ok(web::Json(collection))
 }
 
+// Returns the search capabilities of the given provider
+#[utoipa::path(
+    tag = "Layers",
+    get,
+    path = "/layers/collections/search/capabilities/{provider}",
+    responses(
+        (status = 200, description = "OK", body = SearchCapabilities,
```
michaelmattig, line 343: I guess it would make sense to make the capabilities more generic, and e.g. include the ability to list collections and in the future more

services/src/api/handlers/layers.rs
```diff
@@ -314,6 +334,218 @@ async fn list_collection_handler<C: ApplicationContext>(
...
+async fn search_capabilities_handler<C: ApplicationContext>(
+    app_ctx: web::Data<C>,
+    path: web::Path<DataProviderId>,
+    session: C::Session,
+) -> Result<impl Responder> {
+    let provider = path.into_inner();
+
+    if provider == ROOT_PROVIDER_ID {
+        return Err(NotImplemented {
+            message: "Global search is not supported".to_string(),
```
michaelmattig, line 370: wouldn't users expect this to work?

services/src/layers/listing.rs
```diff
@@ -51,4 +136,40 @@ pub trait DatasetLayerCollectionProvider {
...
+                prefix: false,
+            },
+            autocomplete: false,
+            filters: None,
+        })
+    }
+
+    /// Perform a search
+    #[allow(unused_variables)]
+    async fn search(
```
michaelmattig, line 154: should this be in a separate trait, if it is optional to implement it?

services/src/contexts/migrations/migration_0002_gbif_config.rs
```diff
@@ -0,0 +1,110 @@
...
+                ADD ATTRIBUTE autocomplete_timeout int
+            "#,
+        )
+        .await?;
+
+        // as ALTER TYPE ADD ATTRIBUTE does not support default values we manually have to set the default value of the new field in the layer_providers table
+        tx.batch_execute(
+            r#"
+                UPDATE layer_providers
+                SET definition.gbif_data_provider_definition.autocomplete_timeout = 3
```
michaelmattig, line 34: is 3 a good default?

services/src/contexts/migrations/migration_0002_gbif_config.rs
```diff
@@ -0,0 +1,110 @@
...
+        conn.batch_execute(&test_data_sql).await?;
+
+        // perform the previous migration
+        migrate_database(&mut conn, &[Box::new(Migration0001RasterStacks)]).await?;
+
+        // perform the current migration
+        migrate_database(&mut conn, &[Box::new(Migration0002GbifConfig)]).await?;
+
+        // verify that the default value for autocomplete_timeout has been set correctly
+        let autocomplete_timeout = conn
```
michaelmattig, line 85: this should use load_dataprovider for testing instead, because that is the one that needs to work after migration

services/src/datasets/external/gbif.rs
```diff
@@ -174,7 +207,77 @@ impl GbifDataProvider {
...
+                    },
+                    name: canonicalname.clone(),
+                    description: format!("{num_points} occurrences"),
+                    properties: vec![],
+                })
+            })
+            .collect::<Vec<_>>())
+    }
+
+    async fn get_datasets_search_items(
```
michaelmattig, line 233: why is this function never used?

services/src/datasets/external/gbif.rs
```diff
@@ -197,6 +300,217 @@ impl GbifDataProvider {
             .collect::<Vec<_>>())
     }
 
+    async fn get_datasets_autocomplete_items(
```
michaelmattig, line 303: why is this function never used?

services/src/datasets/external/gbif.rs
```diff
@@ -84,14 +91,21 @@ impl GbifDataProvider {
...
+        autocomplete_timeout: i32,
+    ) -> Result<Self> {
         let pg_mgr = PostgresConnectionManager::new(db_config.pg_config(), NoTls);
         let pool = Pool::builder().build(pg_mgr).await?;
 
         Ok(Self {
             db_config,
             pool,
             cache_ttl,
+            autocomplete_timeout: u64::try_from(autocomplete_timeout).unwrap_or(u64::MAX),
```
michaelmattig, line 108: is u64::MAX a good fallback?

services/src/datasets/external/gbif.rs
```diff
@@ -197,6 +300,217 @@ impl GbifDataProvider {
...
+        search_string: &str,
+        taxonranks: Vec<String>,
+    ) -> Result<Vec<String>> {
+        let path = path.split_once('/').map_or_else(|| "", |(_, path)| path);
+        let filters = GbifDataProvider::get_filters(path);
+        let taxonrank_filter = Self::create_taxonrank_filter(&taxonranks);
+        let conn = self.pool.get().await?;
+        let query = &format!(
+            r#"
+            SELECT DISTINCT canonicalname
```
michaelmattig, line 475: why is the distinct needed here?

services/src/datasets/external/gbif.rs
```diff
@@ -256,6 +570,103 @@ impl GbifDataProvider {
...
+    ) -> Result<Vec<String>> {
+        let column = GbifDataProvider::level_name(path);
+        if !Self::LEVELS.contains(&column.as_str()) {
+            return Err(Error::InvalidLayerCollectionId);
+        }
+        let filters = GbifDataProvider::get_filters(path);
+        let conn = self.pool.get().await?;
+        let query = &format!(
+            r#"
+            SELECT DISTINCT "{column}"
```
michaelmattig, line 637: why is the distinct needed here?

services/src/datasets/external/gbif.rs
```diff
@@ -384,10 +795,154 @@ impl LayerCollectionProvider for GbifDataProvider {
...
+            properties: vec![],
+        })
+    }
+
+    async fn autocomplete_search(
+        &self,
+        collection_id: &LayerCollectionId,
+        search: SearchParameters,
+    ) -> Result<Vec<String>> {
+        let selector = collection_id.0.split_once('/').map_or_else(
```
michaelmattig, line 884: so a search term cannot contain "/"? Shouldn't we escape that then?

services/src/layers/postgres_layer_db.rs
```diff
@@ -617,6 +619,216 @@ where
...
+                found: collection_id.0.clone(),
+            }
+        })?;
+
+        let conn = self.conn_pool.get().await?;
+
+        let stmt = conn
+            .prepare(
+                "
+        SELECT DISTINCT name, description, properties
```
michaelmattig, line 650: why is the distinct needed here? id is the primary key right?

services/src/layers/postgres_layer_db.rs
```diff
@@ -617,6 +619,216 @@ where
...
+
+        let row = conn.query_one(&stmt, &[&collection]).await?;
+
+        let name: String = row.get(0);
+        let description: String = row.get(1);
+        let properties: Vec<Property> = row.get(2);
+
+        let pattern = match search.search_type {
+            SearchType::Fulltext => {
+                format!("%{}%", search.search_string)
```
michaelmattig, line 664: should we add (trigram) indexes to the name column?

services/src/layers/postgres_layer_db.rs
```diff
@@ -617,6 +619,216 @@ where
...
+            }
+            SearchType::Prefix => {
+                format!("{}%", search.search_string)
+            }
+        };
+
+        let stmt = conn
+            .prepare(
+                "
+        WITH RECURSIVE parents AS (
```
michaelmattig, line 787: re-use the query?

services/src/api/handlers/layers.rs
```diff
@@ -40,6 +41,25 @@ where
         web::scope("/layers")
             .service(
                 web::scope("/collections")
+                    .service(
```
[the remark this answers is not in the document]
ChristianBeilschmidt: The idea is that we always search within collections.

services/src/api/handlers/layers.rs
ChristianBeilschmidt: I guess a leftover

services/src/api/handlers/layers.rs
ChristianBeilschmidt: :+1:

services/src/api/handlers/layers.rs
ChristianBeilschmidt: We should think of this separately. How would you search if you don't know the capabilities of the providers? Search only in the ones that have a search? Which type to provide?

services/src/layers/listing.rs
ChristianBeilschmidt: We need to get at least an error from each provider.
How should the method know if some of the providers implement a trait? Dyn-Downcasting by using the capabilities as reference?

services/src/contexts/migrations/migration_0002_gbif_config.rs
```diff
@@ -0,0 +1,110 @@
...
+            "#,
+        )
+        .await?;
+
+        // as ALTER TYPE ADD ATTRIBUTE does not support default values we manually have to set the default value of the new field in the layer_providers table
+        tx.batch_execute(
+            r#"
+                UPDATE layer_providers
+                SET definition.gbif_data_provider_definition.autocomplete_timeout = 3
+                WHERE NOT ((definition).gbif_data_provider_definition IS NULL)
```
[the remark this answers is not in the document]
dbrandenstein: It also works the way it is right now. It's kind of confusing with composite types.
When you add a new field, it is set to NULL by default (and you can't provide another default). And once ANY field of a composite is NULL, the entire composite is no longer considered NOT NULL, as you say, so you can't check `(definition).gbif_data_provider_definition IS NOT NULL`, it will always be false once any field is NULL. However, once ANY field of a composite is NOT NULL, the composite is not NULL either, so you can use the syntax I used to check whether any field of the composite is NOT NULL.

test_data/provider_defs/gbif.json
```diff
@@ -9,5 +9,6 @@
     "database": "geoengine",
     "schema": "gbif"
   },
-  "cacheTtl": 0
+  "cacheTtl": 0,
+  "autocompleteTimeout": 5
```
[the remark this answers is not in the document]
dbrandenstein: No reason. I can set everything to 3, 5 or whatever you prefer. Good idea to have it the same everywhere.

services/src/layers/postgres_layer_db.rs
```diff
@@ -617,6 +619,216 @@ where
...
+        FROM (
+            SELECT 
+                concat(id, '') AS id, 
+                name, 
+                description, 
+                properties, 
+                FALSE AS is_layer
+            FROM layer_collections
+                JOIN (SELECT DISTINCT child FROM collection_children JOIN parents ON (id = parent)) cc ON (id = cc.child)
+            WHERE name ILIKE $4
```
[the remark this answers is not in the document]
dbrandenstein: Left for future extension of the feature

services/src/layers/postgres_layer_db.rs
dbrandenstein: I mindlessly copied it from `load_layer_collection`. IMO it can be removed in both instances

services/src/datasets/external/gbif.rs
dbrandenstein: Removed in https://github.com/geo-engine/geoengine/pull/908/commits/e23802633dc6b4ae701c3913186079f906872523 since https://github.com/geo-engine/geoengine/pull/908/commits/37f6862cc45f1941c5189369051b38a83c79c72b made it obsolete

services/src/datasets/external/gbif.rs
dbrandenstein: Removed in https://github.com/geo-engine/geoengine/commit/e23802633dc6b4ae701c3913186079f906872523 since https://github.com/geo-engine/geoengine/commit/37f6862cc45f1941c5189369051b38a83c79c72b made it obsolete

services/src/datasets/external/gbif.rs
dbrandenstein: Aren't the query parameters already stripped from the path at this point by the handler?

services/src/datasets/external/gbif.rs
dbrandenstein: There can be duplicates otherwise, for autocomplete they are unwanted

services/src/datasets/external/gbif.rs
dbrandenstein: There can be duplicates otherwise. Since we're selecting one item as filter duplicate entries are not wanted

services/src/layers/postgres_layer_db.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/908/commits/5b36378fe3c8ee1e601a86007dc0ea58fb56df75

services/src/api/handlers/layers.rs
ChristianBeilschmidt: [7997979](https://github.com/geo-engine/geoengine/pull/908/commits/7997979ccbc1b79232fd8c8e0f4d3d49be430e83)

services/src/api/handlers/layers.rs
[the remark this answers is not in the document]
ChristianBeilschmidt: [e336e65](https://github.com/geo-engine/geoengine/pull/908/commits/e336e6508a1ab96cc4620a5e91c8a2731dbaffe6)

services/src/api/handlers/layers.rs
ChristianBeilschmidt: [e336e65](https://github.com/geo-engine/geoengine/pull/908/commits/e336e6508a1ab96cc4620a5e91c8a2731dbaffe6)

michaelmattig: the search capabilites for the "Data catalogue" (previously "Layer Collections") now returns no search capabilites and thus can no longer searched in the UI.

services/src/contexts/migrations/migration_0002_gbif_config.rs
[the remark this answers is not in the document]
michaelmattig: but won't this match all other data provider definitions that are not gbif data providers as well?

services/src/datasets/external/gbif.rs
michaelmattig: yes, but why do you split at "/" here?

test_data/provider_defs/gbif.json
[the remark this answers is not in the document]
michaelmattig: I vote for 5 seconds

michaelmattig:
> Note: This works for now only for the non-pro LayerDb and the GBIF provider.

doesn't the VAT run on Pro though?

services/src/contexts/migrations/migration_0002_gbif_config.rs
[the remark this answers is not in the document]
dbrandenstein: They have all fields of the `gbif_data_provider_definition` set to NULL, so the whole GBIF definition composite is considered to be NULL

services/src/api/handlers/layers.rs
ChristianBeilschmidt:
```Rust
#[async_trait]
impl<Tls> LayerCollectionProvider for PostgresDb<Tls>
where
    Tls: MakeTlsConnect<Socket> + Clone + Send + Sync + 'static,
    <Tls as MakeTlsConnect<Socket>>::Stream: Send + Sync,
    <Tls as MakeTlsConnect<Socket>>::TlsConnect: Send,
{
    fn capabilities(&self) -> ProviderCapabilities {
        ProviderCapabilities {
            listing: true,
            search: SearchCapabilities {
                search_types: SearchTypes {
                    fulltext: true,
                    prefix: true,
                },
                autocomplete: true,
                filters: None,
            },
        }
    }
```

This is not true. It returns search capabilities.
If you mean the layerDb in `pro`, this will be implemented in another PR and also did not work before.

services/src/datasets/external/gbif.rs
ChristianBeilschmidt: Isn't it more or less "no timeout"?

services/src/datasets/external/gbif.rs
dbrandenstein: Apparently canonicalnames are potentially used for different "taxonomic paths" so there can be duplicates when you don't apply filters along the whole path leading to the taxonrank you're listing

services/src/datasets/external/gbif.rs
dbrandenstein: It's the way the collection-/layer-IDs are structured for the GBIF provider. In this instance, the selector datasets/filter/select determines whether you want to search within the family/genus/species datasets list, the options for applying a filter along the taxonomic "path" or perform a search across family, genus and species datasets

test_data/provider_defs/gbif.json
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/908/commits/d0386556a00f41398a5f5bcb45451542c2d09043

services/src/layers/postgres_layer_db.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/908/commits/ebe02f98f6fa21ab75fa0a1ece78b562a8614a88

services/src/api/handlers/layers.rs
ChristianBeilschmidt: [54e9d61](https://github.com/geo-engine/geoengine/pull/908/commits/54e9d610d94c8e69131f7b613668c5da39543b81)

michaelmattig (approved):

PR #909, opened 2023-12-20
dataset layer listing with tags

jdroenner:

---

<TEXT>


jdroenner:
image
image

jdroenner:

services/src/api/handlers/layers.rs
```diff
@@ -147,13 +147,13 @@ async fn get_layer_providers<C: ApplicationContext>(
...
+                provider_id: crate::layers::storage::INTERNAL_PROVIDER_ID,
                 collection_id: LayerCollectionId(
-                    crate::datasets::storage::DATASET_DB_ROOT_COLLECTION_ID.to_string(),
+                    crate::layers::storage::INTERNAL_LAYER_DB_ROOT_COLLECTION_ID.to_string(),
                 ),
             },
-            name: "Datasets".to_string(),
-            description: "Basic Layers for all Datasets".to_string(),
+            name: "Layers".to_string(),
+            description: "All available Geo Engine layers".to_string(),
```
michaelmattig, line 156: those are no longer "All layers" if the uploads are not contained

services/src/api/handlers/layers.rs
```diff
@@ -162,13 +162,11 @@ async fn get_layer_providers<C: ApplicationContext>(
...
-                collection_id: LayerCollectionId(
-                    crate::layers::storage::INTERNAL_LAYER_DB_ROOT_COLLECTION_ID.to_string(),
-                ),
+                provider_id: crate::datasets::storage::DATASET_DB_LAYER_PROVIDER_ID,
+                collection_id: LayerCollectionId("upload".to_string()),
             },
-            name: "Layers".to_string(),
-            description: "All available Geo Engine layers".to_string(),
+            name: "User Uploads".to_string(),
+            description: "User uploaded datasets".to_string(),
```
michaelmattig, line 169: What about the user created datasets via "saveWorkflowAsDataset"? They are no longer accesible.

services/src/datasets/postgres.rs
```diff
@@ -585,7 +590,8 @@ where
                 SELECT 
                     concat(id, '') AS id, 
                     display_name, 
-                    description
+                    description,
+                    tags
```
michaelmattig, line 594: filter for tags?

services/src/datasets/storage.rs
```diff
@@ -99,6 +104,17 @@ fn validate_main_file(main_file: &String) -> Result<(), ValidationError> {
     Ok(())
 }
 
+fn validate_tags(tags: &Vec<String>) -> Result<(), ValidationError> {
+    for tag in tags {
+        if tag.is_empty() || tag.contains('/') || tag.contains("..") || tag.contains(' ') {
```
michaelmattig, line 109: should the "uploads" tag be reserved? should upper/lower case play a role?

services/src/api/handlers/datasets.rs
```diff
@@ -378,6 +378,8 @@ pub async fn create_upload_dataset<C: ApplicationContext>(
     let db = app_ctx.session_context(session).db();
     let upload = db.load_upload(upload_id).await.context(UploadNotFound)?;
```
michaelmattig, line 380: comment unrelated to this line:

I think this approach in general is fine. To make Barbara even more happy I think it would be nice if one could navigate back in the UI by means of the back button next to the breadcrumbs instead of the back button on the top right

services/src/pro/datasets/postgres.rs
```diff
@@ -680,23 +691,70 @@ where
     ) -> Result<LayerCollection> {
         let conn = self.conn_pool.get().await?;
 
+        let coll_id = &collection.0;
+
+        if coll_id == "" || coll_id == &DATASET_DB_ROOT_COLLECTION_ID.to_string() {
```
michaelmattig, line 696: we would have to make this root collection accessible in the UI somehow

services/src/api/handlers/datasets.rs
jdroenner: i guess thats a new task?

services/src/datasets/storage.rs
jdroenner: good question. currently i check if upload is set and then add it if not (only for uploads).

jdroenner:
I moved the Data Catalog to the top. The "User Data" is the second entry.
image
In the "User Data" the entries are based on the config:
image
The config is currently in the code: https://github.com/geo-engine/geoengine/blob/6c48754a191c807c39b4b97cb4439cf5e2c14c28/services/src/datasets/dataset_listing_provider.rs#L59

services/src/api/handlers/layers.rs
jdroenner: i changed that as you can see in the screenshot.

services/src/api/handlers/layers.rs
jdroenner: You can also list the datasets with that tag. I thought the task was to remove everything that is not upload related. You can see how it works now in the screenshot.

services/src/datasets/postgres.rs
jdroenner: the non-pro path is now also handled

services/src/pro/datasets/postgres.rs
jdroenner: yes the root collection is now the "User Data". However there might be persons who will find that strange...

services/src/api/handlers/layers.rs
```diff
@@ -289,8 +283,9 @@ async fn list_collection_handler<C: ApplicationContext>(
 
     let db = app_ctx.session_context(session).db();
 
-    if provider == crate::datasets::storage::DATASET_DB_LAYER_PROVIDER_ID.into() {
-        let collection = db
+    if provider == crate::datasets::dataset_listing_provider::DATASET_LISTING_PROVIDER_ID.into() {
```
michaelmattig, line 286: suggestion by Chris: move special case(s) into load_layer_provider instead

PR #913, opened 2024-01-02
Update-2024-01-01

ChristianBeilschmidt:

---

<TEXT>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
           curl --proto '=https' --tlsv1.2 --retry 10 --retry-connrefused -fsSL "https://sh.rustup.rs" | sh -s -- --profile minimal --default-toolchain none -y
           echo "${CARGO_HOME:-$HOME/.cargo}/bin" >> $GITHUB_PATH
       - name: Install SQLFluff
-        run: pip install sqlfluff==2.3.0
+        run: pip install sqlfluff==2.3.5
       - name: Init rustup toolchain
         # somehow rustup show will take care to initialize the version based on the toolchain file
         run: rustup show
--- a/Settings-default.toml
+++ b/Settings-default.toml
@@ -152,7 +152,7 @@ admin_password = "adminadmin"   # Initial password for the user "admin", which i
 
 [quota]
 mode = "disabled" # "disabled", "track", "check"
-# default = 9999
+# initial_credits = 9999
 increment_quota_buffer_size = 100           # number of quota updates to buffer before sending them to the database
 increment_quota_buffer_timeout_seconds = 60 # number of seconds after which the quota updates are sent to the database
 
--- a/Settings-default.toml
+++ b/Settings-default.toml
@@ -146,13 +146,13 @@ anonymous_access = true
 
 # Settings for Geo Engine Pro
 [user]
-user_registration = true
+registration = true
 admin_email = "admin@localhost"
 admin_password = "adminadmin"   # Initial password for the user "admin", which is created on startup
 
 [quota]
 mode = "disabled" # "disabled", "track", "check"
-# default_available_quota = 9999
+# default = 9999
 increment_quota_buffer_size = 100           # number of quota updates to buffer before sending them to the database
 increment_quota_buffer_timeout_seconds = 60 # number of seconds after which the quota updates are sent to the database
 
@@ -183,6 +183,6 @@ volumes = { test_data = "test_data/" }
 [cache]
 enabled = false
 # storage limit for the cache
-cache_size_in_mb = 1_000 # 1 GB
+size_in_mb = 1_000 # 1 GB
 # storage limit for collecting query results before insertion into the cache in 
 landing_zone_ratio = 0.1 # 10% of total cache size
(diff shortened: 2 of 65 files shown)


ChristianBeilschmidt:
[Update Backend](https://trello.com/c/PhtBF8Ge/1464-update-backend)

Settings-default.toml
```diff
@@ -146,13 +146,13 @@ anonymous_access = true
 
 # Settings for Geo Engine Pro
 [user]
-user_registration = true
+registration = true
```
michaelmattig, line 149: if you change stuff in here, you have to update the files in the server-deployment repo as well

Settings-default.toml
```diff
@@ -146,13 +146,13 @@ anonymous_access = true
...
 admin_email = "admin@localhost"
 admin_password = "adminadmin"   # Initial password for the user "admin", which is created on startup
 
 [quota]
 mode = "disabled" # "disabled", "track", "check"
-# default_available_quota = 9999
+# default = 9999
```
michaelmattig, line 155: I think it was more clear previously. It's not clear what "default" is meant here (available quota)

datatypes/src/collections/data_collection.rs
```diff
@@ -16,6 +16,7 @@ impl<'i> IntoGeometryOptionsIterator<'i> for DataCollection {
     }
 }
 
+#[allow(clippy::into_iter_without_iter)] // we provide `.geometries()` instead
```
michaelmattig, line 19: why can we not use Iter there?

services/src/pro/users/oidc.rs
```diff
@@ -122,6 +122,7 @@ pub enum OidcError {
     IllegalProviderConfig {
         source: ParseError,
     },
+    #[snafu(display("ProviderDiscoveryError: {}", source))]
```
michaelmattig, line 125: why is this enum still ending in "Error" when you removed it in the other cases?

.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           curl --proto '=https' --tlsv1.2 --retry 10 --retry-connrefused -fsSL "https://sh.rustup.rs" | sh -s -- --profile minimal --default-toolchain none -y
           echo "${CARGO_HOME:-$HOME/.cargo}/bin" >> $GITHUB_PATH
       - name: Install SQLFluff
-        run: pip install sqlfluff==2.3.0
+        run: pip install sqlfluff==2.3.5
```
michaelmattig, line 56: don't know where else to put it: Latest rust stable is now 1.75, do you want to update it, because the card has a success criterion to be latest stable?

datatypes/src/collections/data_collection.rs
ChristianBeilschmidt: There are multiple ways to iterate over a collection. This, `.geometries()`, means iterating over geometries only. But there are ways to iterate over other columns with `.data()`, for instance, or to get time instances. Thus, it does not make sense to have `.iter()` point to exactly this iterator.

services/src/pro/users/oidc.rs
ChristianBeilschmidt: The enum is not, this is just the message.

.github/workflows/ci.yml
ChristianBeilschmidt: Sure, but I did already update it: [27b766f](https://github.com/geo-engine/geoengine/pull/913/commits/27b766f134b6798d5f524722626efa2fa8190839)

Settings-default.toml
ChristianBeilschmidt: Nightly does not use it and other instances don't use this setting afaik.

Settings-default.toml
ChristianBeilschmidt: [dcfbdd1](https://github.com/geo-engine/geoengine/pull/913/commits/dcfbdd104ab32abd61d215b54ca2a848d4eb4720)

datatypes/src/collections/data_collection.rs
michaelmattig: can we make `.iter()` generic over the return type?

ChristianBeilschmidt: I have two thoughts on this:

1. I don't think so, at least not so easily. We would need to put the iter-Impl into a trait that each of the return types must implement. So the type would know how the FeatureCollection would iterate over itself. Should be possible, but It could get blind doing this.
2. This would not be ergonomic at all. The caller would need to very specific typing instead of just calling .iter.
3. I doubt that it would make the linter happy. Can it recognize, that there is somewhat an `.iter()` method for this? I guess `.geometries()` is better suited for this, as it is now.

We could remove IntoIter and just provide `.geoemetries`, but I don't see the huge advantage.

Settings-default.toml
ChristianBeilschmidt: Actually, I don't know. Maybe start creating a changelog (again)?.
I could open up a PR for these changes as a draft, but we both know that we would forget this PR.

michaelmattig (approved):

PR #916, opened 2024-01-11
Dataset Layer Listing as DataProvider

michaelmattig:

---

<TEXT>

### DIFF
--- a/services/src/contexts/migrations/migration_0002_dataset_listing_provider.rs
+++ b/services/src/contexts/migrations/migration_0002_dataset_listing_provider.rs
@@ -1,0 +1,45 @@
+use async_trait::async_trait;
+use tokio_postgres::Transaction;
+
+use crate::error::Result;
+
+use super::database_migration::{DatabaseVersion, Migration};
+
+/// This migration adds the dataset layer listing provider
+pub struct Migration0002DatasetListingProvider;
+
+#[async_trait]
+impl Migration for Migration0002DatasetListingProvider {
+    fn prev_version(&self) -> Option<DatabaseVersion> {
+        Some("0001_raster_stacks".into())
+    }
+
+    fn version(&self) -> DatabaseVersion {
+        "0002_dataset_listing_provider".into()
+    }
+
+    async fn migrate(&self, tx: &Transaction<'_>) -> Result<()> {
+        tx.batch_execute(
+            r#"   
+            CREATE TYPE "DatasetLayerListingCollection" AS (
+                "name" text,
+                description text,
+                tags text []
+            );
+
+            CREATE TYPE "DatasetLayerListingProviderDefinition" AS (
+                id uuid,
+                "name" text,
+                description text,
+                collections "DatasetLayerListingCollection" []
+            );
+
+            ALTER TYPE "DataProviderDefinition" 
+            ADD ATTRIBUTE dataset_layer_listing_provider_definition "DatasetLayerListingProviderDefinition";
+        "#,
+        )
+        .await?;
+
+        Ok(())
+    }
+}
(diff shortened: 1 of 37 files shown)


michaelmattig:

services/src/pro/layers/postgres_layer_db.rs
```diff
@@ -480,10 +480,15 @@ where
...
+        let id = DataProviderDefinition::<Self>::id(&provider);
         conn.execute(
             &stmt,
-            &[&id, &provider.type_name(), &provider.name(), &provider],
+            &[
+                &id,
+                &DataProviderDefinition::<Self>::type_name(&provider),
+                &DataProviderDefinition::<Self>::name(&provider),
+                &provider,
+            ],
```
ChristianBeilschmidt, line 491: Why can't we easily access the property functions anymore?

services/src/layers/external.rs
```diff
@@ -22,11 +24,9 @@ use geoengine_operators::source::{GdalLoadingInfo, OgrSourceDataset};
 use serde::{Deserialize, Serialize};
 
 #[async_trait]
-pub trait DataProviderDefinition:
-    CloneableDataProviderDefinition + Send + Sync + std::fmt::Debug
-{
+pub trait DataProviderDefinition<D: GeoEngineDb>: Send + Sync + std::fmt::Debug {
     /// create the actual provider for data listing and access
-    async fn initialize(self: Box<Self>) -> Result<Box<dyn DataProvider>>;
+    async fn initialize(self: Box<Self>, db: D) -> Result<Box<dyn DataProvider>>;
```
ChristianBeilschmidt, line 29: Why do we need access to a `GeoEngineDb` for initializing?

ChristianBeilschmidt (changes requested):

services/src/contexts/migrations/migration_0002_dataset_listing_provider.rs
```diff
@@ -0,0 +1,45 @@
...
+                collections "DatasetLayerListingCollection" []
+            );
+
+            ALTER TYPE "DataProviderDefinition" 
+            ADD ATTRIBUTE dataset_layer_listing_provider_definition "DatasetLayerListingProviderDefinition";
+        "#,
+        )
+        .await?;
+
+        Ok(())
```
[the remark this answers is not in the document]
michaelmattig: I don't think we necessarily need to have exactly one Migration per version. E.g. Nextcloud allows multiple migrations https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/migrations.html

For me, it makes sense to have semantic migrations, rather than a single migration that does a whole bunch of unrelated stuff.

services/src/layers/external.rs
michaelmattig: The `DatasetLayerListingProvider` needs access to the Datasets. That's why I had to change the `DataProvider` trait. Furthermore, we always wanted to have credentials per user and data provider. To store and access those the `DataProvider` also needs access to the `GeoEngineDb`. That's why I think this change makes sense.

services/src/pro/layers/postgres_layer_db.rs
ChristianBeilschmidt: If I understand correctly, `TypedDataProviderDefinition` now implements `DataProviderDefinition`.
`DataProviderDefinition` seems to already have the methods `type_name`, `name`, etc.
However, then you implement these methods again in `TypedDataProviderDefinition` and the compiler now does not know which method to call.

Couldn't you just remove the impls in `TypedDataProviderDefinition`?

ChristianBeilschmidt:
Please merge the current master in order to merge it.

services/src/pro/layers/postgres_layer_db.rs
michaelmattig: the problem is that there are now two `DataProviderDefinition` traits, and you have to tell the compiler which one you want. If I remove the impls from `TypedDataProviderDefinition` then I have to do the matching in both free and pro DB. And then the compiler still doesn't know the type

image

PR #917, opened 2024-01-15
Implement multibands for operators

michaelmattig:

---

<TEXT>

### DIFF
--- a/operators/src/processing/meteosat/temperature.rs
+++ b/operators/src/processing/meteosat/temperature.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use crate::adapters::stack_individual_raster_bands;
 use crate::engine::{
     CanonicOperatorName, ExecutionContext, InitializedRasterOperator, InitializedSources, Operator,
     OperatorName, QueryContext, QueryProcessor, RasterBandDescriptor, RasterBandDescriptors,
@@ -10,7 +11,6 @@ use crate::util::Result;
 use async_trait::async_trait;
 use rayon::ThreadPool;
 
-use snafu::ensure;
 use TypedRasterQueryProcessor::F32 as QueryProcessorOut;
 
 use crate::error::Error;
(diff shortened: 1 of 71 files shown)


michaelmattig:
[[Raster-Stacking-Placeholder]](https://trello.com/c/jHuUp2Oc/1566-raster-stacking-placeholder)

operators/src/processing/temporal_raster_aggregation/temporal_aggregation_operator.rs
```diff
@@ -257,14 +257,18 @@ where
     #[allow(clippy::too_many_lines)]
     fn create_subquery_adapter_stream_for_single_band<'a>(
         &'a self,
-        query_rect_to_answer: RasterQueryRectangle,
-        band_idx: u32,
+        query: RasterQueryRectangle,
```
jdroenner, line 260: cant we use that partial query here?

operators/src/processing/meteosat/temperature.rs
```diff
@@ -302,9 +301,13 @@ where
         query: RasterQueryRectangle,
         ctx: &'a dyn QueryContext,
     ) -> Result<BoxStream<'a, Result<Self::Output>>> {
-        let src = self.source.query(query, ctx).await?;
-        let rs = src.and_then(move |tile| self.process_tile_async(tile, ctx.thread_pool().clone()));
-        Ok(rs.boxed())
+        stack_individual_aligned_raster_bands(&query, ctx, |query, ctx| async move {
```
jdroenner, line 304: why is this needed for this operator but not for the other MSG Operators?

michaelmattig: I initially thought it was needed for all msg operators, but as they just map tiles and don't care about the band, I removed it again. I only forgot it here: 02d77a9e4c6b54b8a0ee8a272c89873799792d19

operators/src/processing/temporal_raster_aggregation/temporal_aggregation_operator.rs
michaelmattig: do you mean `PartialQueryRectangle`? If so, then no, because it doesn't include the band but the subquery still needs to know it.

jdroenner (approved):

PR #920, opened 2024-01-19
Data-provider-desc-sort-prio

jdroenner:

---

<TEXT>

### DIFF
--- a/services/src/api/handlers/layers.rs
+++ b/services/src/api/handlers/layers.rs
@@ -220,7 +220,7 @@ async fn get_layer_providers<C: ApplicationContext>(
                 provider_id: provider_listing.id,
                 collection_id,
             },
-            name: provider_listing.name,
+            name: provider.name().to_owned(),
             description: provider.description().to_owned(),
             properties: Default::default(),
         }));
--- a/services/src/api/handlers/layers.rs
+++ b/services/src/api/handlers/layers.rs
@@ -187,6 +187,10 @@ async fn get_layer_providers<C: ApplicationContext>(
         })
         .await?
     {
+        if provider_listing.priority == i16::MIN {
+            continue; // skip providers that are disabled
+        };
+
         // TODO: resolve providers in parallel
         let provider = match external.load_layer_provider(provider_listing.id).await {
             Ok(provider) => provider,
@@ -217,7 +221,7 @@ async fn get_layer_providers<C: ApplicationContext>(
                 collection_id,
             },
             name: provider_listing.name,
-            description: provider_listing.description,
+            description: provider.description().to_owned(),
             properties: Default::default(),
         }));
     }
--- a/services/src/api/handlers/layers.rs
+++ b/services/src/api/handlers/layers.rs
@@ -187,7 +187,7 @@ async fn get_layer_providers<C: ApplicationContext>(
         })
         .await?
     {
-        if provider_listing.priority == i16::MIN {
+        if provider_listing.priority <= -1000 {
             continue; // skip providers that are disabled
         };
 
--- a/services/src/contexts/migrations/migration_0003_gbif_config.rs
+++ b/services/src/contexts/migrations/migration_0003_gbif_config.rs
@@ -48,12 +48,8 @@ mod tests {
     use tokio_postgres::NoTls;
 
     use crate::contexts::{Migration0001RasterStacks, Migration0002DatasetListingProvider};
-    use crate::datasets::external::gbif::GBIF_PROVIDER_ID;
-    use crate::layers::storage::LayerProviderDb;
     use crate::{
-        contexts::{
-            migrate_database, migrations::migration_0000_initial::Migration0000Initial, PostgresDb,
-        },
+        contexts::{migrate_database, migrations::migration_0000_initial::Migration0000Initial},
         util::config::get_config_element,
     };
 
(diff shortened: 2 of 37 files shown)


jdroenner:

services/src/contexts/migrations/migration_0003_gbif_config.rs
```diff
@@ -106,11 +102,13 @@ mod tests {
         // drop the connection because the pool is limited to one connection, s.t. we can reuse the temporary schema
         drop(conn);
 
+        /* Note: this only works if the rust type is
```
michaelmattig, line 105: either remove it or write some SQL that checks the provider survived the migration

services/src/contexts/migrations/migration_0004_dataset_listing_provider_prio.rs
```diff
@@ -0,0 +1,141 @@
...
+        tx.batch_execute(
+            r#"
+            ALTER TABLE layer_providers
+            ADD COLUMN priority smallint NOT NULL DEFAULT 0;
+
+            ALTER TYPE "DatasetLayerListingProviderDefinition"
+            ADD ATTRIBUTE priority smallint;
+
+            ALTER TYPE "ArunaDataProviderDefinition"
+            ADD ATTRIBUTE description text;
```
michaelmattig, line 32: wouldn't it make sense to store data that is common to all data provider, in the layer_providers table?

services/src/layers/listing.rs
```diff
@@ -94,6 +94,16 @@ pub trait LayerCollectionProvider {
...
 
+    /// get the name of the provider
+    fn name(&self) -> &str;
+
+    /// get the description of the provider
+    fn description(&self) -> &str;
+
+    /// get the priority of the provider
+    // higher priority providers are preferred over lower priority providers
+    // fn priority(&self) -> i8;
```
michaelmattig, line 105: remove?

services/src/contexts/migrations/migration_0004_dataset_listing_provider_prio.rs
```diff
@@ -0,0 +1,141 @@
...
+            ALTER TYPE "ArunaDataProviderDefinition"
+            ADD ATTRIBUTE priority smallint;
```
michaelmattig, line 34: this is already stored in the layer_providers table. We could remove it from the corresponding Db types and not store it in the database.

services/src/api/handlers/layers.rs
```diff
@@ -187,6 +187,10 @@ async fn get_layer_providers<C: ApplicationContext>(
         })
         .await?
     {
+        if provider_listing.priority == i16::MIN {
```
michaelmattig, line 190: unrelated to this line: The Data Catalogue will always be on top, right? Is that ok?

test_data/provider_defs/user_datasets.json
```diff
@@ -1,8 +1,9 @@
 {
   "type": "DatasetLayerListingProviderDefinition",
   "id": "cbb21ee3-d15d-45c5-a175-66964adf4e85",
-  "name": "User Data Listing",
-  "description": "User specific datasets grouped by tags.",
+  "name": "Personal Data Catalog",
```
michaelmattig, line 4: why did you change it? Was there already some feedback?

jdroenner: yes there is some feedback in Slack.

services/src/api/handlers/layers.rs
jdroenner: yes it is. I think it should be always on top since it is the main entry point.

services/src/contexts/migrations/migration_0004_dataset_listing_provider_prio.rs
jdroenner: maybe. It will require more logic so we decided to keep it like this.

services/src/layers/listing.rs
jdroenner: https://github.com/geo-engine/geoengine/pull/920/commits/3943d688b4c8460eb0f5315f13e5c2f93e4b563b

services/src/layers/listing.rs
```diff
@@ -94,6 +94,16 @@ pub trait LayerCollectionProvider {
     /// Retrieve the provider's capabilities
     fn capabilities(&self) -> ProviderCapabilities;
 
+    /// get the name of the provider
+    fn name(&self) -> &str;
```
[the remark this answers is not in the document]
jdroenner: they are used in the listing method

services/src/contexts/migrations/migration_0003_gbif_config.rs
jdroenner: [c3fe060](https://github.com/geo-engine/geoengine/pull/920/commits/c3fe060c999ac8b2bb6d30dbd1496247b74dff0e)

services/src/layers/postgres_layer_db.rs
```diff
@@ -917,10 +940,10 @@ where
             SELECT 
                 id, 
                 name,
-                type_name
+                priority
             FROM 
                 layer_providers
-            ORDER BY name ASC
+            ORDER BY priority DESC, name ASC
```
michaelmattig, line 946: I think here and in pro you should also proably do "WHERE priority > -1000", otherwise the user might get fewer than expected results, if they are only filtered out in the handler afterwards.

jdroenner: https://github.com/geo-engine/geoengine/pull/920/commits/6cbdadc04008488233e0737326dab2f535e6db30

michaelmattig (approved):

PR #924, opened 2024-01-23
no unwrap

ChristianBeilschmidt:

---

<TEXT>

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,3 +1,7 @@
 avoid-breaking-exported-api = false
+
 # Adding "OpenAPI" is necessary because the word is used in a doc string in geoengine_services::api.
 doc-valid-idents = ["OpenAPI", ".."]
+
+# It is okay to being able to easily write tests.
+allow-unwrap-in-tests = true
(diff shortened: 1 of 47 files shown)


ChristianBeilschmidt:
[Unwraps im Code](https://trello.com/c/BtzuoNe3/1581-unwraps-im-code)

services/src/pro/util/mod.rs
```diff
@@ -1,4 +1,6 @@
 pub mod config;
 pub mod openapi_examples;
+// TODO: this should actually be only used in tests
```
michaelmattig, line 3: can you not move it into the `tests` module then?

clippy.toml
```diff
@@ -1,3 +1,7 @@
 avoid-breaking-exported-api = false
+
```
michaelmattig, line 2: here is still an unwrap:
https://github.com/geo-engine/geoengine/blob/a17c117c875a83d3566b56efc977ce31c35e6401/datatypes/src/lib.rs#L60

so I guess macros are not checked?

ChristianBeilschmidt: It looks like it:
[a38e04b](https://github.com/geo-engine/geoengine/pull/924/commits/a38e04b63310ff99cf9fe9080b65a045a518b4df)

But it could also be since the macro code is expanded only in test modules, that it was no problem.

michaelmattig (approved):

PR #926, opened 2024-01-29
stream raster bands

michaelmattig:

---

<TEXT>

### DIFF
--- a/services/src/api/model/datatypes.rs
+++ b/services/src/api/model/datatypes.rs
@@ -1,8 +1,7 @@
-use crate::error::{self, Result};
+use crate::error::{self, Error, Result};
 use crate::identifier;
 use geoengine_datatypes::primitives::{
-    AxisAlignedRectangle, BandSelection, MultiLineStringAccess, MultiPointAccess,
-    MultiPolygonAccess,
+    AxisAlignedRectangle, MultiLineStringAccess, MultiPointAccess, MultiPolygonAccess,
 };
 use ordered_float::NotNan;
 use postgres_types::{FromSql, ToSql};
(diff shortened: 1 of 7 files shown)


michaelmattig:
[Man kann in Python mit Multiband-Rastern arbeiten](https://trello.com/c/NYpFDvcG/1579-man-kann-in-python-mit-multiband-rastern-arbeiten)

services/src/api/model/datatypes.rs
```diff
@@ -992,11 +991,31 @@ impl From<QueryRectangle<SpatialPartition2D>>
             spatial_bounds: value.spatial_bounds.into(),
             time_interval: value.time_interval.into(),
             spatial_resolution: value.spatial_resolution.into(),
-            attributes: BandSelection::first(), // TODO: adjust once API supports attribute selection
+            attributes: geoengine_datatypes::primitives::BandSelection::first(), // TODO: adjust once API supports attribute selection
```
ChristianBeilschmidt, line 994: try_into and remove comment?

services/src/api/model/datatypes.rs
```diff
@@ -1567,7 +1586,7 @@ pub enum RasterColorizer {
 impl RasterColorizer {
     pub fn band_selection(&self) -> BandSelection {
         match self {
-            RasterColorizer::SingleBand { band, .. } => BandSelection::new_single(*band),
+            RasterColorizer::SingleBand { band, .. } => BandSelection(vec![*band as usize]),
```
ChristianBeilschmidt, line 1589: why did you remove the constructor?

ChristianBeilschmidt (changes requested):

services/src/api/model/datatypes.rs
michaelmattig: no, the comment is about the fact, that the `QueryRectangle` in the API does not have the `attributes` yet, because they are not yet used by clients. That's why it always select the first band here instead.

services/src/api/model/datatypes.rs
michaelmattig: I didn't remove the constructur but the type is now from `api` and not from `datatype`. The former one just doesn't have this method because it's only created by deserialization of user data and not from code.

ChristianBeilschmidt (approved):

PR #933, opened 2024-02-16
Symbology-manager

michaelmattig:

---

<TEXT>


michaelmattig:
[[DEVOPS-434] Manager: Editieren von Symbologies](https://trello.com/c/5zf2Dcef/1597-devops-434-manager-editieren-von-symbologies)

services/src/api/handlers/datasets.rs
```diff
@@ -251,6 +258,49 @@ pub async fn get_loading_info_handler<C: ApplicationContext>(
     Ok(web::Json(dataset))
 }
 
+/// Updates the dataset's symbology
+#[utoipa::path(
+    tag = "Datasets",
+    post,
```
ChristianBeilschmidt, line 264:
```suggestion
    put,
```

https://stackoverflow.com/questions/630453/what-is-the-difference-between-post-and-put-in-http

you update a symbology of a fixed resource and the output is always the same

services/src/pro/datasets/postgres.rs
```diff
@@ -646,6 +647,32 @@ where
...
+            self.has_permission_in_tx(dataset, Permission::Owner, &tx)
+                .await?,
+            error::PermissionDenied
+        );
+
+        let stmt = tx
+            .prepare("UPDATE datasets SET symbology = $2 WHERE id = $1;")
+            .await?;
+
+        tx.execute(&stmt, &[&dataset, &symbology]).await?;
```
ChristianBeilschmidt, line 669: I guess you don't need to prepare if you only use this statement once.
From what I've read in the PostgreSQL package, this is safe because they use the query and the params using the protocol and not concatenating them into a string.

services/src/datasets/postgres.rs
```diff
@@ -548,6 +549,22 @@ where
...
+        dataset: DatasetId,
+        symbology: &Symbology,
+    ) -> Result<()> {
+        let conn = self.conn_pool.get().await?;
+
+        let stmt = conn
+            .prepare("UPDATE datasets SET symbology = $2 WHERE id = $1;")
+            .await?;
+
+        conn.execute(&stmt, &[&dataset, &symbology]).await?;
```
ChristianBeilschmidt, line 563: same here

by the way, why do we have this method twice?

ChristianBeilschmidt (changes requested):

services/src/api/handlers/datasets.rs
michaelmattig: 6215207e120a5308d1672fe136b500bc4e1ab300

services/src/datasets/postgres.rs
michaelmattig: d5a4af89d4158c1e587be6e14db88fafd70ec890

we have it twice because of pro and non pro, where only pro needs to check permissions.

ChristianBeilschmidt (approved):

PR #934, opened 2024-02-20
make dataset info updatable

michaelmattig:

---

<TEXT>

### DIFF
--- a/services/src/api/handlers/datasets.rs
+++ b/services/src/api/handlers/datasets.rs
@@ -2639,4 +2639,37 @@ mod tests {
 
         Ok(())
     }
+
+    #[ge_context::test()]
+    async fn it_updates_dataset(app_ctx: PostgresContext<NoTls>) -> Result<()> {
+        let session_id = app_ctx.default_session_id().await;
+
+        let ctx = app_ctx.default_session_context().await?;
+
+        let (dataset_id, dataset_name) = add_ndvi_to_datasets(&app_ctx).await;
+
+        let update: UpdateDataset = UpdateDataset {
+            name: DatasetName::new(None, "new_name"),
+            display_name: "new display name".to_string(),
+            description: "new description".to_string(),
+        };
+
+        let req = actix_web::test::TestRequest::post()
+            .uri(&format!("/dataset/{dataset_name}"))
+            .append_header((header::CONTENT_LENGTH, 0))
+            .append_header((header::AUTHORIZATION, Bearer::new(session_id.to_string())))
+            .set_json(update.clone());
+        let res = send_test_request(req, app_ctx).await;
+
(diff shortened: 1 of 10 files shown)


michaelmattig:
[[DEVOPS-434] Manager: Editieren der General Info](https://trello.com/c/6aij3eUR/1598-devops-434-manager-editieren-der-general-info)

services/src/api/handlers/datasets.rs
```diff
@@ -220,6 +221,58 @@ pub async fn get_dataset_handler<C: ApplicationContext>(
     Ok(web::Json(dataset))
 }
 
+/// Update details about a dataset using the internal name.
+#[utoipa::path(
+    tag = "Datasets",
+    post,
```
ChristianBeilschmidt, line 227: Isn't this again put because of idempotency?

services/src/api/model/responses/datasets/errors.rs
```diff
@@ -50,3 +50,21 @@ impl ResponseError for GetDatasetError {
         StatusCode::BAD_REQUEST
     }
 }
+
+#[derive(Debug, Snafu, IntoStaticStr)]
+#[snafu(visibility(pub(crate)))]
+#[snafu(context(suffix(false)))] // disables default `Snafu` suffix
+pub enum UpdateDatasetError {
+    CannotLoadDatasetForUpdate { source: error::Error },
+    CannotUpdateDataset { source: error::Error },
```
ChristianBeilschmidt, line 59: Please add a display for the errors.

ChristianBeilschmidt (changes requested):

services/src/api/handlers/datasets.rs
michaelmattig: it's not idempotent unfortunately. If you have dataset named "A" and you rename it to "B", trying to rename it again will fail because dataset A does no longer exist.

something like this would happen:
> POST /datastet/A 
> {
>  "name": "B"
> }
> 200 OK
>
> POST /datastet/A 
> {
>  "name": "B"
> }
> 404 NOT FOUND

services/src/api/model/responses/datasets/errors.rs
ChristianBeilschmidt: [d10f183](https://github.com/geo-engine/geoengine/pull/934/commits/d10f18383d01222dae649c56b3a16652b69f2b65)

Don't you want to add the reason (source)?

In theory, it should only display things for the user.

michaelmattig: the source is currently a tokio postgres error. I would have to do build more error handling into it to give the source.

ChristianBeilschmidt (approved):

PR #936, opened 2024-02-28
Manage permissions

michaelmattig:

---

<TEXT>


michaelmattig:
[[DEVOPS-434] Manager: Editieren der Permissions](https://trello.com/c/VeFOXE4G/1599-devops-434-manager-editieren-der-permissions)

services/src/pro/permissions/mod.rs
```diff
@@ -150,27 +151,66 @@ pub struct PermissionListing {
...
+    #[snafu(display("Permission {permission:?} for resource {resource_id:?} denied."))]
+    PermissionDenied {
+        resource_id: ResourceId,
+        permission: Permission,
+    },
+    #[snafu(display("Must be admin to perform this action."))]
+    MustBeAdmin,
+    #[snafu(display(
+        "Permission {permission:?} for resource {resource_id:?} and roles {} not found.", role_ids.iter().map(std::string::ToString::to_string).collect::<Vec<String>>().join(", ")
+    ))]
```
ChristianBeilschmidt, line 166: You must not use the `Debug` output for `Display`.
It is important that we don't leak debug messages to the user.
You should use the `Display` output without `?` instead.

ChristianBeilschmidt (changes requested):

services/src/pro/permissions/mod.rs
michaelmattig: 787e9011a39a9c60ca85bc897d6cca57b5837e6f

services/src/contexts/migrations/migration_0006_owner_role.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct Migration0006OwnerRole;
+
+#[async_trait]
+impl Migration for Migration0006OwnerRole {
+    fn prev_version(&self) -> Option<DatabaseVersion> {
+        Some("0005_gbif_column_selection".into())
+    }
+
+    fn version(&self) -> DatabaseVersion {
+        "0006_owner_role".into()
```
ChristianBeilschmidt, line 18: 6 is now ebv.

You could either change it to 7 or say fuck it, no one ever used duplicate roles before now.

ChristianBeilschmidt (changes requested):
main must be merged

services/src/contexts/migrations/migration_0006_owner_role.rs
michaelmattig: 514bf67c5358edad41c2fe7d90c047e98d7b6644

ChristianBeilschmidt (approved):

PR #940, opened 2024-03-07
ebv provider uses main schema

ChristianBeilschmidt:

---

<TEXT>


ChristianBeilschmidt:
[EBV-Analyzer: Overview in DB](https://trello.com/c/Xdtuo8oL/1633-ebv-analyzer-overview-in-db)

services/src/datasets/external/netcdfcf/mod.rs
```diff
@@ -634,15 +618,12 @@ impl NetCdfCfDataProvider {
         resampling_method: Option<ResamplingMethod>,
         task_context: C,
     ) -> Result<OverviewGeneration> {
-        let mut db_connection = self
-            .metadata_db
-            .get()
-            .await
-            .boxed_context(error::DatabaseConnection)?;
+        let transaction = self.db.write_access(self.id).await?;
```
michaelmattig, line 621: is it a good idea to keep a database transaction open while an overview is created (could take potentially hours)?

services/src/datasets/external/netcdfcf/database.rs
```diff
@@ -0,0 +1,796 @@
...
+    transaction
+        .batch_execute("SET CONSTRAINTS ALL DEFERRED")
+        .await
+        .boxed_context(error::UnexpectedExecution)?;
+
+    Ok(transaction)
+}
+
+#[self_referencing]
+struct _NetCdfCfReadAccess<Tls>
```
michaelmattig, line 186: why do you not implement the logic inside the NetCdfCfProvider b instead of giving a Transaction to the outside?

services/src/datasets/external/netcdfcf/database.rs
```diff
@@ -0,0 +1,796 @@
...
+where
+    Tls: MakeTlsConnect<Socket> + Clone + Send + Sync + 'static,
+    <Tls as MakeTlsConnect<Socket>>::Stream: Send + Sync,
+    <Tls as MakeTlsConnect<Socket>>::TlsConnect: Send,
+    <<Tls as MakeTlsConnect<Socket>>::TlsConnect as TlsConnect<Socket>>::Future: Send,
+{
+    client: PooledConnection<'static, PostgresConnectionManager<Tls>>,
+    #[borrows(mut client)]
+    #[covariant] // TODO: check
```
michaelmattig, line 195: I don't know either.

ChristianBeilschmidt: It is, I have to remove it. The crate authors say they cannot determine it, but it would not compile if you specify it wrong.

services/src/datasets/external/netcdfcf/database.rs
ChristianBeilschmidt: I do the exact thing. I implement it in the provider and use this as DAO.

But it seems you have something else in mind, but you have to explain it to me in more detail, e.g., via phone.

services/src/datasets/external/netcdfcf/mod.rs
ChristianBeilschmidt: Good thought (which isn't new for this PR :stuck_out_tongue:).
Pro: It aligns with the overview creation. If this fails, the transaction fails. If it succeeds, it succeeds.
Con: Long-living transactions.

I could manully add another transaction by adding an in-progress flag into the database?

PR #942, opened 2024-03-12
Compare schema after migrations to current schema

ChristianBeilschmidt:

---

<TEXT>


ChristianBeilschmidt:
[Ground-Truth-Schema hinzufügen](https://trello.com/c/UB22Yw7t/1631-ground-truth-schema-hinzuf%C3%BCgen)

services/src/contexts/migrations/current_schema.sql
```diff
@@ -0,0 +1,875 @@
+CREATE TABLE geoengine (
+    clear_database_on_start boolean NOT NULL DEFAULT false,
+    database_version text NOT NULL
+);
+
```
michaelmattig, line 5: do you want to use the current_schema for uninitialized databases or is this a separate task?

services/src/contexts/migrations/schema_info.rs
```diff
@@ -0,0 +1,401 @@
...
+        .iter()
+        .map(|row| row.get("table_name"))
+        .collect::<Vec<String>>())
+}
+
+pub struct AssertSchemaEqConfig {
+    pub has_views: bool,
+}
+
+pub async fn assert_schema_eq(
```
michaelmattig, line 284: seems to work, but it's not really clear to me that we actually check everything here.

Maybe it would also be a good idea to write 2-3 sentences about the approach, because the code is not really that easy to understand.

services/src/contexts/migrations/schema_info.rs
```diff
@@ -0,0 +1,401 @@
...
+        .await?
+        .iter()
+        .map(|row| row.get("table_name"))
+        .collect::<Vec<String>>())
+}
+
+pub struct AssertSchemaEqConfig {
+    pub has_views: bool,
+}
+
```
michaelmattig, line 283: you could maybe add a simple test on a simple schema that shows that the check works.

services/src/contexts/migrations/current_schema.sql
ChristianBeilschmidt: Yes, separate task. First merge this one.

services/src/contexts/migrations/schema_info.rs
ChristianBeilschmidt: [04905a8](https://github.com/geo-engine/geoengine/pull/942/commits/04905a83d155787fb7031fdbf222bc59e208ec1a)

services/src/contexts/migrations/schema_info.rs
ChristianBeilschmidt: [04905a8](https://github.com/geo-engine/geoengine/pull/942/commits/04905a83d155787fb7031fdbf222bc59e208ec1a)

services/src/contexts/migrations/schema_info.rs
ChristianBeilschmidt: [dae5b3f](https://github.com/geo-engine/geoengine/pull/942/commits/dae5b3fcab180d5af0a5eabb0718b43a67362df2)

[8ef77ae](https://github.com/geo-engine/geoengine/pull/942/commits/8ef77aed6bb0d069f0364b4fe2f438728628aed4)

michaelmattig (approved):

PR #945, opened 2024-03-20
migrate from current schema if possible

ChristianBeilschmidt:

---

<TEXT>


ChristianBeilschmidt:
[Ground-Truth-Schema hinzufügen](https://trello.com/c/UB22Yw7t/1631-ground-truth-schema-hinzuf%C3%BCgen)

services/src/contexts/migrations/database_migration.rs
```diff
@@ -106,24 +108,24 @@ where
             }
         );
 
-        info!(target: prefix, "Applying migration: {}", migration.version());
+        info!("Applying migration: {}", migration.version());
```
michaelmattig, line 111: if the database schema is created from scratch it should not speak of migrations but say "initializing database" IMHO

services/src/contexts/migrations/database_migration.rs
```diff
@@ -106,24 +108,24 @@ where
...
 
         migration.migrate(&tx).await?;
 
-        let stmt = tx
-            .prepare("UPDATE geoengine SET database_version = $1")
-            .await?;
-
-        tx.execute(&stmt, &[&migration.version()]).await?;
+        tx.execute(
+            "UPDATE geoengine SET database_version = $1",
```
michaelmattig, line 118: this is not necessary if it is created from scratch and the current_schema.sql already set it.

services/src/contexts/migrations/database_migration.rs
```diff
@@ -70,30 +70,32 @@ pub enum MigrationResult {
 /// # Panics
 ///
 /// Panics if there is an error in the migration logic.
+///
 pub async fn migrate_database<Tls>(
-    conn: &mut PooledConnection<'_, PostgresConnectionManager<Tls>>,
+    connection: &mut PooledConnection<'_, PostgresConnectionManager<Tls>>,
     migrations: &[Box<dyn Migration>],
-    prefix: Option<&str>,
+    migration_if_uninitialized: Option<Box<dyn Migration>>,
```
michaelmattig, line 77: I feel like this shouldn't be a migration. Why should this be a boxed option? There should be exactly one initialization logic to be applied when the schema is uninitialized.

services/src/contexts/migrations/current_schema.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+#[async_trait]
+impl Migration for CurrentSchemaMigration {
+    fn prev_version(&self) -> Option<DatabaseVersion> {
+        None
+    }
+
+    fn version(&self) -> DatabaseVersion {
+        all_migrations()
```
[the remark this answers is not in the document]
ChristianBeilschmidt: How should it do it? You cannot know the other option, e.g., if the schema should be wiped.

services/src/contexts/migrations/database_migration.rs
ChristianBeilschmidt: cf. above

services/src/contexts/migrations/database_migration.rs
ChristianBeilschmidt: I designed it this way because you designed it to get a set of migrations as input.
Already now, you are required to input a list of migration. If you skip the first one, it is not applied.

ChristianBeilschmidt: [eaf44d5](https://github.com/geo-engine/geoengine/pull/945/commits/eaf44d5dbb00383dc5c00149813d0f445adcb75e)

services/src/contexts/migrations/database_migration.rs
ChristianBeilschmidt: [eaf44d5](https://github.com/geo-engine/geoengine/pull/945/commits/eaf44d5dbb00383dc5c00149813d0f445adcb75e)

services/src/contexts/migrations/current_schema.rs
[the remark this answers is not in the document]
michaelmattig: I mean instead of taking the version number from the latest migration you could hard code it in the current_schema.sql and increment it manually when you adjust it for a new migration. This would make the current_schema.sql self-contained.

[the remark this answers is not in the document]
ChristianBeilschmidt: This ia still only one column and you don't know the other columns of that table like wipe/keepDb. Thus, it is not self-contained.

michaelmattig (approved):

PR #946, opened 2024-03-25
Aruna 2.0.0 beta

glombiewski:

---

- Changed Aruna Provider Version to Aruna 2.0.0


glombiewski:
[Update Aruna Dependency](https://trello.com/c/vWDTppNY/1518-update-aruna-dependency)

.github/workflows/ci.yml
```diff
@@ -44,10 +44,10 @@ jobs:
           large-packages: true
           docker-images: true
           swap-storage: true
-      - name: Install lld & GDAL
+      - name: Install lld & GDAL & Protobuf
```
michaelmattig, line 47: Please add it the the README as well

services/src/datasets/external/aruna/mod.rs
```diff
@@ -212,106 +217,299 @@ impl ArunaDataProvider {
...
-        let mut aruna_object_group_overview = object_group_stub
-            .get_object_groups(GetObjectGroupsRequest {
-                collection_id: collection_id.clone(),
-                page_request: None,
-                label_id_filter: self.label_filter.clone(),
-            })
+        if let Some(filter) = label {
+            for key_value in key_values {
+                if key_value.variant() == KeyValueVariant::Label
+                    && key_value.key == *filter
```
michaelmattig, line 260: so key = value?

services/src/datasets/external/aruna/mod.rs
```diff
@@ -212,106 +217,299 @@ impl ArunaDataProvider {
...
-            .id;
-
-        let aruna_objects = object_group_stub
-            .get_object_group_objects(GetObjectGroupObjectsRequest {
-                collection_id: collection_id.clone(),
-                group_id: object_group_id.clone(),
-                page_request: None,
-                meta_only: false,
-            })
+        // //TODO: Workaround until deleted access permissions are fixed.
```
michaelmattig, line 299: what needs to be fixed there?

services/src/datasets/external/aruna/mod.rs
```diff
@@ -212,106 +217,299 @@ impl ArunaDataProvider {
...
+        let mut object_stub = self.object_stub.clone();
+
+        let requested_objects = object_stub
+            .get_objects(GetObjectsRequest { object_ids })
             .await?
             .into_inner()
-            .object_group_objects;
+            .objects;
+
+        let mut verified_objects = vec![];
```
michaelmattig, line 342: why do you collect the objects in the client? Is it not possible to filter in the server?

services/src/datasets/external/aruna/mod.rs
glombiewski: https://github.com/geo-engine/geoengine/pull/946/commits/e114b50df3d522a4f8ee5e84d08ab4389bce4b4a

Forgot to delete the comment.

There was a bug in Aruna when querying deleted datasets that needed the code in the comment as a workaround, but I think it is fixed by now.

services/src/datasets/external/aruna/mod.rs
glombiewski: The Object is just the metadata of the object (i.e., name, description, labels, ...), not the object itself. 

There is a [Search API](https://arunastorage.github.io/documentation/latest/get_started/basic_usage/12_How-To-Search) in Aruna, but I saw two problems with it based on our current Aruna implementation. Our current Aruna implementation assumes there is a Geo Engine project in Aruna that contains the datasets that should be visualized in Geo Engine for which we have an API key. The Search API queries all public objects.

(1) The Geo Engine project and its object might be private, or classified, or whatever. Just having public objects seems limiting. Public objects might also contain some sort of weird relationships between objects that we cannot properly handle.
(2) The documentation of the Search API does not specify fields to query the relation between object, i.e. this meta object belongs to that data object or these objects belong to the Geo Engine project...so we would have to verify the results afterwards anyway.

If the querying of metadata about objects becomes a bottleneck and/or the Search API is changed and/or we change the way we handle Aruna access, I think this should be revisited in another pull request. For now, I chose the solution that mirrors our previous handling of Aruna to update to the new dependency.

services/src/datasets/external/aruna/mod.rs
glombiewski: Yes, that is the way we previously handled the labels in Aruna.

services/src/datasets/external/aruna/mod.rs
```diff
@@ -178,8 +180,11 @@ impl ArunaDataProvider {
     /// Creates a new provider from the given definition.
     async fn new(def: Box<ArunaDataProviderDefinition>) -> Result<ArunaDataProvider> {
         let url = def.api_url;
+        let timeout = 10;
```
glombiewski, line 183: Ideally, this would be a configurable field in the Provider Definition. I assume this would require a migration for the database. There were already some new feature requests the past couple of days (i.e., not having a dedicated geo engine project, ...), so I think it would be best to bundle that into a feature that changes the the Provider Definition anyway.

michaelmattig (approved):

PR #947, opened 2024-04-04
check dataset name before insert

michaelmattig:

---

<TEXT>

### DIFF
--- a/services/src/datasets/postgres.rs
+++ b/services/src/datasets/postgres.rs
@@ -9,6 +9,7 @@ use crate::datasets::upload::FileId;
 use crate::datasets::upload::{Upload, UploadDb, UploadId};
 use crate::error::{self, Result};
 use crate::projects::Symbology;
+use crate::util::postgres::PostgresErrorExt;
 use async_trait::async_trait;
 use bb8_postgres::bb8::PooledConnection;
 use bb8_postgres::tokio_postgres::tls::{MakeTlsConnect, TlsConnect};
@@ -25,7 +26,6 @@ use geoengine_operators::engine::{
 use geoengine_operators::mock::MockDatasetDataSourceLoadingInfo;
 use geoengine_operators::source::{GdalLoadingInfo, OgrSourceDataset};
 use postgres_types::{FromSql, ToSql};
-use snafu::ensure;
 
 impl<Tls> DatasetDb for PostgresDb<Tls>
 where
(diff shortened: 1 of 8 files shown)


michaelmattig:
[[DEVOPS-434] Anlegen von Datasets](https://trello.com/c/06666bcI/1635-devops-434-anlegen-von-datasets)

services/src/datasets/postgres.rs
```diff
@@ -507,13 +508,21 @@ where
...
 
-        let stmt = conn
-            .prepare(
-                "
+        let existing_dataset = tx
+            .query_opt(
+                "SELECT TRUE FROM datasets WHERE name = $1::\"DatasetName\";",
+                &[&name],
+            )
+            .await?;
```
ChristianBeilschmidt, line 520: Isn't there a return code when inserting that does the same thing? I doubt there is the need for a second query.

services/src/pro/datasets/postgres.rs
```diff
@@ -588,7 +589,14 @@ where
 
         let tx = conn.build_transaction().start().await?;
 
-        // unique constraint on `id` checks if dataset with same id exists
+        let existing_dataset = tx
+            .query_opt(
+                "SELECT TRUE FROM datasets WHERE name = $1::\"DatasetName\";",
+                &[&name],
+            )
+            .await?;
```
ChristianBeilschmidt, line 597: same here

ChristianBeilschmidt (changes requested):

services/src/datasets/postgres.rs
michaelmattig: you mean like here: https://github.com/geo-engine/geoengine/blob/780870522d26c485982545b796c9468d54f894e8/services/src/pro/users/postgres_userdb.rs#L52-L59 ?

would also be possible but if we ever have another Unique constraint on the dataset, we wouldn't be able to know which one was violated. So maybe it's not a good idea to use this technique after all?

ChristianBeilschmidt: I guess we should also get that:

> For some types of errors, the server reports the name of a database object (a table, table column, data type, or constraint) associated with the error; for example, the name of the unique constraint that caused a unique_violation error. Such names are supplied in separate fields of the error report message so that applications need not try to extract them from the possibly-localized human-readable text of the message. As of PostgreSQL 9.3, complete coverage for this feature exists only for errors in SQLSTATE class 23 (integrity constraint violation), but this is likely to be expanded in future.

michaelmattig: d7e292016b37acad6e6578bbde878fb19c48c71d

services/src/pro/datasets/postgres.rs
michaelmattig: d7e292016b37acad6e6578bbde878fb19c48c71d

ChristianBeilschmidt (approved):

PR #960, opened 2024-06-05
operator for computing bandwise expressions

michaelmattig:

---

<TEXT>


michaelmattig:
[Operator, der auf jedem Pixel in jedem Band eine Operation ausführt.](https://trello.com/c/EgCWZLfg/1652-operator-der-auf-jedem-pixel-in-jedem-band-eine-operation-ausf%C3%BChrt)

operators/src/processing/bandwise_expression/mod.rs
```diff
@@ -0,0 +1,438 @@
...
+        // TODO: ensure all bands have same measurement unit?
+
+        let result_descriptor = in_descriptor.map_data_type(|_| self.params.output_type);
+
+        let parameters = vec![Parameter::Number("x".into())];
+
+        let expression = ExpressionParser::new(&parameters, DataType::Number)
+            .map_err(RasterExpressionError::from)?
+            .parse(
+                "expression", // TODO: what is the name used for?
```
jdroenner, line 64: what happens if we change it? Should we change it to be able to separate different kinds of exression?

operators/src/processing/bandwise_expression/mod.rs
```diff
@@ -0,0 +1,438 @@
...
+    fn result_descriptor(&self) -> &RasterResultDescriptor {
+        &self.result_descriptor
+    }
+
+    fn query_processor(&self) -> Result<TypedRasterQueryProcessor> {
+        let typed_raster_processor = self.source.query_processor()?.into_f64();
+
+        let output_type = self.result_descriptor().data_type;
+
+        // TODO: spawn a blocking task for the compilation process
```
jdroenner, line 99: could you do that?

operators/src/processing/bandwise_expression/mod.rs
```diff
@@ -0,0 +1,438 @@
...
+    ) -> Result<GridOrEmpty2D<TO>> {
+        let expression = unsafe {
+            // we have to "trust" that the function has the signature we expect
+            expression
+                .function_1::<Option<f64>>()
+                .map_err(RasterExpressionError::from)?
+        };
+
+        let map_fn = |in_value: Option<f64>| {
+            // TODO: could be a |in_value: T1| if map no data is false!
```
jdroenner, line 168: why only if no data is false? no data is a mask so id doesnt matter if it is a float type?

operators/src/processing/bandwise_expression/mod.rs
michaelmattig: this seems to be the name of the rust function that is generated. I don't know why it matters. In the other raster expression it's either the bandname or just "expression".

what do you mean with "kinds of expressions"?

operators/src/processing/bandwise_expression/mod.rs
michaelmattig: if we knew map_no_data is false the function from the expression does not need to be able to handle no data values and thus would not need to have an Option as input. I just copied the code inclusive comment from the other expression operator. This could just be an optimization.

operators/src/processing/bandwise_expression/mod.rs
michaelmattig: no, because the `query_processor` function is not async. I just copied the code here from the other expression operator inclusive the todo

jdroenner (approved):

PR #961, opened 2024-06-06
GBIF time and metadata

dbrandenstein:

---

<TEXT>


dbrandenstein:
[Umwandlung von GIBF Daten in raumzeitliche Daten](https://trello.com/c/FphsNRjY/1673-umwandlung-von-gibf-daten-in-raumzeitliche-daten)

operators/src/source/ogr_source/mod.rs
```diff
@@ -610,6 +610,64 @@ impl FeaturesProvider<'_> {
...
+        mut time_interval: TimeInterval,
+        driver: &str,
+    ) -> Option<String> {
+        match driver {
+            "PostgreSQL" => {
+                let postgres_range: TimeInterval = // https://www.postgresql.org/docs/current/datatype-datetime.html, capped by max supported TimeInstance value
+                    TimeInterval::new_unchecked(-210_895_056_000_000, 8_210_266_876_799_999);
+
+                if !postgres_range.contains(&time_interval) {
+                    return None;
```
michaelmattig, line 625: shouldn't that be an error at least log a warning?

operators/src/source/ogr_source/dataset_iterator.rs
```diff
@@ -119,16 +119,69 @@ impl OgrDatasetIterator {
...
 
         let mut features_provider = if let Some(sql) = dataset_information.sql_query.as_ref() {
+            let query = if let Some(filter) = final_filter {
+                debug!(
+                    "using attribute filter {:?} for layer {:?}",
+                    &filter, &dataset_information.layer_name
+                );
+
+                // This is necessary because otherwise the GDAL postgres driver does not perform a filter-pushdown in case an explicit SQL query is given
+                format!("SELECT * FROM ({sql}) q WHERE {filter}")
```
michaelmattig, line 162: is this nested select expensive? Can you have a look what the query optimizer does her (EXPLAIN PLAN)?
Wouldn't it be possible to extend the `sql` query by extending its `where` clause?

operators/src/source/ogr_source/mod.rs
```diff
@@ -610,6 +610,64 @@ impl FeaturesProvider<'_> {
...
 
+    ///  Creates a time filter string.
+    fn create_time_filter_string(
+        time_type: OgrSourceDatasetTimeType,
+        mut time_interval: TimeInterval,
+        driver: &str,
+    ) -> Option<String> {
+        match driver {
+            "PostgreSQL" => {
+                let postgres_range: TimeInterval = // https://www.postgresql.org/docs/current/datatype-datetime.html, capped by max supported TimeInstance value
```
michaelmattig, line 621: where do the constants come from? I don't find them in the linked doc.

operators/src/source/ogr_source/mod.rs
```diff
@@ -610,6 +610,64 @@ impl FeaturesProvider<'_> {
...
+                duration,
+            } => {
+                match duration {
+                    OgrSourceDurationSpec::Infinite => {
+                        Some(format!(r#""{start_field}" < '{t_end}'"#))
+                    }
+                    OgrSourceDurationSpec::Zero => Some(format!(
+                        r#""{start_field}" {comp_end} '{t_end}' AND "{start_field}" >= '{t_start}'"#
+                    )),
+                    OgrSourceDurationSpec::Value(_) => None, // TODO
```
michaelmattig, line 656: throw an error then or do you want to implement it still?

operators/src/source/ogr_source/mod.rs
```diff
@@ -1034,6 +1092,7 @@ where
         chunk_byte_size: usize,
     ) -> Result<FeatureCollection<G>> {
         let was_spatial_filtered_by_ogr = feature_iterator.was_spatial_filtered_by_ogr();
+        let was_time_filtered_by_ogr = false; // TODO currently, we don't know whether a filter was applied at this point, so we filter again
```
michaelmattig, line 1095: can you not pass this info into the function?

operators/src/source/ogr_source/dataset_iterator.rs
```diff
@@ -119,16 +119,69 @@ impl OgrDatasetIterator {
...
+            .or_else(|| time_filter.clone());
 
         let mut features_provider = if let Some(sql) = dataset_information.sql_query.as_ref() {
+            let query = if let Some(filter) = final_filter {
+                debug!(
+                    "using attribute filter {:?} for layer {:?}",
+                    &filter, &dataset_information.layer_name
+                );
+
+                // This is necessary because otherwise the GDAL postgres driver does not perform a filter-pushdown in case an explicit SQL query is given
```
michaelmattig, line 161: what about the spatial filter? does that not interfere with the sql_query?

operators/src/source/ogr_source/mod.rs
```diff
@@ -610,6 +610,64 @@ impl FeaturesProvider<'_> {
         Ok(())
     }
 
+    ///  Creates a time filter string.
+    fn create_time_filter_string(
```
michaelmattig, line 614: can you add tests for the different cases that can occur?

operators/src/source/ogr_source/dataset_iterator.rs
dbrandenstein: The plans seem to be equivalent. Of course, the inner query must not have something like a limit, then it would not be equivalent. However, this is also how GDAL handles spatial filters. It takes the SQL query as an inner query and adds the spatial filter to the outer query. Of course, it's also possible to extend the WHERE clause instead. The only advantage I see with this, though, is that one can then filter on columns which are not SELECTed in the SQL query through the attribute filter. However, you can already do this now in the explicit query you pass, so I don't really see this as an advantage and would personally prefer this easier approach without modifying the query string.

operators/src/source/ogr_source/mod.rs
dbrandenstein: I can log it as a warning if you want, something like "Time filter was not pushed down to PG because it exceeds the possible range". However, I don't consider it to be an error. What it means is simply that we cannot safely create a WHERE clause which Postgres can use to filter by time on its end, because the endpoints of the query time range are out of PG's timestamp type range. So in this case we skip the filtering and filter on the GE side. In practice, if the column is in fact of the timestamp type, we won't get any values out of range so we could also clip to PG's supported range. But then it would not filter anything anyways so we can just omit the filter as well, which I consider to be the safer approach.
(If the column were of type string, for example, PG would not try to convert the endpoints to a PG timestamp, so then it would not fail. But comparing timestamps by string comparison leads to a whole other lot of potential issues, so I don't think this should be addressed in this PR. The push down is only used for GBIF rn and there we know that the timestamps are represented by the PG timestamp type.)

operators/src/source/ogr_source/mod.rs
dbrandenstein: They are the low and high values from the timestamp with/without time zone in table 8.9

dbrandenstein: converted to ms and then capped by the maximum suported TimeInstance value, because the GE won't be able to handle larger values anyways how I understand it

operators/src/source/ogr_source/mod.rs
dbrandenstein: Same as above. I would not throw an error. Returning None here simply means that we defer the time filtering to the GE instead of already filtering with PG. This case does not occur with the GBIF data and is potentially complicated since the addition of the start field and the duration value must be possible

operators/src/source/ogr_source/dataset_iterator.rs
dbrandenstein: As mentioned in another comment above, for the spatial filter GDAL uses the SQL query as an inner query and adds a spatial filter to the outer query. It does not do this for other filter conditions (attribute filters, where I include the time filter), so this is done manually

services/src/datasets/external/gbif.rs
```diff
@@ -969,12 +970,49 @@ impl MetaDataProvider<OgrSourceDataset, VectorResultDescriptor, VectorQueryRecta
...
         Ok(Box::new(StaticMetaData {
             loading_info: OgrSourceDataset {
                 file_name: self.db_config.ogr_pg_config().into(),
                 layer_name: format!("{}.{}", self.db_config.schema, self.occurrence_table),
                 data_type: Some(VectorDataType::MultiPoint),
-                time: OgrSourceDatasetTimeType::None, // TODO
+                time: OgrSourceDatasetTimeType::Start {
+                    start_field: "eventdate".to_string(),
+                    start_format: OgrSourceTimeFormat::Auto,
+                    duration: OgrSourceDurationSpec::Zero,
```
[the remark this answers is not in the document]
dbrandenstein: If they want to see all occurrences in a certain month/year they should query a time range of this month/year. Is this not how it's expected to work?

operators/src/source/ogr_source/mod.rs
```diff
@@ -610,6 +610,64 @@ impl FeaturesProvider<'_> {
...
+            }
+            OgrSourceDatasetTimeType::StartEnd {
+                start_field,
+                start_format: _,
+                end_field,
+                end_format: _,
+            } => Some(format!(
+                r#""{start_field}" {comp_end} '{t_end}' AND "{end_field}" >= '{t_start}'"#
+            )),
+            OgrSourceDatasetTimeType::StartDuration { .. } => None, // TODO
```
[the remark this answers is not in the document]
dbrandenstein: Same as above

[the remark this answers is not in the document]
dbrandenstein: Does not occurr with GBIF rn and could be problematic if the addition of start and duration is not defined

operators/src/source/ogr_source/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/961/commits/e8e7d0f78a8a0220399e9feb62763dc199a895a9

dbrandenstein: https://github.com/geo-engine/geoengine/pull/961/commits/4dd828dcb5265edac96707e2d4fea3c8a2a66805

operators/src/source/ogr_source/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/961/commits/50d34cbb5e643cc87cb05bd08ba843180e81f081

services/src/datasets/external/gbif.rs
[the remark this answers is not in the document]
michaelmattig: sure, the users could do that. I am only asking because previously we enlarged the temporal validity of occurrences to the whole year, because that makes it easier to compare it with other layers. Otherwise users will have to work with time shifts to e.g. see all occurrences from a certain year over the temperature of a single month.

On default, VAT selects a time instant and not a range, so you will most likely never see a single occurrence when you add GBIF data to the map unless you change to a time interval.

michaelmattig (approved):

PR #963, opened 2024-06-07
Store OIDC tokens

glombiewski:

---

- Added AES password encryption for strings
- Added OIDC manager abstraction. An OIDC manager is always present, and either works or does not. This simplifies the Option<Arc<...>> handling of the previous top level OIDC abstraction.
- Stored access and refresh tokens in the database (encrypted if a password is in the config file).
- Use refresh tokens whenever a SessionContext is requested to extend the session if it has expired.


glombiewski:
[Speicherung und Benutzung von OIDC Tokens](https://trello.com/c/wLmXIBIy/1666-speicherung-und-benutzung-von-oidc-tokens)

services/src/pro/contexts/migrations/current_schema.sql
```diff
@@ -216,3 +216,12 @@ CREATE TABLE ml_models (
     id uuid PRIMARY KEY,
     content text NOT NULL
 );
+
+CREATE TABLE oidc_session_tokens (
+    session_id uuid REFERENCES sessions (id) ON DELETE CASCADE NOT NULL,
```
michaelmattig, line 221: add unique constraint? or can there be multiple tokens per session?

services/src/pro/contexts/migrations/migration_0000_initial.sql
```diff
@@ -211,3 +211,12 @@ CREATE TABLE ml_models (
     id uuid PRIMARY KEY,
     content text NOT NULL
 );
+
+CREATE TABLE oidc_session_tokens (
```
michaelmattig, line 215: why did you add it to the initial migration and not a new one?

services/src/pro/users/oidc.rs
```diff
@@ -359,29 +534,27 @@ impl OidcRequestDb {
...
-            issuer: value.issuer.to_string(),
-            client_id: value.client_id.to_string(),
-            client_secret: value.client_secret.clone(),
-            redirect_uri: value.redirect_uri.to_string(),
-            scopes: value.scopes,
-            users: Arc::new(Default::default()),
-            state_function: || CsrfToken::new(SINGLE_STATE.to_string()),
-            nonce_function: || Nonce::new(SINGLE_NONCE.to_string()),
-        }
+    async fn refresh_token(
```
michaelmattig, line 547: shouldn't this be called `access_token` as it return an access token?

services/src/pro/users/postgres_userdb.rs
```diff
@@ -425,6 +435,17 @@ where
...
             .map_err(|_error| error::Error::InvalidSession)?;
 
+        let valid_session: bool = row.get(7);
+
+        let valid_until = if valid_session {
+            row.get(4)
+        } else {
+            self.refresh_tokens(session, &tx)
+                .await
+                .map_err(|_error| Error::InvalidSession)?
```
michaelmattig, line 445: is there a more meaningful error here? When can the token refresh fail? if the refresh token is not valid anymore?

services/src/pro/users/postgres_userdb.rs
```diff
@@ -449,12 +470,204 @@ where
...
+                        &flat_tokens.access_token_nonce,
+                        &(oidc_tokens.expires_in.num_seconds() as f64),
+                        &flat_tokens.refresh_token_value,
+                        &flat_tokens.refresh_token_nonce,
+                    ],
+                )
+                .await?;
+
+            if updated_token_rows != 1 {
+                return Err(Error::InvalidSession);
```
michaelmattig, line 601: when can this happen and is there a more meaningful error (to log at least)?

services/src/pro/users/postgres_userdb.rs
```diff
@@ -405,18 +413,20 @@ where
...
-                s.view
-            FROM 
-                sessions s JOIN user_sessions us ON (s.id = us.session_id) 
+                s.view,
+                CASE WHEN CURRENT_TIMESTAMP < us.valid_until THEN TRUE ELSE FALSE END AS valid_session
+            FROM
+                sessions s JOIN user_sessions us ON (s.id = us.session_id)
                     JOIN users u ON (us.user_id = u.id)
-            WHERE s.id = $1 AND CURRENT_TIMESTAMP < us.valid_until;",
+                    LEFT JOIN oidc_session_tokens t ON (s.id = t.session_id)
```
michaelmattig, line 428: would it make sense to first check whether the session is an oidc session and then have a separate logic for the token refresh that is separated from the local login?

services/src/pro/users/postgres_userdb.rs
```diff
@@ -425,6 +435,17 @@ where
             .await
             .map_err(|_error| error::Error::InvalidSession)?;
 
+        let valid_session: bool = row.get(7);
+
+        let valid_until = if valid_session {
+            row.get(4)
+        } else {
+            self.refresh_tokens(session, &tx)
```
michaelmattig, line 443: don't you have to check whether the refresh token is still valid or where does this happen?

services/src/pro/users/oidc.rs
glombiewski: Correct. I think refresh_access_token describes the entire process. ee8a2fc34b0071a07bfcd86d003e9da7adb57ff2

services/src/pro/users/postgres_userdb.rs
glombiewski: Since the same transactions first checks if this row exists, I don't think this can occur. ee8a2fc34b0071a07bfcd86d003e9da7adb57ff2

services/src/pro/users/postgres_userdb.rs
glombiewski: ee8a2fc34b0071a07bfcd86d003e9da7adb57ff2

I put a log debug message with the original error. My intention of mapping the error was that session_by_id call either gets a session or returns InvalidSession to whomever calls the function. I believe most API handlers use this as a first step, and they probably shouldn't care or propagate to the user that some internal refresh mechanism failed? But you are right, it should be noted somewhere, so I opted for a log for now.

When can this fail: 
The token refresh can fail if the token is not valid anymore. Or if the client does not respond for any reason. The other call to refresh_tokens is from get_access_token, where there might be no refresh_token (InvalidSession).

services/src/pro/util/config.rs
```diff
@@ -37,14 +37,15 @@ pub enum QuotaTrackingMode {
...
-#[derive(Debug, Deserialize)]
+#[derive(Clone, Debug, Deserialize)]
 pub struct Oidc {
     pub enabled: bool,
     pub issuer: String,
     pub client_id: String,
     pub client_secret: Option<String>,
     pub redirect_uri: String, //TODO: Maybe URL type
     pub scopes: Vec<String>,
+    pub postgres_password: Option<String>,
```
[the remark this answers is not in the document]
glombiewski: Correct. 42eaf79b23a01e1176ac8ccaa39472d4ee46f9db

services/src/pro/users/postgres_userdb.rs
glombiewski: We discussed this and opted to keep it as is.

services/src/pro/users/postgres_userdb.rs
glombiewski: The Identity Provider does not send a expiration date with the refresh token. The refresh token might be invalid for a variety of reasons (expired, revoked, ...). The only way to check is to send it to the Identity Provider.

services/src/pro/contexts/migrations/current_schema.sql
glombiewski: Added a Primary Key, or would Unique be preferable for some reason? c09d87f913f283a000cff7c3c61c9aefb288fd1c

michaelmattig (approved):

PR #964, opened 2024-06-07
Increase SQL linter file size;

glombiewski:

---

- Increased the number of bytes for files in the sql linter such that the current_schema file is not skipped.


glombiewski:
[SQL linter überspring SQL file](https://trello.com/c/S6b2oGD3/1693-sql-linter-%C3%BCberspring-sql-file)

michaelmattig (approved):

PR #967, opened 2024-06-13
onnx operator (wip)

michaelmattig:

---

<TEXT>


michaelmattig:
[ONNX-Operator bauen](https://trello.com/c/kFjWkS4z/1575-onnx-operator-bauen)

operators/src/pro/machine_learning/onnx.rs
```diff
@@ -0,0 +1,473 @@
...
+            .await?
+            .chunks(num_bands) // chunk the tiles to get all bands for a spatial index at once
+            .map(move |chunk| {
+                if chunk.len() != num_bands {
+                    // if there are not exactly N tiles, it should mean the last tile was an error and the chunker ended prematurely
+                    if let Some(Err(e)) = chunk.into_iter().last() {
+                        return Err(e);
+                    }
+                    // if there is no error, the source did not produce all bands, which likely means a bug in an operator
+                    unreachable!("the source did not produce all bands");
```
jdroenner, line 146: i guess it should be an error not unreachable? I tought unreachable must never happen. Maybe also add a warn! log here?

operators/src/pro/machine_learning/onnx.rs
```diff
@@ -0,0 +1,473 @@
...
+                let first_tile = &tiles[0];
+                let time = first_tile.time;
+                let tile_position = first_tile.tile_position;
+                let global_geo_transform = first_tile.global_geo_transform;
+                let cache_hint = first_tile.cache_hint;
+
+                let width = tiles[0].grid_array.axis_size_x();
+                let height = tiles[0].grid_array.axis_size_y();
+
+                // TODO: collect into a ndarray directly
```
jdroenner, line 160: i guess we need to solve the nodata mistery here? Ndarray still has no nodata right?

operators/src/pro/machine_learning/onnx.rs
```diff
@@ -0,0 +1,473 @@
...
+
+                let mut pixels: Vec<Vec<f32>> = vec![vec![0.0; num_bands]; width * height];
```
jdroenner, line 162: i guess we should add an into_ndarray or something to Grid / MaskedGrid?

operators/src/pro/machine_learning/onnx.rs
```diff
@@ -0,0 +1,473 @@
...
+            .into_runnable()
+            .unwrap();
+
+        let array = arr2(&[[0.1f32, 0.2], [0.2, 0.3], [0.2, 0.2], [0.3, 0.1]]);
+
+        let tensor: Tensor = array.into();
+
+        // run the model on the input
+
+        // TODO: fix the input because the prediction is wrong(?)
```
jdroenner, line 248: is this still the case?

operators/src/pro/machine_learning/onnx.rs
michaelmattig: yes, for now I take NAN or zero on None, but we maybe should either just skip the pixel or even throw an error and let the user handle it, e.g. by putting an expression mapping nodata to something else beforehand?

operators/src/pro/machine_learning/onnx.rs
michaelmattig: yes. If we make a final decision on ort than we can just remove this test. I still can't get tract to work.

operators/src/pro/machine_learning/onnx.rs
michaelmattig: yes would make sense, but we would need a solution for the nodata case (cf my other comment). Also, here, we turn a stack of grids into an ndarray. So can't really be an `From` on Grid and we have no explicit type for stacks.

operators/src/pro/machine_learning/onnx.rs
jdroenner: i guess it also depends on the model. There are some that can handle a kind of nodata. I guess thats a nodata-value in most cases?

operators/Cargo.toml
```diff
@@ -59,11 +61,13 @@ tokio = { version = "1.36", features = [
 ] }
 tokio-postgres = "0.7"
 tracing = "0.1"
+tract-onnx = "0.21"
```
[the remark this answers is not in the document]
jdroenner: yes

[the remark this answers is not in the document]
michaelmattig: 07019881364e95b6abedb43483a4a80a1b598564

operators/src/error.rs
```diff
@@ -490,6 +490,54 @@ pub enum Error {
...
+    #[snafu(display("Number of bands in source ({source_bands}) does not match the model input bands ({model_input_bands})."))]
+    OnnxInputBandsMismatch {
+        model_input_bands: usize,
+        source_bands: usize,
+    },
+    #[snafu(display("Raster data types of source ({source_type:?}) does not match model input type ({model_input_type:?})."))]
+    OnnxInputTypeMismatch {
+        model_input_type: RasterDataType,
+        source_type: RasterDataType,
+    },
```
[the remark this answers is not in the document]
michaelmattig: cd90401e00c43ee52c15d87a0ffb75e466d14e9a

jdroenner (approved):

PR #968, opened 2024-06-17
First_last_nodata_fill

jdroenner:

---

<TEXT>

### DIFF
--- a/operators/src/mock/mock_raster_source.rs
+++ b/operators/src/mock/mock_raster_source.rs
@@ -8,10 +8,10 @@ use crate::util::Result;
 use async_trait::async_trait;
 use futures::{stream, stream::StreamExt};
 use geoengine_datatypes::dataset::NamedData;
-use geoengine_datatypes::primitives::CacheExpiration;
+use geoengine_datatypes::primitives::{CacheExpiration, CacheHint, TimeInterval};
 use geoengine_datatypes::primitives::{RasterQueryRectangle, SpatialPartitioned};
 use geoengine_datatypes::raster::{
-    GridShape2D, GridShapeAccess, GridSize, Pixel, RasterTile2D, TilingSpecification,
+    GridOrEmpty2D, GridShape2D, GridShapeAccess, GridSize, Pixel, RasterTile2D, TilingSpecification
 };
 use serde::{Deserialize, Serialize};
 use snafu::Snafu;
@@ -109,17 +109,82 @@ where
         _ctx: &'a dyn crate::engine::QueryContext,
     ) -> Result<futures::stream::BoxStream<crate::util::Result<RasterTile2D<Self::RasterType>>>>
     {
+        let mut parts: Vec<RasterTile2D<T>> = self.data
+        .iter()
+        .filter(move |t| {
+            t.time.intersects(&query.time_interval)
+                && t.tile_information()
+                    .spatial_partition()
+                    .intersects(&query.spatial_bounds)
+        })
+        .cloned().collect();
+
+        // This is a workaround to avoid errors when a provider does not fill the complete query rectangle.
+        if let Some(first) = parts.first() {
+            if first.time.start() > query.time_interval.start() {
+
+                // try to find an element before
+                let element_before = self.data.iter().filter(|item| item.time.end() < query.time_interval.start()).last();
+                
+                let (no_data_start, no_data_cache) = if let Some(eb) = element_before {
+                    if eb.time.is_instant() {
+                        (eb.time.end() +1, eb.cache_hint)    
+                    } else {
+                        (eb.time.end(), eb.cache_hint)
+                    }
+               
+                 } else {
+                    (query.time_interval.start(), CacheHint::max_duration()) // TODO: if there is no data before can we assume that it is MIN?
+                 };
+
+                parts.insert(0, RasterTile2D::new(
+                    TimeInterval::new(no_data_start, first.time.start()).expect("Only insert element thats before start"),
+                    first.tile_position, // we can use any tile position inside the query. the Fill adapter will take care of the rest.
+                    first.band,
+                    first.global_geo_transform,
+                    GridOrEmpty2D::new_empty_shape(first.grid_shape()),
+                    no_data_cache
+                ));
+            }
+        }
+
+        if let Some(last) = parts.last() {
+            if (
+               !last.time.is_instant() && last.time.end() < query.time_interval.end()
+            ) || (
+                last.time.is_instant() && last.time.end() +1 < query.time_interval.end()
+            ) {                
+
+                 // try to find an element following
+                 let element_following = self.data.iter().find(|item| item.time.start() > query.time_interval.end());
+
+                let no_data_start = if last.time.is_instant() {
+                    last.time.end() +1
+                } else {
+                    last.time.end()
+                };
+
+                 let (no_data_end, no_data_cache) = if let Some(ef) = element_following {
+                    (ef.time.start(), ef.cache_hint)
+                 } else {
+                    (query.time_interval.end(), CacheHint::max_duration()) 
+                 };
+
+                parts.push(RasterTile2D::new(
+                    TimeInterval::new(no_data_start, no_data_end).expect("Only insert element thats before start"),
+                    last.tile_position, // we can use any tile position inside the query. the Fill adapter will take care of the rest.
+                    last.band,
+                    last.global_geo_transform,
+                    GridOrEmpty2D::new_empty_shape(last.grid_shape()),
+                    no_data_cache
+                ));
+            }
+        }
+
+        // TODO: if there is no data at all the fill adapter will take care and produce a inf no data tile
+        
         let inner_stream = stream::iter(
-            self.data
-                .iter()
-                .filter(move |t| {
-                    t.time.intersects(&query.time_interval)
-                        && t.tile_information()
-                            .spatial_partition()
-                            .intersects(&query.spatial_bounds)
-                })
-                .cloned()
-                .map(Result::Ok),
+            parts.into_iter().map(Result::Ok),
         );
 
         // TODO: evaluate if there are GeoTransforms with positive y-axis
--- a/operators/src/source/gdal_source/loading_info.rs
+++ b/operators/src/source/gdal_source/loading_info.rs
@@ -30,18 +30,44 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
         let valid = self.time.unwrap_or_default();
 
         let parts = if valid.intersects(&query.time_interval) {
-            vec![GdalLoadingInfoTemporalSlice {
+            let mut p = vec![];
+            
+            if valid.start() > query.time_interval.start() {
+                log::debug!("Missing temporal slice information to provide loading info for the whole query rectangle. Inserting fake first element! Query time: {} First element awailable: {}.", query.time_interval, valid);
+                p.push(GdalLoadingInfoTemporalSlice {
+                    time: TimeInterval::new(query.time_interval.start(), valid.start()).expect("Condition is checked above"),
+                    params: None,
+                    cache_ttl: self.cache_ttl // TODO: maybe this needs to be zero.
+                });
+            }
+
+            p.push(GdalLoadingInfoTemporalSlice {
                 time: valid,
                 params: Some(self.params.clone()),
                 cache_ttl: self.cache_ttl,
-            }]
-            .into_iter()
+            });
+
+            if valid.end() < query.time_interval.end() {
+                log::debug!("Missing temporal slice information to provide loading info for the whole query rectangle. Inserting fake last element! Query time: {} Last element awailable: {}.", query.time_interval, valid);
+                p.push(GdalLoadingInfoTemporalSlice {
+                    time: TimeInterval::new(valid.end(), query.time_interval.end(), ).expect("Condition is checked above"),
+                    params: None,
+                    cache_ttl: self.cache_ttl // TODO: maybe this needs to be zero.
+                });
+            }
+
+            p
+            
         } else {
-            vec![].into_iter()
+            vec![GdalLoadingInfoTemporalSlice {
+                time: query.time_interval,
+                params: None,
+                cache_ttl: self.cache_ttl, // can't change?
+            }]
         };
 
         Ok(GdalLoadingInfo {
-            info: GdalLoadingInfoTemporalSliceIterator::Static { parts },
+            info: GdalLoadingInfoTemporalSliceIterator::Static { parts: parts.into_iter() },
         })
     }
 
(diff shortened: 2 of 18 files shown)


jdroenner:
[Daten für Lena vorbereiten](https://trello.com/c/7wwvqxIc/1639-daten-f%C3%BCr-lena-vorbereiten)

operators/src/mock/mock_raster_source.rs
```diff
@@ -109,18 +109,90 @@ where
...
+            .filter(move |t| {
+                t.time.intersects(&query.time_interval)
+                    && t.tile_information()
+                        .spatial_partition()
+                        .intersects(&query.spatial_bounds)
+            })
+            .cloned()
+            .collect();
+
+        // This is a workaround to avoid errors when a provider does not fill the complete query rectangle.
```
michaelmattig, line 124: what does the mock raster source have to do with providers?

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -30,18 +30,47 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
         let valid = self.time.unwrap_or_default();
 
         let parts = if valid.intersects(&query.time_interval) {
-            vec![GdalLoadingInfoTemporalSlice {
+            let mut p = vec![];
+
+            if valid.start() > query.time_interval.start() {
+                log::debug!("Missing temporal slice information to provide loading info for the whole query rectangle. Inserting fake first element! Query time: {} First element awailable: {}.", query.time_interval, valid);
+                p.push(GdalLoadingInfoTemporalSlice {
```
michaelmattig, line 37: wouldn't it make more sense to "fix" the GdalMetadataStatic on creation instead of on access?

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -30,18 +30,47 @@ impl MetaData<GdalLoadingInfo, RasterResultDescriptor, RasterQueryRectangle>
...
+                    time: TimeInterval::new(query.time_interval.start(), valid.start())
+                        .expect("Condition is checked above"),
+                    params: None,
+                    cache_ttl: self.cache_ttl, // TODO: maybe this needs to be zero.
```
michaelmattig, line 41: why should it be zero? I think it's fine to use the ttl of the data.

operators/src/mock/mock_raster_source.rs
michaelmattig: but also here: wouldn't it make more sense to insert the fillers at creation time (of the source) instead of query time?

operators/src/source/gdal_source/mod.rs
```diff
@@ -733,7 +734,27 @@ where
             loading_info.info
         };
 
-        let source_stream = stream::iter(loading_iter);
+        let loading_inter_checked = loading_iter.with_position().map(move |(pos, ele)| {
```
michaelmattig, line 737: so you go through all loading infos just to check the first and last? Wouldn't it suffice to lazily check the loading infos as the are consumed or is that to complicated?

operators/src/source/gdal_source/mod.rs
```diff
@@ -733,7 +734,27 @@ where
...
+                (Ok(e), Position::First) if q_time.start() < e.time.start() => {
+                    warn!("The data provider does not cover the query start. Query time {}, First element time {}", q_time, e.time);
+                    Err(error::GdalSourceError::MissingTemporalDataAtQueryStart{query_time: q_time, first_element_time: e.time}.into())
+                },
+                (Ok(e), Position::Last) if q_time.end() > e.time.end() => {
+                    warn!("The data provider does not cover the query end. Query time {}, Last element time {}", q_time, e.time);
+                    Err(error::GdalSourceError::MissingTemporalDataAtQueryEnd{query_time: q_time, last_element_time: e.time}.into())
+                },
+                (Ok(e), Position::Only) if q_time.start() < e.time.start() || q_time.end() > e.time.end() => {
+                    warn!("The data provider does not cover the query end. Query time {}, Only element time {}", q_time, e.time);
```
michaelmattig, line 749: why do you only check the end if there is only one element?

services/src/contexts/migrations/migration_0010_s2_stack_time_buffers.rs
```diff
@@ -0,0 +1,26 @@
...
+    fn prev_version(&self) -> Option<DatabaseVersion> {
+        Some("0009_oidc_tokens".into())
+    }
+
+    fn version(&self) -> DatabaseVersion {
+        "0010_s2_stac_time_buffers".into()
+    }
+
+    async fn migrate(&self, _tx: &Transaction<'_>) -> Result<()> {
+        // permissions only exist in Pro, nothing to do here
```
michaelmattig, line 22: remove comment

services/src/pro/contexts/migrations/migration_0010_s2_stack_time_buffers.rs
```diff
@@ -0,0 +1,32 @@
...
+use super::database_migration::{ProMigration, ProMigrationImpl};
+use crate::{contexts::Migration0010S2StacTimeBuffers, error::Result};
+
+#[async_trait]
+impl ProMigration for ProMigrationImpl<Migration0010S2StacTimeBuffers> {
+    async fn pro_migrate(&self, tx: &Transaction<'_>) -> Result<()> {
+        // add priority column to layer_providers table and move description column to the definition types
+        tx.batch_execute(
+            r#"
+            CREATE TYPE "StacQueryBuffer" AS (start_seconds bigint, end_seconds bigint);
```
michaelmattig, line 13: why are start and end individually configurable? Do you need this?

services/src/pro/contexts/migrations/migration_0010_s2_stack_time_buffers.rs
```diff
@@ -0,0 +1,32 @@
...
+            ADD ATTRIBUTE query_buffer "StacQueryBuffer";            
+        "#,
+        )
+        .await?;
+
+        tx.batch_execute(
+            r"
+            UPDATE pro_layer_providers
+            SET
+                definition.sentinel_s2_l2_a_cogs_provider_definition.query_buffer.start_seconds = -60,
```
michaelmattig, line 25: does it make sense that this is negative? I think it should be positive...

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -427,7 +454,7 @@ impl SentinelS2L2aCogsMetaData {
             let end = if i < num_features - 1 {
                 start_times[i + 1]
             } else {
-                start + 1000 // TODO: determine correct validity for last tile
+                start + 1000 // there was no tile before and there is no tile following. Use 1 sec to be safe. We could assume that there is no other tile in the query and use the query length?
```
michaelmattig, line 457: could we use the buffer here also? Or add another buffer to the config?

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -455,6 +482,49 @@ impl SentinelS2L2aCogsMetaData {
         }
         debug!("number of generated loading infos: {}", parts.len());
 
+        // This is a workaround to avoid errors when a provider does not fill the complete query rectangle.
+        let buffered_start =
+            query.time_interval.start() + Duration::seconds(self.stac_query_buffer.start_seconds);
```
michaelmattig, line 487: I would make the start_seconds positive and subtract here.

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -455,6 +482,49 @@ impl SentinelS2L2aCogsMetaData {
...
+        let buffered_end =
+            query.time_interval.end() + Duration::seconds(self.stac_query_buffer.end_seconds);
+
+        if let Some(first) = parts.first() {
```
michaelmattig, line 491: is this the same logic as in gdal and mock source? Maybe we should reuse the code

test_data/provider_defs/pro/sentinel_s2_l2a_cogs.json
```diff
@@ -59,5 +59,9 @@
       "name": "UTM37S",
       "epsg": 32737
     }
-  ]
+  ],
+  "query_buffer": {
+    "start_seconds": -60,
```
michaelmattig, line 64: use camel case?

operators/src/mock/mock_raster_source.rs
jdroenner: i removed the fillers and now all providers can provide known times covering the query time. If they don't the gdal source will fallback to the query time and log warnings (no error since it is not wrong only unfortunate wrt the cache)

operators/src/source/gdal_source/loading_info.rs
jdroenner: see above

operators/src/source/gdal_source/loading_info.rs
jdroenner: fill adapter will now take care of ttls

operators/src/source/gdal_source/mod.rs
jdroenner: changed that and the providers using lists and/or generators now iterate while generating. The only exception is the metadata regular which must iterate the time steps in advance. This is needed since we need the time bounds when initializing the adapter

operators/src/source/gdal_source/mod.rs
jdroenner: i removed that and use the filler instead

services/src/contexts/migrations/migration_0010_s2_stack_time_buffers.rs
jdroenner: [b8e0dd9](https://github.com/geo-engine/geoengine/pull/968/commits/b8e0dd9c5298aadc3bb8e9559db85b5451a9979e)

services/src/pro/contexts/migrations/migration_0010_s2_stack_time_buffers.rs
jdroenner: i think it is helpfull when we work with different kinds of granularity. It does not hurt to be able to set them individually, right?.

services/src/pro/contexts/migrations/migration_0010_s2_stack_time_buffers.rs
jdroenner: changed it to positive numbers and subtraction (sorry forgot extra commit)

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: [fea3921](https://github.com/geo-engine/geoengine/pull/968/commits/fea392130e5c99e71f6ce2a3f1292d6d402a4f89)

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: changed it (sorry no extra commit)

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: logic still is similar however, i don't know how to reuse in a good way

test_data/provider_defs/pro/sentinel_s2_l2a_cogs.json
jdroenner: changed it (sorry, no extra commit)

operators/src/processing/map_query.rs
```diff
@@ -58,6 +60,7 @@ where
                 &query,
                 self.additional_data,
                 FillerTileCacheExpirationStrategy::NoCache,
+                FillerTimeBounds::from(query.time_interval),
```
michaelmattig, line 63: add a comment to derive this from input stream instead?

operators/src/source/gdal_source/loading_info.rs
```diff
@@ -460,6 +538,47 @@ impl Iterator for NetCdfCfGdalLoadingInfoPartIterator {
...
+    /// This is the end of the last part produced to answer the query.
+    /// IF there is no data part that intersects the query end, this is the first time known after the query end.
+    /// ELSE IF there is a known element with an end after the query time it is the start of that `TimeInterval`.
+    /// ELSE IF there is no data at all (and never will be) it might be `TimeInstance::MAX`.
+    /// ELSE it should be None and the GdalSource will replace it with the start of the query AND issue a warning about missing information.
+    pub last_part_to_answer_query_time_end: Option<TimeInstance>,
+}
+
+impl GdalLoadingInfo {
+    /// This method generates a new `GdalLoadingInfo`.
```
michaelmattig, line 557: maybe start by explaining that the temporal slice iterator may not start/end with the query. in this case the first temporal slice starts after the query start and there is a gap that has to be filled with a nodata tile. This nodata tile needs a temporal validity, thus we need to know how long before the start of the first loading info there is no data. This information is passed as known_time_before/after

but also think about renaming this fields to something like:
- start_time_of_output_stream
- result_temporal_validity (and combining both time instances into a timeinterval)

services/src/datasets/external/aruna/mod.rs
```diff
@@ -650,11 +650,10 @@ impl ArunaDataProvider {
             cache_ttl,
         };
 
-        GdalLoadingInfo {
-            info: GdalLoadingInfoTemporalSliceIterator::Static {
-                parts: vec![part].into_iter(),
-            },
-        }
+        // FIXME: someone with more detailed insights needs to add the time bounds!
```
michaelmattig, line 653: maybe present the problem to Niki on Wednesday?

operators/src/processing/reprojection.rs
```diff
@@ -657,6 +657,7 @@ where
                 tiling_strat.geo_transform,
                 self.tiling_spec.tile_size_in_pixels,
                 FillerTileCacheExpirationStrategy::DerivedFromSurroundingTiles,
+                FillerTimeBounds::from(query.time_interval),
```
michaelmattig, line 660: maybe add todo to derive the bounds from source stream?

operators/src/adapters/sparse_tiles_fill_adapter.rs
```diff
@@ -193,6 +263,89 @@ impl<T: Pixel> StateContainer<T> {
...
+    }
+
+    fn current_time_fill_to_end_interval(&mut self) {
+        let current_time: TimeInterval = self
+            .current_time
+            .expect("current time must exist for fill to end state.");
+        debug_assert!(current_time.end() < self.time_bounds.end());
+
+        let new_time = if current_time.is_instant() {
+            TimeInterval::new_unchecked(current_time.end() + 1, self.time_bounds.end())
```
michaelmattig, line 325: wouldn't this create a gap in the output time series? if I have a dataset with only one raster image that is valid for an instant. it must fill the everything up to that point with nodata and everything from that point on with nodata. but right now it would create the nodata gap at the end one milli(?)second later and there is now a gap of one millisecond in the time series.

I think we should just disallow instants in time series. We have no use case where we need them anyway right?

jdroenner: yes but it can't be the end of the element before since then it would have the same start. We have this case everywhere where time is modified.. We should remove the posibility of instants. However, this will require us to separate query and element Timeinterval. If we do not change the type it is still possible.

services/src/datasets/external/aruna/mod.rs
jdroenner: yes but cant reference a commit here

operators/src/source/gdal_source/loading_info.rs
jdroenner: [5a4c3e1](https://github.com/geo-engine/geoengine/pull/968/commits/5a4c3e1d58f61a93f49a9e7a07c1e45bafe5befe)

operators/src/processing/reprojection.rs
jdroenner: [1287d44](https://github.com/geo-engine/geoengine/pull/968/commits/1287d44f040444b0007238601f4f0fbf671a4e3f)

services/src/datasets/external/aruna/mod.rs
michaelmattig: if we ask Niki we can insert a better comment here, of even fix it and then merge the PR

jdroenner: [2a0012f](https://github.com/geo-engine/geoengine/pull/968/commits/2a0012f1e7a20e0ecd34dbf85995cb4163403080)

michaelmattig (approved):

PR #970, opened 2024-07-03
FAIR dataset deletion

glombiewski:

---

- Associated user uploads with user datasets.
- Allow new combination of meta data/data deletions.
- Allow for expiration of datasets (deletion after a specified time).


services/src/pro/datasets/storage.rs
```diff
@@ -0,0 +1,197 @@
...
+use geoengine_datatypes::primitives::DateTime;
+use postgres_types::{FromSql, ToSql};
+use serde::{Deserialize, Serialize};
+use tokio_postgres::Transaction;
+use utoipa::ToSchema;
+
+#[derive(Deserialize, Serialize, Debug, Clone, ToSchema)]
+pub struct Expiration {
+    pub deletion_timestamp: Option<DateTime>,
+    pub delete_record: bool,
```
michaelmattig, line 16: does it make sense to delete the dataset record without deleting the data? Maybe make it an enum DeleteRecordAndData, DeleteRecord?
or throw an an error if delete_record ist true but delete_data is false?

services/src/pro/datasets/storage.rs
```diff
@@ -0,0 +1,197 @@
...
+            dataset_id,
+            expiration_change: ExpirationChange::SetExpire(Expiration {
+                deletion_timestamp: None,
+                delete_record: true,
+                delete_data: true,
+            }),
+        }
+    }
+
+    pub fn delete_none(dataset_id: DatasetId) -> Self {
```
michaelmattig, line 60: when would this be used and not expire_none?

services/src/pro/datasets/storage.rs
```diff
@@ -0,0 +1,197 @@
...
+                delete_data: true,
+            }),
+        }
+    }
+
+    pub fn expire_none(dataset_id: DatasetId, timestamp: DateTime) -> Self {
+        ChangeDatasetExpiration {
+            dataset_id,
+            expiration_change: ExpirationChange::SetExpire(Expiration {
+                deletion_timestamp: Some(timestamp),
```
michaelmattig, line 108: does it make sense to expire but not delete anything? What is the timestamp needed for?

services/src/pro/datasets/postgres.rs
```diff
@@ -68,7 +79,7 @@ where
             pos += 1;
             (format!("AND d.tags @> ${pos}::text[]"), filter_tags.clone())
         } else {
-            (String::new(), vec![])
+            ("AND NOT d.tags @> '{deleted}'::text[]".to_string(), vec![])
```
michaelmattig, line 82: if the tags contain "deleted" it still would filter out the deleted datasets right?

services/src/pro/datasets/postgres.rs
```diff
@@ -48,6 +58,7 @@ where
     <<Tls as MakeTlsConnect<Socket>>::TlsConnect as TlsConnect<Socket>>::Future: Send,
 {
     async fn list_datasets(&self, options: DatasetListOptions) -> Result<Vec<DatasetListing>> {
+        self.update_datasets_status().await?;
```
michaelmattig, line 61: maybe add a comment here that we lazily delete uploads on the fly...
and a todo for a regular deletion task to replace this?

operators/src/error.rs
```diff
@@ -500,6 +500,10 @@ pub enum Error {
     Bb8Postgres {
         source: bb8::RunError<tokio_postgres::Error>,
     },
+
+    DatasetDeleted {
```
michaelmattig, line 504: add a snafu displaymessage that the dataset could not be accessed because the data was deleted.

services/src/pro/datasets/postgres.rs
```diff
@@ -861,11 +912,631 @@ impl From<FileUpload> for crate::datasets::upload::FileUpload {
...
+        });
+
+        log::info!(
+            "Adding dataset with name: {:?}, tags: {:?}",
+            name,
+            dataset.tags
+        );
+
+        if let Some(tags) = &dataset.tags {
+            if tags.contains(&"deleted".to_string()) {
```
michaelmattig, line 942: maybe we should have a reserved internal representation that users are not allowed to use themselves?
then again we don't do this for the special "upload" tag either...

services/src/pro/datasets/postgres.rs
```diff
@@ -861,11 +912,631 @@ impl From<FileUpload> for crate::datasets::upload::FileUpload {
...
+        self.update_dataset_status_in_tx(&tx, &expire_dataset.dataset_id)
+            .await?;
+
+        match expire_dataset.expiration_change {
+            ExpirationChange::SetExpire(expiration) => {
+                let num_changes = if let Some(delete_timestamp) = expiration.deletion_timestamp {
+                    let stmt = tx
+                        .prepare("
+                        UPDATE uploaded_user_datasets
+                        SET status = 'Expires', expiration = $2, delete_data = $3, delete_record = $4
```
michaelmattig, line 1050: maybe set the "Expires" as a parameter with a Rust enum variant instead of a string? Or is it fine like this?

services/src/pro/datasets/postgres.rs
```diff
@@ -167,6 +178,7 @@ where
     }
 
     async fn load_dataset(&self, dataset: &DatasetId) -> Result<Dataset> {
+        self.update_dataset_status(dataset).await?;
```
michaelmattig, line 181: can you reuse the connection from the line below?

services/src/error.rs
```diff
@@ -491,6 +491,14 @@ pub enum Error {
         resource_type: String,
         resource_id: String,
     },
+
+    ExpirationTimestampInPast,
+    IllegalExpirationUpdate {
+        reason: String,
+    },
+    IllegalDatasetStatus {
+        status: String,
```
michaelmattig, line 500: add dataset it and snafu display

services/src/pro/datasets/storage.rs
```diff
@@ -0,0 +1,197 @@
...
+    Available,
+    Deleted,
+}
+
+#[derive(Debug, FromSql, ToSql)]
+pub enum InternalUploadedDatasetStatus {
+    Available,
+    Expires,
+    Expired,
+    UpdateExpired,
```
michaelmattig, line 140: what does this state mean?

services/src/pro/util/tests.rs
```diff
@@ -450,6 +450,14 @@ where
     }
 }
 
+#[cfg(test)]
+#[allow(clippy::missing_panics_doc)]
```
michaelmattig, line 454: maybe make it a result instead?

services/src/pro/contexts/migrations/current_schema.sql
```diff
@@ -227,3 +227,25 @@ CREATE TABLE oidc_session_tokens (
...
+    user_id uuid,
+    upload_id uuid,
+    dataset_id uuid,
+    status "InternalUploadedDatasetStatus" NOT NULL,
+    created timestamp with time zone NOT NULL,
+    expiration timestamp with time zone,
+    deleted timestamp with time zone,
+    delete_data boolean NOT NULL,
+    delete_record boolean NOT NULL,
+    PRIMARY KEY (user_id, dataset_id, upload_id)
```
michaelmattig, line 250: why is the user_id part of the primary key? every dataset should only ever point to one upload right?

glombiewski: There is a table "user_uploads" with PRIMARY KEY (user_id, upload_id). So every upload has a user. This table additionally covers every upload has at most one dataset. So we could limit the primary key to that, if you like, but essentially, every upload has exactly one user and at most one dataset.

services/src/pro/datasets/storage.rs
glombiewski: Available: Dataset is available.
Expires: Dataset was set to expire at some point in the future or dataset was "deleted", which translates to expire now.
Expired: Dataset has reached the expiration timestamp, and the deletion action can be triggered now.
UpdateExpired: Dataset was previously Expired or Deleted, and the user has "upgraded" this expiration/deletion. The only way now, after incorporating the other review comments, is that the metadata should also be deleted. Previously this covered other cases. Will be marked as Expired once done.
Deleted: Action after Expired was taken.

services/src/pro/datasets/postgres.rs
```diff
@@ -721,36 +765,46 @@ where
...
+                    return Err(Error::OperationRequiresOwnerPermission);
+                }
+
+                let stmt = tx.prepare("DELETE FROM datasets WHERE id = $1;").await?;
+
+                tx.execute(&stmt, &[&dataset_id]).await?;
+
+                tx.commit().await?;
+
+                return Ok(());
```
glombiewski, line 802: This is a bit odd without the option to delete data, but keep metadata. If the dataset is not a user upload, the engine would do that, otherwise the default now would but a full deletion.

services/src/pro/datasets/storage.rs
glombiewski: I don't think it does, but that's how GeoEngine works, so I originally kept the option for it. I changed it now to the suggested enum (7b4fa9c2b0233db64d3505387f34555d02c2b069 c56db3069fa8e7e6bc42f6a2626134a21a82ce95 3fc317a8357d323dec83b88a67c4419ca9dc507f). An odd part now is delete_dataset from the DatasetStore, I added a new comment there, feel free to comment or resolve if this is not a problem: https://github.com/geo-engine/geoengine/pull/970/files#r1672610833

services/src/pro/datasets/storage.rs
glombiewski: When the deletion of meta data should happen immediately. But since this is no longer an option, this might be resolved.

services/src/pro/datasets/postgres.rs
glombiewski: The intention was:
If tags contains deleted, then it will list deleted datasets. 
If the tags contain anything else, but not the deleted tag, then it will not list deleted datasets (because AND d.tags does not search for it.)
If the tags do not contain anything (case of the marked line), then do not list deleted datasets, hence the not contains. 

I hope it this covers it. The discussed alternative (negative tag filter) was seemingly not desired, so I did not implement it.

services/src/pro/datasets/postgres.rs
glombiewski: I'm not sure I understand.

The requirement to update the datasets status is that the expiration date might have been reached. So before any operation, this needs to be checked. In that case, listing does have to do with dataset status.

I wrapped the functions into a lazy_dataset_updates and made a comment above it. Hopefully, this is what you wanted. 2865e61058c4a931cf41c73e497f2e812bc78b17

operators/src/error.rs
glombiewski: 265eba09b4725423b2064fad2ad936bbf66f69c0

services/src/error.rs
glombiewski: 265eba09b4725423b2064fad2ad936bbf66f69c0

services/src/pro/datasets/postgres.rs
glombiewski: 2865e61058c4a931cf41c73e497f2e812bc78b17

services/src/pro/datasets/postgres.rs
glombiewski: Not sure if this is actually more readable? Or did you have a format! with an inplace Enum in mind (if that works)?  7b4fa9c2b0233db64d3505387f34555d02c2b069

services/src/pro/datasets/postgres.rs
glombiewski: I added Reserved Tags enum and have log a warning at the handler level. Could throw an error there if desired. c287c3459b425802b5a54c2060145cbb3a955115

services/src/contexts/migrations/migration_0010_delete_uploaded_datasets.rs
```diff
@@ -0,0 +1,24 @@
+use async_trait::async_trait;
+use tokio_postgres::Transaction;
+
+use crate::error::Result;
+
+use super::database_migration::{DatabaseVersion, Migration};
+
+/// This migration adds new delete options for uploaded user datasets
+pub struct Migration0011DeleteUploadedDatasets;
```
michaelmattig, line 9: maybe be a better name would be fair_data_deletion? otherweise one might think the migration would delete datasets...

services/src/pro/datasets/postgres.rs
michaelmattig: what exactly is werid about it?

services/src/pro/contexts/migrations/current_schema.sql
michaelmattig: ok so leave it in.

Another point: should the expiration be a separate table? I'd figure it to be NULL for almost every dataset because deletion is rare?

services/src/pro/datasets/postgres.rs
```diff
@@ -745,36 +794,46 @@ where
...
-            )
-            .await?;
-
-        let rows = tx
-            .query(&stmt, &[&dataset_id, &self.session.user.id])
-            .await?;
-
-        if rows.is_empty() {
-            return Err(Error::OperationRequiresOwnerPermission);
+        let _uploaded = self.uploaded_dataset_status_in_tx(&dataset_id, &tx).await;
```
michaelmattig, line 797: why do you ignore the result?

services/src/pro/datasets/storage.rs
michaelmattig: maybe add it as a comment to the code

services/src/pro/datasets/postgres.rs
```diff
@@ -745,36 +794,46 @@ where
...
+                return Ok(());
+            }
         }
 
-        let stmt = tx.prepare("DELETE FROM datasets WHERE id = $1;").await?;
-
-        tx.execute(&stmt, &[&dataset_id]).await?;
-
-        tx.commit().await?;
+        self.expire_uploaded_dataset(ChangeDatasetExpiration::delete_full(dataset_id))
```
michaelmattig, line 835: is every delete an expiration now?

services/src/pro/datasets/postgres.rs
glombiewski: The previous way to delete data was to delete meta data, but keep data. That still exists for non-user uploads. But user uploads do not have that option.

services/src/pro/datasets/postgres.rs
glombiewski: If it is a naming thing, I'm...open either way.

From a technical standpoint, the current implementation interprets _delete_ as _expire now_, which means there is only one internal logic (i.e., set of queries) for this process. I could also change that.

PR #977, opened 2024-08-15
db for ml models (wip)

michaelmattig:

---

<TEXT>


michaelmattig:
[ML-Model-DB](https://trello.com/c/UReOmLzH/1707-ml-model-db)

datatypes/src/machine_learning.rs
```diff
@@ -0,0 +1,125 @@
...
+use serde::{de::Visitor, Deserialize, Serialize};
+
+use crate::{
+    dataset::{is_invalid_name_char, SYSTEM_NAMESPACE},
+    raster::RasterDataType,
+};
+
+const NAME_DELIMITER: char = ':';
+
+// TODO: custom serialization as ns:name
```
jdroenner, line 12: you did that, right?

operators/src/pro/machine_learning/metadata_from_file.rs
```diff
@@ -0,0 +1,93 @@
...
+use geoengine_datatypes::{machine_learning::MlModelMetadata, raster::RasterDataType};
+use snafu::{ensure, ResultExt};
+
+use crate::pro::machine_learning::error::{
+    InvalidInputDimensions, InvalidOutputDimensions, MultipleInputsNotSupported, Ort,
+};
+
+use super::MachineLearningError;
+
+// TODO: extract metadata during model import and load it from database here instead of accessing the model file here
```
jdroenner, line 12: do you still intend to do this?

services/src/machine_learning/name.rs
```diff
@@ -0,0 +1,154 @@
...
+            namespace,
+            name: name.into(),
+        }
+    }
+}
+
+impl From<MlModelName> for geoengine_datatypes::machine_learning::MlModelName {
+    fn from(name: MlModelName) -> Self {
+        Self {
+            namespace: name.namespace.clone(),
```
jdroenner, line 37: why is that clone needed?

jdroenner (commented):
how did you solve the type conversion of the model output?

datatypes/src/machine_learning.rs
michaelmattig: yes, I removed the comment 07374b850dd9a0d5638233a4a1df5deabb1dceea

operators/src/pro/machine_learning/metadata_from_file.rs
michaelmattig: I removed the comment, as it originally targeted loading the model data in the operator, which now does use the database. 07374b850dd9a0d5638233a4a1df5deabb1dceea

We could still think about adding an API handler that autogenerates the MlMetadata, like we do for the OgrDataset. Currently, the user has to specify this manually when adding the model. But right now I think it only saves very little and adds another potential source of failure, if we use it by default.

services/src/machine_learning/name.rs
michaelmattig: I removed it 07374b850dd9a0d5638233a4a1df5deabb1dceea

operators/src/pro/machine_learning/metadata_from_file.rs
jdroenner: yes i guess it is file like this

jdroenner (approved):

PR #986, opened 2024-09-12
Gfbio collections adjustments

dbrandenstein:

---

<TEXT>


services/src/datasets/external/gfbio_collections.rs
```diff
@@ -186,14 +183,14 @@ fn gfbio_dataset_identifier_to_dataset_unit(
     dataset_identifier: &str,
 ) -> Result<(String, Option<String>)> {
     // urn:gfbio.org:abcd:{dataset}:{unit} (unit is optional)
-    let id_parts: Vec<_> = dataset_identifier.split(':').collect();
+    let id_parts: Vec<_> = dataset_identifier.splitn(5, ':').collect();
 
     Ok(match id_parts.as_slice() {
-        [_, _, _, dataset, unit] => (
+        ["urn", "gfbio.org", "abcd", dataset, unit] => (
```
michaelmattig, line 189: if there are 4 colons in the identifier: how do you know whether the last part is a unit or part of the dataset name?

PR #988, opened 2024-10-15
ajudst ogr time type to json format

michaelmattig:

---

<TEXT>

### DIFF
--- a/services/src/api/handlers/datasets.rs
+++ b/services/src/api/handlers/datasets.rs
@@ -16,7 +16,7 @@ use crate::{
         },
         DatasetName,
     },
-    error::{self, Result},
+    error::{self, Error, Result},
     projects::Symbology,
     util::{
         config::{get_config_element, Data},
@@ -47,12 +47,14 @@ use geoengine_operators::{
         raster_descriptor_from_dataset,
     },
 };
+use serde::{Deserialize, Serialize};
 use snafu::ResultExt;
 use std::{
     collections::HashMap,
     convert::{TryFrom, TryInto},
     path::Path,
 };
+use utoipa::{ToResponse, ToSchema};
 
 pub(crate) fn init_dataset_routes<C>(cfg: &mut web::ServiceConfig)
 where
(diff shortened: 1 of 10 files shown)


michaelmattig:
[UI für OGR Datasets](https://trello.com/c/xBzHbJxU/1728-ui-f%C3%BCr-ogr-datasets)

services/src/api/model/datatypes.rs
```diff
@@ -1719,13 +1720,122 @@ impl From<RasterPropertiesEntryType> for geoengine_datatypes::raster::RasterProp
...
     has_time: bool,
 }
 
+enum FormatStrLoopState {
+    Normal,
+    Percent(String),
+}
+
+impl DateTimeParseFormat {
+    pub fn custom(fmt: String) -> Self {
```
ChristianBeilschmidt, line 1736: This looks rather complicated. Maybe test it?

services/src/api/handlers/datasets.rs
```diff
@@ -1407,6 +1413,63 @@ pub async fn delete_dataset_handler<C: ApplicationContext>(
...
+    ),
+    params(
+        ("volume_name" = VolumeName, description = "Volume name"),
+        ("file_name" = String, description = "File name")
+    ),
+    security(
+        ("session_token" = [])
+    )
+)]
+pub async fn list_volume_file_layers_handler<C: ApplicationContext>(
```
ChristianBeilschmidt, line 1438: Add one success case test?

ChristianBeilschmidt (changes requested):

services/src/api/handlers/datasets.rs
michaelmattig: c2418df70afee617ce159f7f6dcca966efacbbb8

services/src/api/model/datatypes.rs
ChristianBeilschmidt: Yeah, back then it was tested and then I copied it doesn't sound good.
Rather copy the test or do the conversion.

michaelmattig: 5b5c0e77ea24da8f4adb41af2be807332aa341ef

ChristianBeilschmidt (approved):

PR #989, opened 2024-10-23
SparseTilesFillAdapter: add requested bounds, fill only if needed

jdroenner:

---

- Added requested time bounds to SparseTilesFillAdapter
- Adjust adapter to only fill in requested time ranges
- StacSentinel2Provider. clip times on full days to avoid wrong aggregations

### DIFF
--- a/operators/src/adapters/sparse_tiles_fill_adapter.rs
+++ b/operators/src/adapters/sparse_tiles_fill_adapter.rs
@@ -98,7 +98,8 @@ struct StateContainer<T> {
     global_geo_transform: GeoTransform,
     state: State,
     cache_hint: FillerTileCacheHintProvider,
-    time_bounds: FillerTimeBounds,
+    requested_time_bounds: TimeInterval,
+    data_time_bounds: FillerTimeBounds,
 }
 
 struct GridIdxAndBand {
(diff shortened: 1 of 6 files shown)


services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
```diff
@@ -466,6 +466,35 @@ impl SentinelS2L2aCogsMetaData {
                 query.time_interval.end() + query_end_buffer
             };
 
+            // clip end to the same day to avoid filling wrong days
```
michaelmattig, line 469: not wrong days, but partial days right? Please explain it some more. Does it have to do with our aggregation granularity? Otherwise I am not sure why it would matter.

operators/src/adapters/sparse_tiles_fill_adapter.rs
```diff
@@ -361,15 +371,26 @@ where
...
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         stream: S,
         tile_grid_bounds: GridBoundingBox2D,
         num_bands: u32,
         global_geo_transform: GeoTransform,
         tile_shape: GridShape2D,
         cache_expiration: FillerTileCacheExpirationStrategy, // Specifies the cache expiration for the produced filler tiles. Set this to unlimited if the filler tiles will always be empty
-        time_bounds: FillerTimeBounds,
+        requested_time_bounds: TimeInterval,
```
michaelmattig, line 382: please add a test that shows that setting the req. time bounds actually changes the result.

services/src/pro/datasets/external/sentinel_s2_l2a_cogs.rs
jdroenner: https://github.com/geo-engine/geoengine/pull/989/commits/02c4d8e7f32d2916e6c038d9e60b1751752d0fd9

operators/src/adapters/sparse_tiles_fill_adapter.rs
jdroenner: https://github.com/geo-engine/geoengine/pull/989/commits/b23c163bf393f3a6e68b45c62bde13ee4d290395

PR #998, opened 2024-11-20
Quota and Data usage Logging

michaelmattig:

---

<TEXT>


michaelmattig:
[Quota Logging Mechanismus](https://trello.com/c/ewnNwB6c/1789-quota-logging-mechanismus)

datatypes/src/primitives/datetime.rs
```diff
@@ -323,6 +323,12 @@ impl From<chrono::DateTime<chrono::Utc>> for DateTime {
     }
 }
 
+impl From<DateTime> for chrono::DateTime<chrono::Utc> {
+    fn from(datetime: DateTime) -> Self {
+        datetime.datetime
+    }
+}
+
```
ChristianBeilschmidt, line 331: CI fails

ChristianBeilschmidt, line 331: test coverage is below 70%

operators/src/adapters/stream_statistics_adapter.rs
```diff
@@ -16,6 +16,8 @@ pub struct StreamStatisticsAdapter<S> {
     span: Span,
     quota: QuotaTracking,
     path: WorkflowOperatorPath,
+    operator_name: &'static str,
+    data: Option<String>,
```
ChristianBeilschmidt, line 20: what is data?

operators/src/engine/operator.rs
```diff
@@ -149,6 +149,17 @@ pub trait InitializedRasterOperator: Send + Sync {
...
     fn canonic_name(&self) -> CanonicOperatorName;
+
+    /// Get the unique name of the operator
+    fn name(&self) -> &'static str;
+
+    // Get the path of the operator in the workflow
+    fn path(&self) -> WorkflowOperatorPath;
+
+    /// Return the name of the data loaded by the operator (if any)
+    fn data(&self) -> Option<String> {
```
ChristianBeilschmidt, line 160: `Option<&String>` or `Option<Arc<str>>` for cheap copy?

operators/src/machine_learning/onnx.rs
```diff
@@ -124,6 +130,14 @@ impl InitializedRasterOperator for InitializedOnnx {
...
         self.name.clone()
     }
+
+    fn name(&self) -> &'static str {
+        Onnx::TYPE_NAME
+    }
+
+    fn path(&self) -> WorkflowOperatorPath {
+        self.path.clone()
+    }
```
ChristianBeilschmidt, line 140: this whole thing looks like it should be covered by a macro

services/src/api/handlers/wfs.rs
```diff
@@ -499,14 +498,15 @@ async fn wfs_feature_handler<C: ApplicationContext>(
...
+
+        let workflow_operator_path_root = WorkflowOperatorPath::initialize_root();
 
-        // create the inititalized operator directly, to avoid re-initializing everything
-        let ivp = InitializedVectorReprojection::try_new_with_input(
-            CanonicOperatorName::from(&reprojected_workflow),
-            reprojection_params,
-            initialized,
-        )?;
+        // TODO: avoid re-initialization and re-use unprojected workflow. However, this requires updating all operator paths
```
ChristianBeilschmidt, line 506: I don't get it.

services/src/contexts/mod.rs
```diff
@@ -80,7 +80,8 @@ pub trait SessionContext: 'static + Send + Sync + Clone {
     fn tasks(&self) -> Self::TaskManager;
 
     /// Create a new query context for executing queries on processors
-    fn query_context(&self) -> Result<Self::QueryContext>;
+    // TODO: assign computation id inside SessionContext or let it be provided from outside?
```
ChristianBeilschmidt, line 83: I guess this is ok?

services/src/error.rs
```diff
@@ -133,6 +133,7 @@ pub enum Error {
 
     NoWorkflowForGivenId,
 
+    #[snafu(display("Database error: {source}"))]
```
ChristianBeilschmidt, line 136: do we really output the db internas to the user?

services/src/pro/api/handlers/users.rs
```diff
@@ -338,6 +357,170 @@ where
...
+
+    let db = app_ctx.session_context(session).db();
+    let computations_quota = db
+        .quota_used_by_computations(params.offset, params.limit)
+        .await?;
+
+    Ok(web::Json(computations_quota))
+}
+
+/// Retrieves the quota used by computations
```
ChristianBeilschmidt, line 399: Should be different than `computations_quota_handler`

services/src/pro/quota/mod.rs
```diff
@@ -170,6 +182,41 @@ pub fn initialize_quota_tracking<U: UserDb + 'static>(
     QuotaTrackingFactory::new(quota_sender)
 }
 
+#[derive(Debug, Serialize, Deserialize, ToSchema, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorQuota {
+    pub operator_name: String,
+    pub operator_path: String,
```
ChristianBeilschmidt, line 189: is this cloned a lot? maybe `Arc<str>`

services/src/pro/users/userdb.rs
```diff
@@ -121,6 +125,56 @@ pub trait UserDb: Send + Sync {
         quota_used_updates: I,
     ) -> Result<()>;
 
+    /// Log quota usage (computation units)
+    ///
+    /// # Errors
+    ///
+    /// This call fails if database cannot be accessed
+    ///
+    // TODO: move this method to some AdminDb?
```
ChristianBeilschmidt, line 134: why?

ChristianBeilschmidt (changes requested):

datatypes/src/primitives/datetime.rs
michaelmattig: 70dc35aa72786046519038f50f349871299d9ad5

michaelmattig: I cannot really comprehend how this is calculated if I look at the report. I mean I touched all of the operators so maybe that leads to this measurement? But it also reports `Measurement`s to be untested now, and I didn't touch any of that.  https://coveralls.io/builds/71398158/source?filename=services%2Fsrc%2Fapi%2Fmodel%2Fdatatypes.rs#L765

If you want anything specific to be tested just let me know.

operators/src/engine/operator.rs
michaelmattig: doing `Option<&String>` won't help much because we need to send the result down a channel, which requires owning it.

`Option<Arc<str>>`  would work I guess. But is the synchronization overhead really better than the cloning? The String will always be small.

services/src/api/handlers/wfs.rs
michaelmattig: in order to check whether we need to inject a reprojection, we first need to initialize the original workflow. Then we can check the result projection. Previously, we then just wrapped the initialized workflow with an initialized reprojection. IMHO this is wrong because initialization propagates the workflowpath down the children and append a new segment for each level. so we can't re-use an already initialized workflow.

operators/src/machine_learning/onnx.rs
michaelmattig: I would not know how to achieve this in a way that actually saves code... We have no way to know which Operator an InitializedOperator belongs to so we can't access its name. Also each InitializedOperator is responsible for storing it's path, so we would only be able to save 3 lines in each operator but would not really gain anything and make it more unreadable.

I guess there would be a better way with an associated type and a InitializedOperatorBase that handles all the common stuff like saving the result descriptor and the name and the path but that refactoring would not gain us much I think and would cost several hours.

services/src/error.rs
michaelmattig: was only for debugging c7e5c8ae4b1a41260f1e2d545f57f2baf849fac9

services/src/pro/quota/mod.rs
michaelmattig: no, it's used for API output and filled from the database

services/src/pro/users/userdb.rs
michaelmattig: the `UserDb` was initially designed to provide users access to their own metadata, like name, session, quota, etc. Now we also use it for stuff like updating the quota from the QuotaTracking module which is something only the system/admin should do. So all the methods that don't really fit in the UserDb have this comment...

operators/src/adapters/stream_statistics_adapter.rs
ChristianBeilschmidt: Could you add rustdoc?

operators/src/machine_learning/onnx.rs
ChristianBeilschmidt: You did 25x basically the same stuff. This is why it becomes Hard to write operators. Can be a different PR, tho.

services/src/api/handlers/wfs.rs
ChristianBeilschmidt: Sollte nicht nur der Elternknoten länger werden? Die source würde ja Stufe 1 bleiben, aber oben kommt etwas dazu.

services/src/contexts/mod.rs
ChristianBeilschmidt: Ich weiß nicht, Dependency Injection klingt schon ganz gut, vielleicht sollte man eine Factory reingeben statt einer Workflow ID

ChristianBeilschmidt (approved):

ChristianBeilschmidt (approved):

PR #1000, opened 2024-11-26
feat: Add ml model input and output shape to allow models run on entire tiles

jdroenner:

---

<TEXT>


operators/src/machine_learning/metadata_from_file.rs
```diff
@@ -34,44 +35,58 @@ pub fn load_model_metadata(path: &Path) -> Result<MlModelMetadata, MachineLearni
     };
 
     // Input dimensions must be [-1, b] to accept a table of (arbitrarily many) single pixel features (rows) with `b` bands (columns)
-    ensure!(
-        input_dimensions.len() == 2 && input_dimensions[0] == -1 && input_dimensions[1] > 0,
-        InvalidInputDimensions {
-            dimensions: input_dimensions.clone()
-        }
-    );
+    let input_shape = try_dimensions_to_tensor_shape(input_dimensions)?;
```
michaelmattig, line 38: the comment above this line is no longer valid, right?

operators/src/machine_learning/metadata_from_file.rs
```diff
@@ -34,44 +35,58 @@ pub fn load_model_metadata(path: &Path) -> Result<MlModelMetadata, MachineLearni
...
+    } else if dimensions.len() == 2 && dimensions[0] == -1 && dimensions[1] > 0 {
+        Ok(TensorShape3D::new_y_x_bands(1, 1, dimensions[1] as u32))
+    } else if dimensions.len() == 4
+        && dimensions[0] == -1
+        && dimensions[1] > 0
+        && dimensions[2] > 0
+        && dimensions[3] > 0
+    {
+        Ok(TensorShape3D::new_y_x_bands(
+            dimensions[1] as u32, // TODO: figure out how the axis in the dimensions are ordered!
```
michaelmattig, line 79: shouldn't we figure this out before merging?

operators/src/machine_learning/mod.rs
```diff
@@ -18,12 +21,23 @@ pub enum MachineLearningError {
...
+        "Onnx model must have one dimensional output or match the tiling tile size. Found {:?} and {:?}.",
+        tensor_shape, tiling_shape
     ))]
-    InvalidOutputDimensions { dimensions: Vec<i64> },
+    InvalidInputShape {
+        tensor_shape: TensorShape3D,
+        tiling_shape: GridShape2D,
+    },
+    #[snafu(display(
+        "Onnx model must have one dimensional output or match the tiling tile size. Found {:?} and {:?}.",
```
michaelmattig, line 34: just a question: wouldn't it also be relatively easy to support model outputs that are multiples/divided by two?

operators/src/machine_learning/onnx.rs
```diff
@@ -54,11 +56,37 @@ impl RasterOperator for Onnx {
 
         let model_metadata = context.ml_model_metadata(&self.params.model).await?;
 
+        let tiling_shape = context.tiling_specification().tile_size_in_pixels;
+
+        // check that we can use the model input shape with the operator
+        ensure!(
```
michaelmattig, line 62: isn't this checked by try_dimensions_to_tensorshape?

operators/src/machine_learning/onnx.rs
```diff
@@ -258,18 +295,41 @@ where
...
-                let samples = Array2::from_shape_vec((rows, cols), pixels).expect(
-                    "Array2 should be valid because it is created from a Vec with the correct size",
-                );
-
-                let input_name = &session.inputs[0].name;
-
-                let outputs = session
-                    .run(ort::inputs![input_name => samples].context(Ort)?)
-                    .context(Ort)?;
+                let outputs = if self.model_metadata.input_is_single_pixel() {
```
michaelmattig, line 299: could/should we have an enum of shapes and match here instead of if else if else?

operators/src/machine_learning/mod.rs
jdroenner: yes but then the input might get tricky. There is still a lot to do here but maybe not in this PR

operators/src/machine_learning/onnx.rs
jdroenner: no this check is that the tensor is compatible to the tiling_shape the engine uses.

operators/src/machine_learning/onnx.rs
jdroenner: its only two cases and a default that results in an error. I don't know how an enum will make it less complex.

operators/src/machine_learning/onnx.rs
```diff
@@ -258,18 +295,41 @@ where
...
+                        .context(Ort)?;
+                    Ok(out)
+                } else if self.model_metadata.input_shape.yx_matches_tile_shape(&tile_shape){
+                    let samples = Array4::from_shape_vec((1, height, width, num_bands), pixels).expect( // y,x, attributes
+                        "Array2 should be valid because it is created from a Vec with the correct size",
+                    );
+
+                    let input_name = &session.inputs[0].name;
+
+                    let out = session
```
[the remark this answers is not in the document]
jdroenner: We still assume that we only want the first output. I agree that we could add a selection which outputs to put into bands but thats not for this PR.

operators/src/machine_learning/onnx.rs
jdroenner: i moved all of the testing to onnx_utils.rs we need to check the number of input bands in the operator because we need the ResultDescriptor of the specific workflow for that. (check_model_input_features)

operators/src/machine_learning/onnx.rs
```diff
@@ -187,22 +194,10 @@ where
...
-                );
-            })?;
-
-        tracing::debug!(
-            "Created ONNX session for {:?}",
-            &self.model_path.file_name()
-        );
+        let session = load_onnx_model_from_metadata(&self.model_metadata)?;
+
+        // TODO: check @ initialize?
```
michaelmattig, line 199: or: check when adding the model to Geo Engine?

operators/src/machine_learning/onnx.rs
```diff
@@ -256,18 +252,41 @@ where
...
-                    "Array2 should be valid because it is created from a Vec with the correct size",
-                );
-
-                let input_name = &session.inputs[0].name;
 
-                let outputs = session
-                    .run(ort::inputs![input_name => samples].context(Ort)?)
-                    .context(Ort)?;
+                let outputs = if self.model_metadata.input_is_single_pixel() {
+                    let rows = width * height;
```
michaelmattig, line 257: use pixels.len()?

operators/src/machine_learning/onnx.rs
```diff
@@ -256,18 +252,41 @@ where
...
+
+                    let input_name = &session.inputs[0].name;
+
+                    let out = session
+                        .run(ort::inputs![input_name => samples].context(Ort)?)
+                        .context(Ort)?;
+                    Ok(out)
+                } else if self.model_metadata.input_shape.yx_matches_tile_shape(&tile_shape){
+                    let samples = Array4::from_shape_vec((1, height, width, num_bands), pixels).expect( // y,x, attributes
+                        "Array2 should be valid because it is created from a Vec with the correct size",
```
michaelmattig, line 272: Array4?

operators/src/machine_learning/onnx.rs
```diff
@@ -347,7 +374,62 @@ mod tests {
     };
     use ndarray::{Array1, Array2, arr2, array};
 
-    use super::*;
+    fn generate_model_metadata_from_onnx(
```
michaelmattig, line 377: shouldn't this method itself be tested to generate the correct metadata for the individual onnx files?

operators/src/machine_learning/onnx_util.rs
```diff
@@ -0,0 +1,318 @@
...
+        UnsupportedInOutMapping {
+            in_shape: model_metadata.input_shape,
+            out_shape: model_metadata.output_shape
+        }
+    );
+
+    Ok(())
+}
+
+pub fn try_onnx_tensor_to_ml_tensorshape_3d(
```
michaelmattig, line 103: maybe test this method

services/src/api/handlers/machine_learning.rs
```diff
@@ -60,12 +66,27 @@ pub(crate) async fn add_ml_model<C: ApplicationContext>(
...
+
+    // This call also checks that the file is available!
+    let ml_model_metadata = model.metadata_for_operator()?;
+    // Check that the in/out shapes are ok
+    check_model_shape(
+        &ml_model_metadata,
+        exe_context.tiling_specification().tile_size_in_pixels,
+    )?;
+    // initialize model
+    // TODO: re-use session accross queries?
```
michaelmattig, line 84: there is no query happening? coment copy pasted?

operators/src/machine_learning/onnx.rs
```diff
@@ -256,18 +252,41 @@ where
...
+                    let out = session
+                        .run(ort::inputs![input_name => samples].context(Ort)?)
+                        .context(Ort)?;
+                    Ok(out)
+                } else if self.model_metadata.input_shape.yx_matches_tile_shape(&tile_shape){
+                    let samples = Array4::from_shape_vec((1, height, width, num_bands), pixels).expect( // y,x, attributes
+                        "Array2 should be valid because it is created from a Vec with the correct size",
+                    );
+
+                    let input_name = &session.inputs[0].name;
```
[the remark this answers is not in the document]
jdroenner: samples has a different type so i can not move it into one case. I can move it into a method but this will not be less code...

operators/src/machine_learning/onnx.rs
jdroenner: [b1facd8](https://github.com/geo-engine/geoengine/pull/1000/commits/b1facd8ff3be816d9088e99465db1d043d4d9230)

operators/src/machine_learning/onnx.rs
jdroenner: [b1facd8](https://github.com/geo-engine/geoengine/pull/1000/commits/b1facd8ff3be816d9088e99465db1d043d4d9230)

services/src/api/handlers/machine_learning.rs
jdroenner: [e936c08](https://github.com/geo-engine/geoengine/pull/1000/commits/e936c087dee32a3a244f7a2dc6972fda4787f8d6)

operators/src/machine_learning/onnx_util.rs
jdroenner: [7707c68](https://github.com/geo-engine/geoengine/pull/1000/commits/7707c68b2ee460f3454cd961e966523434cfadc5)

PR #1008, opened 2025-01-15
user ctx in ge_test

ChristianBeilschmidt:

---

<TEXT>


services/src/contexts/migrations/database_migration.rs
```diff
@@ -315,53 +311,4 @@ mod tests {
 
         Ok(())
     }
```
michaelmattig, line 313: what happened to the test?

services/src/layers/add_from_directory.rs
```diff
@@ -92,6 +117,19 @@ pub async fn add_layer_collections_from_directory<L: LayerDb>(db: &mut L, file_p
         )
         .await?;
 
+        // share with users
+        debug!("sharing collection");
```
michaelmattig, line 121: add more infos (which collection) or remove debug

services/src/contexts/migrations/database_migration.rs
ChristianBeilschmidt: There was one here for non-pro and one in pro/contexts/… for pro. Since the former does not exist anymore, so doesn't the test.

services/src/layers/add_from_directory.rs
ChristianBeilschmidt: I just moved the code. I don't know who put it here and why.

ChristianBeilschmidt: [bcb0f9d](https://github.com/geo-engine/geoengine/pull/1008/commits/bcb0f9d958e515b6137e5ecfb81547892e713ba6)

michaelmattig (approved):

PR #1048, opened 2025-04-17
1.86

ChristianBeilschmidt:

---

<TEXT>


.github/workflows/ci.yml
```diff
@@ -77,12 +77,7 @@ jobs:
       - name: Clippy
         run: cargo clippy --all-targets --locked -- -D warnings
       - name: SQLFluff
-        run: pipx run sqlfluff==3.3.0 lint
-      - name: Verifying expression dependencies workspace
-        run: |
-          rustup toolchain install nightly
-          chmod +x ./.scripts/check-expression-deps.rs
-          ./.scripts/check-expression-deps.rs
```
michaelmattig, line 85: delete the script then (or share code between the new test and the script)

operators/src/source/ogr_source/mod.rs
```diff
@@ -1422,8 +1432,9 @@ impl TryFromOgrGeometry for MultiPoint {
 impl TryFromOgrGeometry for MultiLineString {
     fn try_from(geometry: Result<&gdal::vector::Geometry>) -> Result<Self> {
         fn coordinates(geometry: &gdal::vector::Geometry) -> Vec<Coordinate2D> {
-            geometry
-                .get_point_vec()
+            let mut point_vec = Vec::new();
```
michaelmattig, line 1435: set capacity?

operators/src/source/ogr_source/mod.rs
```diff
@@ -1297,7 +1307,7 @@ where
         builder.push_time_interval(time_interval);
 
         for (column, data_type) in data_types {
-            let field = feature.field(column);
+            let field = feature.field_index(column).and_then(|i| feature.field(i));
```
michaelmattig, line 1310: maybe we should look the index up only once (same for the other occurrences)? I know before the change to the gdal API the index was also looked up everytime but I think we weren't aware of that when we wrote this.

.github/workflows/ci.yml
ChristianBeilschmidt: [ce97279](https://github.com/geo-engine/geoengine/pull/1048/commits/ce972795010c5b783bb6e508061ce0a588c4d93d)

operators/src/source/ogr_source/mod.rs
ChristianBeilschmidt: [da49d4f](https://github.com/geo-engine/geoengine/pull/1048/commits/da49d4fb4bcefffba577b8bd603c1851fa96dd2c)

PR #1071, opened 2025-08-12
feat(datatypes): reproject outside area of use

dbrandenstein:
Combines clipped and non-clipped reprojection of border sample points. The largest possible bbox (a combination of both or the one which is non-empty) is taken as result.

---

<TEXT>


dbrandenstein:
[feat(datatypes): reproject outside area of use](https://trello.com/c/wefRZdXG/1868-featdatatypes-reproject-outside-area-of-use)

PR #1080, opened 2025-09-26
fix(services): permission queries

dbrandenstein:

---

<TEXT>


dbrandenstein:
[Query Fixes](https://trello.com/c/ipJFuUdx/1902-query-fixes)

services/src/contexts/migrations/current_schema.sql
```diff
@@ -1168,60 +1168,76 @@ CREATE UNIQUE INDEX ON permissions (
...
+    )
+GROUP BY r.user_id, p.layer_collection_id;
 
 CREATE VIEW user_permitted_layers
 AS
 SELECT
     r.user_id,
     p.layer_id,
-    p.permission
+    MAX(p.permission) AS max_permission
```
ChristianBeilschmidt: This works because enums are ints and read=0, owner=1?

services/src/datasets/postgres.rs
```diff
@@ -792,7 +792,7 @@ where
             user_permitted_datasets p JOIN datasets d 
                 ON (p.dataset_id = d.id)
         WHERE 
-            d.id = $1 AND p.user_id = $2 AND p.permission = 'Owner';",
+            d.id = $1 AND p.user_id = $2 AND p.max_permission = 'Owner';",
```
ChristianBeilschmidt: Please add a test that has multiple rights for a user.

services/src/layers/postgres_layer_db.rs
```diff
@@ -884,7 +884,7 @@ where
                     user_permitted_providers p
                     JOIN layer_providers l ON (p.provider_id = l.id)
                 WHERE
-                    p.user_id = $3 AND priority > -1000
+                    p.user_id = $3
```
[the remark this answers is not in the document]
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1080/commits/0b9bc5df62c181c623e707452ca08dff59bd6d7a

services/src/contexts/migrations/current_schema.sql
dbrandenstein: Yes, exactly

PR #1166, opened 2026-04-16
refactor: add ui to monorepo

ChristianBeilschmidt:
DO NOT SQUASH THIS


.github/workflows/ci.yml
```diff
@@ -314,3 +314,74 @@ jobs:
         run: just api-clients build
       - name: Check for changes in git repo
         run: just check-no-changes-in-git-repo
+
```
michaelmattig, line 317: unrelated to this line:

in api-clients/package.json there is still @geoengine/openapi-client instead of api-client

.github/workflows/ci.yml
```diff
@@ -314,3 +314,74 @@ jobs:
...
+  ui-build:
+    name: "ui / build"
+    runs-on: ubuntu-24.04
+    container: quay.io/geoengine/devcontainer:latest
+
```
michaelmattig, line 322: unrelated to this line:

what is this change:

> --- geoengine-ui/projects/core/src/lib/util/pipes/css-string-to-rgba.pipe.ts    2025-11-12 09:50:15.316960975 +0100
+++ geoengine-review/ui/projects/core/src/lib/util/pipes/css-string-to-rgba.pipe.ts     2026-04-16 18:30:09.465292861 +0200
@@ -10,10 +10,7 @@
             return TRANSPARENT.rgbaTuple();
         }
 
-        const rgba =
-            /^rgba\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+.*\d*)\s*\)$/i.exec(rgbaCssString) ??
-            /^rgba\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i.exec(rgbaCssString) ??
-            /^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i.exec(rgbaCssString);
+        const rgba = /^rgba?\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})(?:\s*,\s*(0?\.?\d+|1(?:\.0+)?))?\s*\)$/i.exec(rgbaCssString);
 
         if (rgba) {
             return [parseInt(rgba[1], 10), parseInt(rgba[2], 10), parseInt(rgba[3], 10), rgba[4] === undefined ? 1 : parseFloat(rgba[4])];
@@ -23,7 +20,7 @@
         if (threeDigitMatch) {
             const threeDigit = threeDigitMatch[1];
             // in three-character format, each value is multiplied by 0x11 to give an
-            // even scale fromRgbaLike 0x00 to 0xff
+            // even scale from 0x00 to 0xff
             return [
                 parseInt(threeDigit.charAt(0), 16) * 0x11,
                 parseInt(threeDigit.charAt(1), 16) * 0x11,

justfile
```diff
@@ -2,18 +2,19 @@ mod api-clients
...
 
 # Install dependencies for all submodules.
 [group("install")]
-install: backend::install python::install
+install: backend::install python::install ui::install
 
 # Call lint for all submodules.
 [group("lint")]
-lint: lint-justfiles lint-openapi-spec backend::lint python::lint www::lint
+lint: lint-justfiles lint-openapi-spec backend::lint python::lint ui::lint www::lint
```
michaelmattig, line 17: would be nice to have a just ci command that runs all the commands the ci does. could be a new task

.github/workflows/ci.yml
ChristianBeilschmidt: CodeQL said the old one was slowing the Browser in harmful inputs

justfile
ChristianBeilschmidt: Yes, lets add this to success criteria

.github/workflows/ci.yml
ChristianBeilschmidt: And it was Kind of stupid, too. Why 3 regex?

.github/workflows/ci.yml
```diff
@@ -314,3 +314,74 @@ jobs:
...
+        app:
+          - gis
+          - manager
+          - dashboards/gfbio
+          - dashboards/data-atlas
+          - dashboards/nfdi-portal-demo
+          - dashboards/ebv-analyzer
+          - dashboards/ecometrics
+          - dashboards/esg-indicator-service
+
```
michaelmattig, line 337: unrelated to this line: if I regenerate the api-clients just api-clients::build, some files change in api-client:

expression.rs: comment changes
Expression.d.ts :```Rust -> ```rust,ignore

also www/src/content/docs/docs/python creates some new files

- I think you have to commit these files
- why did the ci not catch this? it checks for idempotency
- why is the folder path even doc/docs? one docs too much

michaelmattig (approved):

PR #1227, opened 2026-07-30
feat: external data path

michaelmattig:

- MultiBandGdalSource always reads data relative to Volume (or Upload) base path
- Data from DataConnects cannot always refer to Volumes, because the base paths may not be known or varying (e.g. CDN) or need more metadata (e.g. AWS_ENDPOINT)
- this PR introduce a new DataPath variant EXTERNAL where the tile files path is absolute and external (not refering the file system but rather a URL or gdal virtual file system)

### DIFF
--- a/geoengine/services/src/api/model/services.rs
+++ b/geoengine/services/src/api/model/services.rs
@@ -12,7 +12,8 @@ use crate::api::model::operators::{
 use crate::datasets::DatasetName;
 use crate::datasets::external::{GdalRetries, WildliveDataConnectorAuth};
 use crate::datasets::storage::validate_tags;
-use crate::datasets::upload::{UploadId, VolumeName};
+use crate::datasets::upload::{UploadId, UploadRootPath, VolumeName, Volumes};
+use crate::error::{Error, Result};
 use crate::projects::Symbology;
 use crate::util::Secret;
 use crate::util::oidc::RefreshToken;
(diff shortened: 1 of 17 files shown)


geoengine/services/src/api/model/services.rs
ChristianBeilschmidt, line 170: I know it is not part of this PR, but maybe add docs to a DataPath

geoengine/services/src/api/model/services.rs
```diff
@@ -175,6 +178,63 @@ impl TestDefault for DataPath {
...
+    ///
+    /// For `External`, the file path must be a GDAL virtual filesystem path
+    /// (e.g., `/vsicurl/` or `/vsis3/`), not a local filesystem path.
+    /// For `Volume` and `Upload`, the file path must be a relative path
+    /// with no root, parent, or current-directory components.
+    pub fn validate_file_path(&self, file_path: &Path) -> Result<()> {
+        match self {
+            DataPath::External => {
+                let path_str = file_path.to_string_lossy();
+                if !path_str.starts_with("/vsicurl/") && !path_str.starts_with("/vsis3/") {
```
ChristianBeilschmidt: Add more (https://gdal.org/en/stable/user/virtual_file_systems.html) or YAGNE?

Maybe also… GDAL External Dataset? Is specific to GDAL, right?

ChristianBeilschmidt (changes requested):

geoengine/services/src/api/model/services.rs
michaelmattig: 1e2f0171761913860158a569dc3e0fb57536b89e

geoengine/services/src/api/model/services.rs
michaelmattig: for the other virtual file systems we would have to extend the Ogr/GdalDatasetParameters I think. Then the file_path is the location and other stuff like untaring or caching or streaming could be defined as metadata.

ChristianBeilschmidt (dismissed):

ChristianBeilschmidt (approved):

PR #1267, opened 2026-09-02
feat(backend): implement CORS support for OGC API responses

ChristianBeilschmidt:

Added CORS configuration and support for OGC API Tiles.

### DIFF
--- a/geoengine/Settings-default.toml
+++ b/geoengine/Settings-default.toml
@@ -76,6 +76,11 @@ type = "Value"
 start = "2014-01-01T00:00:00Z"
 end = "2014-01-01T00:00:00Z"
 
+[ogc.cors]
+allow_origin = "*"
+allow_headers = ["Authorization", "Content-Type"]
+allow_methods = ["GET", "OPTIONS", "HEAD"]
+
 [wcs]
 # max number of tiles to be produced for generating output tiff
 tile_limit = 4
--- a/geoengine/services/src/api/handlers/ogc/cors.rs
+++ b/geoengine/services/src/api/handlers/ogc/cors.rs
@@ -1,0 +1,183 @@
+use crate::config;
+use actix_web::{
+    Error, HttpResponse,
+    body::MessageBody,
+    dev::{ServiceRequest, ServiceResponse},
+    http::{
+        Method,
+        header::{
+            ACCESS_CONTROL_ALLOW_HEADERS, ACCESS_CONTROL_ALLOW_METHODS,
+            ACCESS_CONTROL_ALLOW_ORIGIN, HeaderValue, VARY,
+        },
+    },
+};
+
+fn ogc_config() -> Result<config::Ogc, Error> {
+    config::get_config_element::<config::Ogc>().map_err(Into::into)
+}
+
+/// Applies CORS headers to the given response based on the OGC configuration.
+///
+/// # Errors
+///
+/// Returns an `Error` if the OGC configuration cannot be retrieved or if any of the header values are invalid.
+///
+fn apply_cors_headers<B>(response: &mut ServiceResponse<B>) -> Result<(), Error>
+where
+    B: MessageBody + 'static,
+{
+    let Some(config) = ogc_config()?.cors else {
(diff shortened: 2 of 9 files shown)


geoengine/services/src/api/handlers/ogc/cors.rs
```diff
@@ -0,0 +1,183 @@
...
+///
+/// # Errors
+///
+/// Returns an `Error` if the OGC configuration cannot be retrieved or if any of the header values are invalid.
+///
+pub async fn cors_middleware(
+    req: ServiceRequest,
+    next: actix_web::middleware::Next<impl MessageBody + 'static>,
+) -> Result<ServiceResponse<actix_web::body::BoxBody>, Error> {
+    if req.method() == Method::OPTIONS {
```
michaelmattig: call `next` also for OPTIONS request? otherweise other middlewares won't execute.

geoengine/Settings-default.toml
```diff
@@ -76,6 +76,11 @@ type = "Value"
 start = "2014-01-01T00:00:00Z"
 end = "2014-01-01T00:00:00Z"
 
+[ogc.cors]
+allow_origin = "*"
```
michaelmattig, line 80: should default maybe be the external_address? would also be nice to avoid typing it in two places, so allow_origin could be an enum ExternalAddress, Any, Custom(String)

michaelmattig: maybe that's overkill though.

geoengine/Settings-default.toml
```diff
@@ -76,6 +76,11 @@ type = "Value"
 start = "2014-01-01T00:00:00Z"
 end = "2014-01-01T00:00:00Z"
 
+[ogc.cors]
```
michaelmattig, line 79: does it make sense to configure cors on a per handler basis or should it be a global setting?

geoengine/Settings-default.toml
```diff
@@ -76,6 +76,11 @@ type = "Value"
...
+allow_origin = "*"
+allow_headers = ["Authorization", "Content-Type"]
```
michaelmattig, line 81: what about
Accept-Encoding: gzip, deflate, br, zstd

and other headers?

geoengine/Settings-default.toml
ChristianBeilschmidt: Why would we need to set CORS for the same origin?

geoengine/Settings-default.toml
ChristianBeilschmidt: Maybe, but could be YAGNI

geoengine/Settings-default.toml
ChristianBeilschmidt: So wie ich das sehe nein, da Teil von Forbidden Request Headers: https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_request_header

geoengine/services/src/api/handlers/ogc/cors.rs
ChristianBeilschmidt: No, I guess we should not call next for OPTIONS in a CORS preflight middleware. A browser preflight is not a normal route request, so the middleware should short-circuit with a 204 response header instead of letting it fall through.

Also, we don't want the request to be executed or be messed up by other middleware and suddenly have a 404.

michaelmattig: ok, CORS preflights should short circuit.

however, you short-circuit any OPTIONS request now, which changes behavior when "ogc.cors = None". Currently, a missing header would yield an error even on OPTIONS request, now it would 204 and only on the actual request throw an error. this means one more request in this case.

but if you think it'S ok we can do it like this.

ChristianBeilschmidt: [5d9b4f3](https://github.com/geo-engine/geoengine/pull/1267/commits/5d9b4f357082d20640f61d3c2fa1c290cfb3865b)

michaelmattig (approved):

PR #1277, opened 2026-09-08
feat(backend): new raster cache

dbrandenstein:

<TEXT>


dbrandenstein:
[feat(backend): new raster cache](https://trello.com/c/NOEcNxh8/2030-featbackend-new-raster-cache)

geoengine/operators/src/cache/new_raster_cache/mod.rs
```diff
@@ -0,0 +1,1705 @@
...
+                cache.get(key).is_none_or(|sf| Arc::strong_count(sf) > 1)
+            })?;
+
+        let needed = (current_size + required_space).saturating_sub(eviction_strategy.capacity());
+        if eviction_plan.freed_bytes < needed {
+            return Err(CacheError::NotEnoughSpaceInCache.into());
+        }
+
+        if existing_size.is_some() {
+            cache.remove(&key);
```
michaelmattig, line 813: don't you have to remove the key from the eviction plan in case it was included? maybe also add a test for the case where the inserted key is already in the cache and also is the next item to evict. should panic currently.

geoengine/operators/src/cache/new_raster_cache/mod.rs
```diff
@@ -0,0 +1,1705 @@
...
+        let required_space = value.byte_size()?;
+
+        let mut cache = self.cache.write().await;
+        let mut eviction_strategy = self.eviction_strategy.write().await;
+
+        let mut current_size = self.total_size.load(Ordering::SeqCst);
+        let existing_size = cache.get(&key).map(|e| e.size);
+
+        if let Some(size) = existing_size {
+            current_size = current_size.saturating_sub(size);
```
michaelmattig, line 799: if a reader still holds a copy of the arc the item will not actually be released from memory. can we still subtract the value here even though the cache entry is still in memory?

geoengine/operators/src/cache/new_raster_cache/mod.rs
```diff
@@ -0,0 +1,1705 @@
...
+            }
+        }
+
+        Ok(hit)
+    }
+
+    async fn insert(&self, key: CacheKey, tile: TypedRasterTile2D) -> Result<()> {
+        let cache_hint = tile.cache_hint();
+        let value = SF::store(tile)?;
+        let required_space = value.byte_size()?;
```
michaelmattig, line 790: what about the size of the keys? they can be pretty large considering they store a whole worfklow.

geoengine/operators/src/cache/new_raster_cache/mod.rs
```diff
@@ -0,0 +1,1705 @@
...
+            .query(query_rectangle, &query_ctx)
+            .await
+            .unwrap()
+            .collect::<Vec<_>>()
+            .await
+            .into_iter()
+            .collect::<crate::util::Result<Vec<_>>>()
+            .unwrap();
+
+        assert_eq!(tiles, vec![expected_tile.clone()]);
```
michaelmattig, line 1638: this does not check if the cache is actually populated right?

geoengine/operators/src/cache/new_raster_cache/mod.rs
```diff
@@ -0,0 +1,1705 @@
...
+                self.eviction_strategy.write().await.record_hit(key);
+            }
+        }
+
+        Ok(hit)
+    }
+
+    async fn insert(&self, key: CacheKey, tile: TypedRasterTile2D) -> Result<()> {
+        let cache_hint = tile.cache_hint();
+        let value = SF::store(tile)?;
```
michaelmattig, line 789: should'nt we use a seprate task for compression?

michaelmattig: also for decompression

geoengine/operators/src/engine/query.rs
```diff
@@ -153,6 +158,9 @@ impl MockQueryContext {
             cache: None,
             quota_checker: None,
             quota_tracking: None,
+            new_raster_cache: Some(Arc::new(
+                crate::cache::new_raster_cache::NewRasterCacheEnum::new_lru(usize::MAX / 2),
```
michaelmattig, line 162: is this a good default for the size?

geoengine/services/src/config.rs
```diff
@@ -563,6 +563,32 @@ pub struct Cache {
...
+    pub enable_new_raster_cache: bool,
+    /// if `enable_new_raster_cache` is set, the fraction of `size_in_mb` reserved for the new
+    /// raster cache; the remainder is used by the old cache (which then only holds vector data).
+    /// Ignored if `enable_new_raster_cache` is `false`, in which case the old cache holds both
+    /// vector and raster data and gets the full `size_in_mb`.
+    pub raster_cache_size_ratio: f64,
+}
+
+impl Cache {
+    pub fn shared_cache_size(&self) -> usize {
```
michaelmattig, line 577: maybe add units to all the cache size methods (shared and new)

geoengine/services/src/config.rs
```diff
@@ -563,6 +563,32 @@ pub struct Cache {
...
     pub size_in_mb: usize,
     pub landing_zone_ratio: f64,
+    /// use the new raster cache implementation for raster data instead of the old cache
+    #[allow(clippy::struct_field_names)]
+    pub enable_new_raster_cache: bool,
+    /// if `enable_new_raster_cache` is set, the fraction of `size_in_mb` reserved for the new
+    /// raster cache; the remainder is used by the old cache (which then only holds vector data).
+    /// Ignored if `enable_new_raster_cache` is `false`, in which case the old cache holds both
+    /// vector and raster data and gets the full `size_in_mb`.
+    pub raster_cache_size_ratio: f64,
```
michaelmattig, line 573: maybe check if it is between 0 and 1?

geoengine/operators/src/cache/new_raster_cache/mod.rs
```diff
@@ -0,0 +1,1705 @@
...
+                    );
+
+                    let cache_store = ctx
+                        .new_raster_cache()
+                        .expect("Cache should have been created for this operator");
+
+                    match cache_store.get(&key).await {
+                        Err(err) => Some((Err(err), (work, idx + 1))),
+                        Ok(Some(stored_tile)) => {
+                            tracing::debug!("Cache HIT for tile {:?}", key);
```
michaelmattig, line 990: set level to tracing

geoengine/operators/src/cache/new_raster_cache/mod.rs
```diff
@@ -0,0 +1,1705 @@
...
+                                Err(err) => Err(err)
+                            };
+
+                            Some((
+                                tile,
+                                (work, idx + 1),
+                            ))
+                        }
+                        Ok(None) => {
+                            tracing::debug!("Cache MISS for tile {:?}", key);
```
michaelmattig, line 1004: set level to tracing

geoengine/operators/src/cache/new_raster_cache/mod.rs
```diff
@@ -0,0 +1,1705 @@
...
+use geoengine_datatypes::util::ByteSize;
+use geoengine_datatypes::util::test::TestDefault;
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::iter;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use tokio::sync::RwLock;
+
```
michaelmattig, line 31: maybe as a lest step split up the module further: eviction, storage, operators_wrappings ?

geoengine/operators/src/engine/query.rs
dbrandenstein: It's just the mock so it probably doesn't matter much. But I can change it to 1GB or something. If you need something else you can always use the constructor with extensions.

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: You mean a tokio task? Why would it be better if we wait for it to finish to know the compressed size anyways?

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: I think key size was also ignored in the old cache and I just assumed it should stay this way. But we could include it if you think it's important.

michaelmattig: you are right, but previously the key was only stored once per query, now it is stored for each tile, so the impact is larger.

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: Actually when I think about it it might be better to just leave the old entry in the cache and throw away the new one, then we don't have issues with pinned entries etc. and can already skip the compression step. It might make sense though to update the tile's cache hint if the new entry has a longer validity, but that can also be a future improvement and should not trigger or make a big difference too often I think since the "double-insert" only happens if two queries start computing from source at a similar point in time leading to a similar cache hint.

geoengine/operators/src/cache/new_raster_cache/mod.rs
michaelmattig: you shouldn't do cpu-heavy stuff in async functions because it blocks other async functions from completion

> To give a sense of scale of how much time is too much, a good rule of thumb is no more than 10 to 100 microseconds between each .await. That said, this depends on the kind of application you are writing.

https://ryhl.io/blog/async-what-is-blocking/

do you know how costly the compression is?

geoengine/operators/src/engine/query.rs
michaelmattig: does it even make sense to use lru here if you set the limit so high that nothing ever gets evicted?

geoengine/operators/src/cache/new_raster_cache/mod.rs
michaelmattig: we assume that tiles never change right? so the newly inserted tile has to be the same one that is already present in the cache?

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/32d17362f92cf38f30dac1e631523768c3186126

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/32d17362f92cf38f30dac1e631523768c3186126

geoengine/services/src/config.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/d66635a79719db9db5c244486194808e24ad1b2f

geoengine/services/src/config.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/b07a557e76524331e978d45cfd0a3fe6817360e6

geoengine/operators/src/engine/query.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/5386ed0b96d090bdde821fcf0f099c44a13f22bc

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/97d593c880d5cea3d97d52232d12aad11e9b1bc4

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/73e995bf5ed3d802d1edfae0ff89df80f66a75bb

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/99ad90202bccd9f785bc9bc7c9a55c9f72bbb0fa

geoengine/operators/src/cache/new_raster_cache/mod.rs
dbrandenstein: https://github.com/geo-engine/geoengine/pull/1277/commits/8558d954f3943fa04759e4c979b6db1842afbd90

PR #1280, opened 2026-09-09
Fix/query abort propagation

jdroenner:

- improve disconnect detection
- use specific error in streams to cancel try_fold and other methods


ui/projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -85,6 +86,21 @@ export abstract class MapLayerComponent<OL extends OlLayer<OS, any>, OS extends
...
 
+    /**
+     * Remove a tile from the renderer cache so OpenLayers re-creates and re-fetches it (as an IDLE tile)
+     * when it is requested again.
+     */
+    protected evictTileFromRendererCache(tile: OlImageTile): void {
+        // Tile layers only, so narrow the (generic) base layer to a tile layer once.
+        const tileLayer = this._mapLayer as unknown as OlLayerTile;
+        const tileCache = tileLayer.getRenderer()?.getTileCache();
+        const renderSource = tileLayer.getSource();
```
michaelmattig: you use the current source and not the source that created the cancelled tile. 
If the time or projection changed in the meantime, the source might have been replaced.
in this case the cache entry will not be found.

I think you have to keep track of the source to build the proper cache key...

michaelmattig: maybe check if it works with `containsKey` on the cache?

ui/projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -465,22 +494,55 @@ export class OlOgcApiMapTileLayerComponent extends MapLayerComponent<
...
                 tileLoadFunction: (olTile, src): void => {
                     void (async (): Promise<void> => {
+                        const controller = new AbortController();
+                        this.tileAbortControllers.add(controller);
+
+                        let cancelSub: Subscription | undefined;
                         try {
+                            const tile = olTile as OlImageTile;
+                            const tileCoord = tile.getTileCoord();
+                            const tileZoomLevel = tileCoord[0];
```
michaelmattig, line 522: I think for ogc api tile you need to use `tileGrid.getResolution(tileZoomLevel)` instead of `const tileResolution = this.mapService.getView().getResolutionForZoom(tileZoomLevel);` (which the `createQueryAbortStream` does) because map resolution and tile resolution can differ. so maybe pass the resolution into the `createQueryAbortStream` function instead of the zoomlevel.

geoengine/services/src/util/server.rs
```diff
@@ -343,38 +361,82 @@ pub fn connection_init(connection: &dyn Any, data: &mut Extensions) {
...
             let fd = RawFd::from(fd);
+
+            // If a previous, now-closed connection is still being monitored under this fd
+            // number, the kernel re-used it for the new connection. Invalidate the old
+            // monitors so they report the close immediately instead of polling the fd of
+            // an unrelated, live connection forever (which would delay cancellation).
+            let mut monitored = MONITORED_FDS
+                .lock()
+                .unwrap_or_else(std::sync::PoisonError::into_inner);
+            if let Some(still_open) = monitored.remove(&fd) {
```
michaelmattig: I think there could be a race condition where the connection_closed was not yet called, but a new connection on the same fd is already established. Linux can reassign the fd immediately on fd close and it's not guaranteed that geo engine already processed this before it is reused.

in this case the fd would not exists yet in the monitored map, because it is only inserted during connection_close.

I think you need to insert the fd connection in `connection_init` already.

jdroenner: https://github.com/geo-engine/geoengine/pull/1280/commits/289667b073aa59fbbcd57e493f5be4355c0e9df8

ui/projects/core/src/lib/map/map-layer.component.ts
jdroenner: https://github.com/geo-engine/geoengine/pull/1280/commits/d33afcb6ee422d3ce2202e6ab4a6150c04c45445

ui/projects/core/src/lib/map/map-layer.component.ts
jdroenner: https://github.com/geo-engine/geoengine/pull/1280/commits/d33afcb6ee422d3ce2202e6ab4a6150c04c45445

ui/projects/core/src/lib/project/project.service.ts
```diff
@@ -976,24 +976,25 @@ export class ProjectService implements OnDestroy {
...
         const layerRemovedSubject = new BehaviorSubject<boolean>(false);
-        const layerStreamSub = layerStream.subscribe({
-            complete: () => {
-                layerRemovedSubject.next(true);
-                layerRemovedSubject.complete();
-            },
-        });
+        let layerStreamSub: Subscription | undefined;
+        if (layerStream) {
+            layerStreamSub = layerStream.subscribe({
```
michaelmattig: this subscription is only unsubscribed here: https://github.com/geo-engine/geoengine/blob/7dc1a89185e938bcce98c8247151c907e001eeaa/ui/projects/core/src/lib/project/project.service.ts#L1035

I think this leaks the subscription when the query is never aborted

ui/projects/core/src/lib/map/map-layer.component.ts
michaelmattig: I thought maybe do a console.log on containskey. is there ever anything logged? in the final code containskey is not necessary I think because remove checks this as well.

ui/projects/core/src/lib/map/map-layer.component.ts
michaelmattig: https://github.com/geo-engine/geoengine/blob/7dc1a89185e938bcce98c8247151c907e001eeaa/ui/projects/core/src/lib/project/project.service.ts#L1029

AI comment:
> These are different resolution ladders. OpenLayers may render a custom-grid tile with resolution 10 in a view whose resolution is 9.5546. A pan emits another viewport with the same 9.5546 resolution, but the comparison against 10 remains true, aborting an otherwise relevant visible request.

AI propsal:
> Capture the initial viewportSize.resolution inside the observable and compare subsequent viewport values against that. The tile-resolution parameter can then be removed.

ui/projects/core/src/lib/project/project.service.ts
jdroenner: https://github.com/geo-engine/geoengine/pull/1280/commits/2d42d0e90e4e8b289a1eeb89b52008ee4f4291d4

ui/projects/core/src/lib/map/map-layer.component.ts
jdroenner: https://github.com/geo-engine/geoengine/pull/1280/commits/45041c8ee79ebb14ac948c82c0e8e3fb1f934150

ui/projects/core/src/lib/map/map-layer.component.ts
jdroenner: hab ich gemacht. da wird auch was geloggt. ABER es hat tatsächlich einige Fehler erzeugt. Nach weiterer Recherche kann man sich das manipulieren des Cache auch sparen und einfach den State des Tile anders setzten:
 https://github.com/geo-engine/geoengine/pull/1280/commits/45041c8ee79ebb14ac948c82c0e8e3fb1f934150

ui/projects/core/src/lib/project/project.service.ts
```diff
@@ -1009,34 +1014,32 @@ export class ProjectService implements OnDestroy {
...
             take(1),
             map(() => undefined),
         );
     }
 
+    /**
+     * Read the current value of a stream that emits synchronously on subscription
+     * (e.g. one backed by a `ReplaySubject` or `BehaviorSubject`).
+     */
+    private captureSync<T>(stream: Observable<T>): T {
```
ChristianBeilschmidt: https://rxjs.dev/api/index/function/firstValueFrom

ui/projects/core/src/lib/project/project.service.ts
```diff
@@ -1009,34 +1014,32 @@ export class ProjectService implements OnDestroy {
...
         return combineLatest(observables).pipe(
-            tap(([time, _viewportSize, session, sref, _layerRemoved]) => {
-                // capture the initial values at the start of the query
-                // s.t. we can detect a change later
-                initialTime ??= time;
-                initialSref ??= sref;
-                initialSession ??= session;
-            }),
-            skip(1),
             filter(
```
michaelmattig: somehow it makes no sense to combine the `this.userService.getSessionTokenForRequest(),`. it uses pipe(first()) and thus never fires uupon session change.

geoengine/operators/src/engine/query.rs
```diff
@@ -261,3 +298,50 @@ impl GdalProcessPoolAccess for MockQueryContext {
...
+
+        assert!(matches!(
+            stream.next().await,
+            Some(Err(error::Error::QueryCanceled))
+        ));
+        assert!(stream.next().await.is_none());
+    }
+
+    #[tokio::test]
+    async fn it_completes_normally_when_trigger_is_dropped_unfired() {
```
michaelmattig: do we want it to continue normally when trigger is dropped? previously it cancelled on drop of the trigger.

ui/projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -85,6 +85,19 @@ export abstract class MapLayerComponent<OL extends OlLayer<OS, any>, OS extends
...
+    /**
+     * Reset an aborted tile to `IDLE` so OpenLayers re-requests it (e.g. after panning away and
+     * back). OpenLayers only requests IDLE tiles, so an ERROR tile would stay invisible forever.
+     *
+     * `setState` must not go from `LOADING` to `IDLE` directly ("Tile load sequence violation"),
+     * so it is routed through `ERROR` first, which also lets the source fire the matching
+     * `tileloaderror` and keep its in-flight tile bookkeeping balanced.
+     */
+    protected resetAbortedTile(tile: OlImageTile): void {
+        tile.setState(TileState.ERROR);
```
michaelmattig, line 105: doesn't this trigger a tile load error that is captured here? https://github.com/geo-engine/geoengine/blob/2d42d0e90e4e8b289a1eeb89b52008ee4f4291d4/ui/projects/core/src/lib/map/map-layer.component.ts#L433

michaelmattig: maybe we need to track aborted layers separately

ui/projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -335,19 +348,22 @@ export class OlRasterLayerComponent
...
             const tileExtent = tileGrid.getTileCoordExtent(tileCoord) as Extent;
 
             const client = new XMLHttpRequest();
 
-            const cancelSub = this.projectService
-                .createQueryAbortStream(this.layerId(), tileZoomLevel, tileExtent)
-                .subscribe(() => client.abort());
+            let aborted = false;
+
+            const cancelSub = this.projectService.createQueryAbortStream(this.layerId(), tileExtent).subscribe(() => {
```
michaelmattig, line 371: problem is the query abort stream can emit an event before the request is even sent (e.g. during panning). then the client is aborted before client.send() is called but then it is sent anyway.

so before client.send check if it was already aborted?

ui/projects/core/src/lib/map/map-layer.component.ts
michaelmattig: I think you need to clear abortedTiles when a new loading attempt starts, in both wms and ogc. Otherwise, after an aborted tile is retried, a genuine failure is still treated as an abort and the layer is reported as OK.

jdroenner: meine Kei sagt jetz man könne das auch ohne das Set machen: https://github.com/geo-engine/geoengine/pull/1280/commits/445e9ddf29091870c3227b90f74c53905e6835c9

ui/projects/core/src/lib/project/project.service.ts
jdroenner: das wurde ersetzt durch pairwise() weil das auch sync ist aber nicht so einen Helfer braucht.

ui/projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -514,6 +613,10 @@ export class OlOgcApiMapTileLayerComponent extends MapLayerComponent<
             };
             const onEnd = (): void => {
                 tilesPending--;
+                if (this.resettingAbortedTile) {
+                    // abort is transient, the retry's tileloadstart will update loading
+                    return;
```
michaelmattig, line 711: if you return here, the loading indicator may spin indefinetely if the user e.g. pans to an area that was already cached. then no retry will happen and no new tile will be loaded and onEnd never gets called leaving the loading indicator spinning forever, I think.

michaelmattig: maybe just remove this check?

michaelmattig: if the loading indicator starts to flicker, maybe we can just debounce it 😅

ui/projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -255,6 +285,12 @@ export class OlRasterLayerComponent
         if (this.timeSubscription) {
             this.timeSubscription.unsubscribe();
         }
+
+        // abort all WMS tile requests that are still in flight
+        for (const client of this.tileAbortClients) {
+            client.abort();
```
michaelmattig, line 300: you abort here without setting the `aborted` flag. in the loadend handler it will then publish an LoadingState.ERROR, I think.

maybe add an abort listener to each client

```
client.addEventListener('abort', () => {
    aborted = true;
});
```

jdroenner: https://github.com/geo-engine/geoengine/pull/1280/commits/8dc160a87364fc334812855d0e4c1db4215eb586
there is now a per client send() check and each client registers the abort handler. so tiles are no longer se to ERROR but IDLE

ui/projects/core/src/lib/map/map-layer.component.ts
jdroenner: Also die KI sagt das funktioniert so weil der eine Teil nur theoretisch passiert. Vielleicht gibt es den Edge-Case aber müsste man mal beobachten, odeR?

>Removing the check would bring the flicker back: a tile can only be reset by routing it through `ERROR` first, which fires a synchronous `tileloaderror` — without the guard, `loading` would briefly drop to `false` right before the retry sets it to `true` again.

> The stuck-indicator case stays theoretical: when an aborted tile is still in view it's re-requested (`IDLE` → `tileloadstart` keeps `loading` true until it finishes), and when it's not, the pan/zoom that aborted it always brings the new viewport's tile events, which drive `loading` back to `false`. So the guard only hides the transient blip, never real completion.

michaelmattig: >  it always brings the new viewport's tile events

nicht wenn man irgendwohin zoomt wo schon alles gecacht ist. dann wird nichts geladen und es gibt kein onEnd. 

> Suggested fix: remove that early return and debounce the loading signal if flicker appears.

jdroenner: https://github.com/geo-engine/geoengine/pull/1280/commits/c6dd705c04c35afee72703a2f8966dc459bb45e1

michaelmattig: bei mr kommt jetzt gar keine ladeanmination mehr. teste vielleicht mal selber gegen die staging-instanz:

npm run serve:edv:staging

dafür musst du allerdings ins Wireguard VPN. Wäre aber eh sinnvoll wenn du das mal einrichtest.

die KI schlägt folgendes vor:

statt

```
debounce((loading) => (loading ? EMPTY : timer(150)))
```
ersetzen durch:
```
debounce((loading) => (loading ? of(undefined) : timer(150)))
```
bei mir funktionniert es

PR #1299, opened 2026-09-22
feat: make cache ttl configurable in settings and provider def

michaelmattig:

<TEXT>


geoengine/Settings-default.toml
```diff
@@ -230,6 +230,8 @@ volumes = { test_data = "test_data/" }
 
 [cache]
 enabled = false
+# Default cache lifetime in seconds for values without an explicit TTL (0 disables caching).
+default_ttl_seconds = 0
```
michaelmattig, line 234: set to one day? may be bad for testing stuff, though.

